PR #10, opened 2023-02-08
feat: Mosaic Jupyter widget

domoritz:
Adds code for a Jupyter widget. The widget is published at https://pypi.org/project/mosaic-widget/. 

The widget works in colab as well: https://colab.research.google.com/drive/1Txy6L_of8_lJFImKEkhUCqZX70yKpYnv

Screenshot 2023-02-07 at 21 44 16

### DIFF
--- a/dev/notebooks/weather.ipynb
+++ b/dev/notebooks/weather.ipynb
@@ -19,7 +19,9 @@
    "cell_type": "code",
    "execution_count": 2,
    "id": "ed9edb9e-4ed2-4e5e-b44c-81a2c7a73af4",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "# Optional: enable logging in Mosaic widget\n",
@@ -51,7 +53,7 @@
    },
    "outputs": [],
    "source": [
-    "# for now add the table to duckdb since otherwise the table_info pragam doesn't work\n",
+    "# for now add the table to duckdb since otherwise the table_info pragma doesn't work\n",
     "con = duckdb.connect()\n",
     "con.execute(\"CREATE TABLE weather AS SELECT * FROM weather\");"
    ]
@@ -67,7 +69,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "7ca4bf8e840e41cba7f21d26c514e9c4",
+       "model_id": "403308ec1a9447828a4c3bd2916ec9d9",
        "version_major": 2,
        "version_minor": 0
       },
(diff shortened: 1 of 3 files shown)


dev/notebooks/weather.ipynb
```diff
@@ -0,0 +1,109 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "e5dd9878-74bd-4b18-bc15-1cb12f6ac71b",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# for now add the table to duckdb since otherwise the table_info pragam doesn't work\n",
```
jheer: Typo: pragam -> pragma

dev/notebooks/weather.ipynb
```diff
@@ -0,0 +1,109 @@
...
+    "con = duckdb.connect()\n",
+    "con.execute(\"CREATE TABLE weather AS SELECT * FROM weather\");"
```
jheer, line 58: Looks like we can't query metadata in pandas tables without explicitly adding the table to DuckDB? That's too bad... it would be nice to be able to use Mosaic against Pandas tables without having to do a formal ingest. That said, I suppose a DSL could handle this behind the scenes when one passes a Pandas data frame reference as the source data for a component.

packages/widget/pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "mosaic-widget"
+description = "A Juptyter widget for mosaic"
```
jheer: Typo: Juptyter -> Jupyter

packages/widget/src/index.js
```diff
@@ -0,0 +1,60 @@
...
+      }
+      default: {
+        query.resolve({});
+        break;
+      }
+    }
+    console.groupEnd("query");
+  });
+
+  coordinator(new Coordinator(client));
```
jheer: Probably safer to do this instead: `coordinator().databaseClient(client)`

Just set the client rather than create a new Coordinator instance. Is the `coordinator()` singleton shared in the browser client across multiple widgets? If so, not resetting the coordinator is important.

dev/notebooks/weather.ipynb
domoritz: I am having a discussion about this on the DuckDB discord. We can use `DESCRIBE` apparently but need to change Mosaic since the column names are different. I was going to do that in a follow up pull request but wanted to get this version out that doesn't need changes in Mosaic.

packages/widget/src/index.js
domoritz: Done

PR #12, opened 2023-02-08
Use DESCRIBE instead of PRAGMA table_info.

jheer:
- Use DESCRIBE instead of PRAGMA table_info to get table schema metadata.

In response to discussion in #10, where apparently DuckDB Python can get schema info for Pandas tables using DESCRIBE but not with `PRAGMA table_info`.

In the future we may wish to support schema lookup not just for tables, but also queries with derived/transformed columns, in which case we will need to use `DESCRIBE` anyway.

### DIFF
--- a/packages/core/src/Catalog.js
+++ b/packages/core/src/Catalog.js
@@ -20,20 +20,19 @@ export class Catalog {
     }
 
     const q = this.mc.query(
-      `PRAGMA table_info('${table}')`,
+      `DESCRIBE "${table}"`,
       { type: 'json', cache: false }
     );
 
     return (cache[table] = q.then(result => {
       const columns = object();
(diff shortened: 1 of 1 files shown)


packages/core/src/Catalog.js
```diff
@@ -20,20 +20,19 @@ export class Catalog {
...
           table,
-          column: name,
-          sqlType: type,
-          type: jsType(type),
-          notnull,
-          pk
+          column: entry.column_name,
+          sqlType: entry.column_type,
+          type: jsType(entry.column_type),
+          nullable: entry.null === 'YES'
```
domoritz, line 35: Knowing whether a column is a key may be useful in the future. 

```suggestion
          nullable: entry.null === 'YES',
          key
```

PR #37, opened 2023-03-03
Reset selection source upon clause eviction

jheer:
Selections with a _single_ resolution strategy will evict an existing clause upon updates. This PR adds a mechanism for reporting this eviction back to a clause _source_: if a _source_ has a `reset()` method, it will be invoked upon clause eviction.

Both the `menu` and `search` widgets and the `Interval1D` and `Interval2D` interactors now have `reset()` methods that clear their interactive state upon selection clause eviction.


manzt:
Is there a mechanism to batch `.reset()` calls? For instance, I'd like to wire up a button to "Reset all" `1DInterval`s.

PR #43, opened 2023-03-06
feat: make selections available in Python

domoritz:
I also fixed some issues with resetting I found along the way.

### DIFF
--- a/dev/notebooks/specs.ipynb
+++ b/dev/notebooks/specs.ipynb
@@ -13,7 +13,8 @@
     "import yaml\n",
     "from mosaic_widget import MosaicWidget\n",
     "import ipywidgets as widgets\n",
-    "import pyarrow as pa"
+    "import pyarrow as pa\n",
+    "from pprint import pprint"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


packages/widget/src/index.js
```diff
@@ -25,36 +27,53 @@ export async function render(view) {
...
+    reset();
+    logger.log('Setting spec:', spec);
     view.el.replaceChildren(await parseJSON(spec));
+
+    // Update the selections traitlet
+    const c = coordinator();
+    const selections = new Set([...c.clients].flatMap(c => c.filterBy).filter(s => s))
+    for (const s of selections) {
+      s.addEventListener('value', () => {
+        const s = [...selections].map(s => s.clauses.map(c => ({predicate: c.predicate})));
```
jheer: This should include the clause _value_ property as well. It might also aid understanding/reuse to include generated SQL for each predicate (easy to do as `sql: String(c.predicate)`).

Down the road, we should also discuss to what degree Mosaic should track all Param/Selection instances (likely via the Coordinator) to simplify getting/setting interactive state. This would also happen naturally if we later want to put a dedicated reactive runtime (rather than the current listeners) in place.

domoritz: Yes. It would also be great to be able to set selections easily. It's super easy to listen to changes from python via the traitlets.

PR #45, opened 2023-03-15
feat: add gzip compression

domoritz:


packages/duckdb/src/data-server.js
```diff
@@ -17,15 +21,14 @@ export function dataServer(db, {
...
+function createHTTPServer(handleQuery, rest, gzip) {
   return http.createServer((req, resp) => {
-    const res = httpResponse(resp);
+    const res = httpResponse(resp, gzip);
     if (!rest) {
       res.done();
       return;
     }
 
-    resp.setHeader('Content-Type', 'application/json');
```
domoritz, line 28: I think this was incorrect because we don't always send json.

package.json
```diff
@@ -13,7 +13,7 @@
     "build": "lerna run build",
     "lint": "lerna run lint",
     "test": "lerna run test",
-    "server": "node bin/run-server.js",
+    "server": "nodemon bin/run-server.js",
```
domoritz, line 16: This automatically restarts the server when there is a change.

PR #46, opened 2023-03-15
chore: update duckdb

domoritz:
Update to duckdb 0.7.1

Need to not use streaming because of https://github.com/duckdb/duckdb/issues/6743. The non-streaming version is supposedly faster because "the streaming query results currently cannot use parallelism properly" (according to DuckDB dev).

I also had to add some casting because we don't handle BigInts well and https://github.com/duckdb/duckdb/issues/6742.

### DIFF
--- a/packages/widget/vite.config.mjs
+++ b/packages/widget/vite.config.mjs
@@ -1,5 +1,5 @@
 import { defineConfig } from "vite";
-import anywidget from "anywidget/vite";
+import anywidget from "@anywidget/vite";
 
 export default defineConfig({
   build: {


packages/vgplot/src/marks/RegressionMark.js
```diff
@@ -31,8 +31,8 @@ export class RegressionMark extends Mark {
         ssy: regrSYY(y, x),
         ssx: regrSXX(y, x),
         xm: regrAvgX(y, x),
-        x0: min(x).where(isNotNull(y)),
-        x1: max(x).where(isNotNull(y))
+        x0: minInt(x).where(isNotNull(y)),
```
jheer: Why assume integers here? The underlying x domain may be any numeric type (e.g., doubles). The result is not a pixel `x` value but rather a continuous value that will later map to an `x` encoding channel.

domoritz: I'll change to double then.

packages/sql/src/aggregate.js
```diff
@@ -35,23 +36,25 @@ export class Aggregate {
...
+  return (...args) => new Aggregate(op, args, cast);
 }
 
-export const count = agg('COUNT');
+export const count = agg('COUNT', 'INTEGER');
 export const avg = agg('AVG');
 export const mean = agg('AVG');
 export const mad = agg('MAD');
 export const max = agg('MAX');
+export const maxInt = agg('MAX', 'INTEGER');
```
jheer: Do we really need these extra convenience methods? I don't think `min(col, 'INTEGER')` is all that onerous, and all things being equal would prefer to not increase the API surface area.

domoritz: That doesn't work I think since `'INTEGER'` is passed as an arg to the aggregate. We would still need to change the code here somehow.

jheer: Oh, right! In general using cast (or cast helper) methods might prove to be more reusable & decomplected :)

``` js
cast(max('col'), 'INTEGER') // general cast method
asInt(max('col')) // helper method, exact naming TBD
```

domoritz: I can add that. It still feels like we are mixing responsibilities a bit. The mark shouldn't have to worry about database specific issue (as you pointed out above) but I think casting is a useful feature in general and would help us update duckdb here.

PR #48, opened 2023-03-21
chore: upgrade deps and remove type hack from widget

domoritz:
Resolves a typing issue: https://twitter.com/maartenbreddels/status/1638204921701883905.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,14 +10,14 @@
         "packages/*"
       ],
       "devDependencies": {
-        "esbuild": "^0.17.12",
-        "eslint": "^8.36.0",
-        "lerna": "^6.5.1",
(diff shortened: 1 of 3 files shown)


packages/duckdb/package.json
```diff
@@ -24,6 +24,7 @@
     "prepublishOnly": "npm run test && npm run lint"
   },
   "dependencies": {
+    "@mapbox/node-pre-gyp": "^1.0.10",
```
jheer: Why is this needed?

domoritz: Let's try again but the CI build for duckdb failed without it when I tried two weeks ago.

domoritz: Seems to still work even when I remove the explicit dependency now.

PR #56, opened 2023-04-07
feat: switch to http2

domoritz:
This switches from rest + ws to [http2](https://en.wikipedia.org/wiki/HTTP/2), which is supposed to give better performance. I haven't done extensive testing to compare performance so we need to do that before merging.


domoritz:
Unclear benefit right now. Will maybe look into later.

PR #110, opened 2023-06-04
docs: add Vega client example

### LINKED ISSUE #82: Create example Vega + Mosaic integration.
Potential ideas for a Mosaic / Vega-Lite component:
1. Take a Vega-Lite specification as input, analyze transforms/encodings to extract SQL queries.
2. Instantiate Mosaic clients for data sources.
3. Feed Mosaic-provided query data to resulting (simplified) Vega-Lite visualizations.
4. Add listeners for param/selection changes and propagate those to Mosaic (we would need to "disable" downstream selection updates, so step 1 above might involve rewriting to the VL spec in order to leave filtering/domain binding up to Mosaic).

Achieving all the steps above might be a big lift, but even piecemeal integrations might be useful to showcase.

I view this as a stretch goal &ndash; _not_ a requirement &ndash; for an initial release.

domoritz:
Fixes #82 

This is a hard coded example just to show that Vega works with Mosaic. In future versions, it would be nice to have a method `vegaClients` that takes a spec, returns a set of clients (built from the components of the spec) and a Vega view that was built from updated specs.

- [x] Use active selection
- [x] Fix the selection (doesn't seem to align correctly yet)

### DIFF
--- a/packages/example-vega/main.js
+++ b/packages/example-vega/main.js
@@ -14,6 +14,7 @@ import {
 } from "@uwdata/mosaic-sql";
 import embed from "vega-embed";
 
+/** @type {import('vega-lite').TopLevelSpec} */
 const spec = {
   $schema: "https://vega.github.io/schema/vega-lite/v5.json",
   layer: [
@@ -28,7 +29,7 @@ const spec = {
       mark: "bar",
       encoding: {
         x: {
-          timeUnit: "month",
+          timeUnit: "utcmonth",
           field: "date",
           type: "ordinal",
         },
@@ -72,7 +73,7 @@ class SelectionVegaClient extends MosaicClient {
 
     if (this.selection) {
       this.view.addSignalListener("brush", (_name, signal) => {
-        const dates = signal.month_date;
+        const dates = signal.utcmonth_date;
         const value = dates
           ? isBetween(
               dateMonth("date"),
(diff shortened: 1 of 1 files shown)


packages/example-vega/main.js
```diff
@@ -0,0 +1,166 @@
...
+              dateMonth("date"),
+              [
+                dateMonth(literal(dates.at(0))),
+                dateMonth(literal(dates.at(-1))),
+              ],
+              false
+            )
+          : null;
+
+        const clause = {
```
jheer, line 90: The DataTileIndexer will not be able to optimize this clause because the `schema` is lacking `scale` information: we need to include the `type` (e.g., `"linear"`) as well as `domain` and `range` arrays:

```js
schema: {
  type: "interval",
  scale: { type: "linear", domain, range }
}
```

Here's where they get accessed and subsequently used: https://github.com/uwdata/mosaic/blob/main/packages/core/src/DataTileIndexer.js#L150

jheer: We should be able to extract that info directly from the Vega runtime API, assuming we know the corresponding scale name.

domoritz: Hmm, I just added that but need something else.

domoritz: Ah, it needs to be `scales`.

PR #119, opened 2023-06-20
chore: update deps

domoritz:


lerna.json
```diff
@@ -1,5 +1,4 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "useWorkspaces": true,
```
domoritz, line 3: Lerna showed a warning saying it's deprecated.

PR #120, opened 2023-06-22
feat: add support for persistent indexes and tables with a database context

domoritz:
We can now create indexes not just as temp tables but as persistent tables. This can be useful when storing those indexes while the data may be from another [attached database](https://duckdb.org/docs/sql/statements/attach.html). 

I also added some more types and fixed a few formatting issues.

### DIFF
--- a/packages/core/src/Catalog.js
+++ b/packages/core/src/Catalog.js
@@ -1,3 +1,4 @@
+import { quoteTableName } from '@uwdata/mosaic-sql';
 import { jsType } from './util/js-type.js';
 import { summarize } from './util/summarize.js';
 
@@ -61,7 +62,7 @@ export class Catalog {
 
 async function getTableInfo(mc, table) {
   const result = await mc.query(
-    `DESCRIBE "${table}"`,
+    `DESCRIBE ${quoteTableName(table)}`,
     { type: 'json', cache: false }
   );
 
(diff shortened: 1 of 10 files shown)


jheer:
I modified the PR to use a different approach. I do _not_ want to attach arbitrary properties to the coordinator instance.

Instead, we can now use the coordinator `configure()` call to pass an options object to the indexer. The sole external option at present is the boolean flag `temp` (default `true`) which configures whether temp tables are created. I also cleaned up the internal indexer implementation a bit.

I updated the widget code to use this mechanism. I think it might be a good idea to also rename the traitlet here. A non-temp table is not necessarily *persisted*, as DuckDB may be a memory-only instance. A naming more consistent with the coordinator / indexer that indicates whether or not indexes uses temp tables might be a better choice?

Finally, I also renamed `DataTileIndexer` to `DataCubeIndexer`, as I think that is a bit more descriptive. This is a change I've been meaning to make for a while and this seems like as good a time as any to make it!

domoritz:
Thank you. That looks a lot cleaner. I renamed the traitlet to `temp_cubes`. I'd also be happy with `temp_indexes`.

jheer:
I like temp_indexes!

domoritz:
Done with the rename.

jheer (approved):
LGTM. Thanks!

PR #154, opened 2023-08-17
fix: change binField to use min provided by bins

kwonoh:
If I understood `binField` and `bins` functions correctly, `min` provided by `bins` should be used for the output of `binField` since `nice=true` changes the `min`.

This PR replaces `min` to `b.min`

### DIFF
--- a/packages/vgplot/src/transforms/bin.js
+++ b/packages/vgplot/src/transforms/bin.js
@@ -38,7 +38,7 @@ function binField(mark, column, options) {
       const base = b.min === 0 ? col : `(${col} - ${b.min})`;
       const alpha = `${(b.max - b.min) / b.steps}::DOUBLE`;
       const off = options.offset ? `${options.offset} + ` : '';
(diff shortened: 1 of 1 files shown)


kwonoh:
Repro: https://observablehq.com/d/137aafc8775f84fb

Screenshot 2023-08-18 at 4 36 43 PM

The min of `values` is 0.3 and the max of `values` is 9. However, the generated histogram looks like there is a value between 9.3 and 10.3.

I believe this is because the below reason.

[`base` and `alpha` in `binField` function are computed with `b.min`](https://github.com/uwdata/mosaic/blob/27f2ec23ef952b6ef5243640e8969b8286f2a415/packages/vgplot/src/transforms/bin.js#L38-L39) which is the "nicer" `min` returned by `bin` function.

However, [the returning value of `binField` is computed with the original min.](https://github.com/uwdata/mosaic/blob/27f2ec23ef952b6ef5243640e8969b8286f2a415/packages/vgplot/src/transforms/bin.js#L41)

Thus, I think the `min` in line 41 should be changed to `b.min`.

jheer:
Good catch, thanks for the PR!

PR #155, opened 2023-08-21
Convert TIMESTAMP sql type to Date object

jzavala-gonzalez:
When using DuckDB TIMESTAMP data types, Arrow serializes to the int64 ms since epoch, but Mosaic never reconverts back into a Javascript Date object before passing the value to Plot. Existing behavior is that we see integers in the axis instead of formatted Date/time. Added a bit of a hack to the Marks class to catch this datatype but I'm not sure if it's best to place it inside queryResult or elsewhere. Also I'm not sure if other timestamp-ish datatypes will cause the same issue. Please let me know if I can be of any help. Thank you!!

### DIFF
--- a/packages/vgplot/src/marks/Mark.js
+++ b/packages/vgplot/src/marks/Mark.js
@@ -138,7 +138,19 @@ export class Mark extends MosaicClient {
   }
 
   queryResult(data) {
+    const timestamp_keys = Object.entries(this.stats).filter(([c, s]) => {
+      return s.sqlType === 'TIMESTAMP';
+    }).map(([c, s]) => c);
+    
     this.data = Array.from(data);
+    if (timestamp_keys.length > 0) {
+      this.data = this.data.map(d => {
+        timestamp_keys.forEach(k => {
+          d[k] = new Date(d[k]);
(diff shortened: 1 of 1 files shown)


jheer:
Thanks for raising the issue and for the PR! The fix in this PR will work for some vgplot marks, but not for `Mark` subclasses that override the `queryResult` method. For better or worse, it won't have any effect for other (non-vgplot) Mosaic clients. Meanwhile, related problems can also arise with large integers mapped to JS `BigInt` instances.

At the moment Mosaic passes whatever is provided by the current database connector as the query results to clients (and these results may or may not be Arrow tables). It is currently left to clients to parse/coerce data as needed. We need to think more about how to best handle in-browser data type management across the library as a whole, and what facilities/defaults we might want to impose.

jzavala-gonzalez:
Hi! Is there any update yet on using Timestamps in Mosaic? I'm still very interested in using Mosaic but my use case dataset is very heavy on timestamps is all.

jheer:
Yes, I've started PR #184, which supercedes this PR. Feel free to chime in there. Thanks!

PR #158, opened 2023-08-27
Add interactor mark selection to vgplot

spren9er:

Currently, when using interactors in _vgplot_, you need to make sure that the mark you interact with is the last one in your mark stack. This only allows to have interactive behavior with rudimentary plots. Several use cases can't be implemented in _vgplot_, e.g. having interactors with plots that have

- `text` marks in top layer (charts with labels)
- decorator marks in top layer (grid overlay)

- Add optional `index` attribute for interactors like _toggle_, _highlight_, _interval_ and _nearest_ for specific mark selection
- Add `peers` attribute to _toggle_ interactor — analog to _interval_ interactors (makes it possible to use _toggle_ in conjunction with charts with text labels)
- Update `highlight-toggle.yaml` example (Note: Previously, it was not possible to have toggle behavior for updated chart specification)
  - Add grid mark
  - Add text mark (bar labels)
  - Show index usage
- Update documentation
  - Add note that mark indexes can be used for selection
  - Add `index` attribute to interactors
  - Add `peers` attribute to _toggle_ interactor
 
## How does it work?

Still — by default — interactors are used for the most recently added mark. However, when providing an index to an interactor, the corresponding mark in mark stack of plot will be selected.

## Breaking changes

None, as existing API has not changed.

## Examples

- See new `higlight-toggle.yaml`
- Example `hexbin.yaml` could have an _intervalXY_ interactor (currently not possible because of hex grid overlay)

    ```yaml
    - mark: hexbin
      data: { from: flights, filterBy: $query }
      x: time
      y: delay
      fill: { count: }
      binWidth: 10
    - mark: hexgrid
      binWidth: 10
    - select: intervalXY
      as: $query
      index: 0
    ```

### DIFF
--- a/packages/vgplot/src/directives/interactors.js
+++ b/packages/vgplot/src/directives/interactors.js
@@ -7,29 +7,37 @@ import { Nearest } from '../interactors/Nearest.js';
 
 function interactor(InteractorClass, options) {
   return plot => {
-    const mark = plot.marks[plot.marks.length - 1];
+    const mark = findMark(plot.marks, options.index);
     plot.addInteractor(new InteractorClass(mark, options));
   };
 }
 
-export function highlight({ by, ...channels }) {
-  return interactor(Highlight, { selection: by, channels });
+function findMark(marks, index) {
+  const mark = marks[index ?? marks.length - 1];
+
+  if (!mark) throw new Error(`Mark not found with index ${index}.`);
+
+  return mark;
+}
+
+export function highlight({ by, index, ...channels }) {
+  return interactor(Highlight, { selection: by, index, channels });
 }
 
 export function toggle({ as, ...rest }) {
-  return interactor(Toggle, { ...rest, selection: as });
+  return interactor(Toggle, { selection: as, ...rest });
 }
 
-export function toggleX({ as }) {
-  return toggle({ as, channels: ['x'] });
+export function toggleX({ as, index }) {
+  return toggle({ as, index, channels: ['x'] });
 }
 
-export function toggleY({ as }) {
-  return toggle({ as, channels: ['y'] });
+export function toggleY({ as, index }) {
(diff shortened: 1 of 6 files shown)


packages/vgplot/src/directives/interactors.js
```diff
@@ -7,29 +7,37 @@ import { Nearest } from '../interactors/Nearest.js';
 
 function interactor(InteractorClass, options) {
   return plot => {
-    const mark = plot.marks[plot.marks.length - 1];
+    const mark = findMark(plot.marks, options.index);
```
jheer, line 10: This approach has some possible leakage: the index is used to find the mark, but the index is also passed as an option to the interactor class itself. The index is not currently an option that is formally accepted by those classes; if it were added in the future it would have to conform exactly with how it is being used here.

jheer:
Thanks for the PR! I see two separate additions here:

1. Expose a `peers` option for the toggle interactor. This looks good to me!
2. Make marks index-addressable by interactors. I'm not (yet) fully convinced this is necessary. In terms of design, exposing integer indices could be error-prone and brittle. An alternative might be to allow marks to have unique names/ids and let interactors refer to those instead.

Moreover, is addition 2 just a syntactic nicety? One should be able to include an interactor directly after its target mark, in the middle of a list of marks (they don't have to be at the end of a specification). So I'm not yet sure why the mentioned use cases can't be implemented as-is.

spren9er:
1. Ok, I created another pull request for adding peers attribute to `Toggle`, see #164.
2. Thanks for clarification! I was not aware that one can place interactors in the middle of marks.
This PR was not about adding syntactic nicety, but about adding functionality. It was not clear to me, that it is possible to have interactors between marks. Thus, this PR is obsolete.

PR #164, opened 2023-09-01
Add peers attribute to toggle interactor

spren9er:

Currently, a _toggle_ interactor — interacting with one mark — will filter out data of other marks in same plot, when used with cross-filter selection.

- Add `peers` attribute to _toggle_ interactor (analog to _interval_ interactors)
- Update `highlight-toggle.yaml` example, which uses now `peers` attribute under the hood to show **all** text labels when toggling (before, it showed only labels for selected bars)
- Update documentation of _toggle_ interactor

### DIFF
--- a/packages/vgplot/src/directives/interactors.js
+++ b/packages/vgplot/src/directives/interactors.js
@@ -20,16 +20,16 @@ export function toggle({ as, ...rest }) {
   return interactor(Toggle, { ...rest, selection: as });
 }
 
-export function toggleX({ as, ...rest }) {
-  return toggle({ as, ...rest, channels: ['x'] });
+export function toggleX(options) {
+  return toggle({ ...options, channels: ['x'] });
 }
 
-export function toggleY({ as, ...rest }) {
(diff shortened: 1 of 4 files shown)


packages/vgplot/src/directives/interactors.js
```diff
@@ -20,16 +20,16 @@ export function toggle({ as, ...rest }) {
   return interactor(Toggle, { ...rest, selection: as });
 }
 
-export function toggleX({ as }) {
-  return toggle({ as, channels: ['x'] });
+export function toggleX({ as, ...rest }) {
+  return toggle({ as, ...rest, channels: ['x'] });
```
jheer: We can simplify the signature and implementation to:

```js
export function toggleX(options) {
  return toggle({ ...options, channels: ['x'] });
}
```

...and similarly for the other modified `toggle` methods.

jheer (changes requested):
Everything looks good, but we should simplify the toggle function signatures.

jheer (approved):

jheer:
Thanks @spren9er!

PR #191, opened 2023-09-29
Query builder creates wrong where clause when used with disjunctions in subclauses 

spren9er:
This bug occurred while testing toggle interactor with multiple bar charts.

When selecting two bars in one bar chart and a bar in another bar chart, total number of records is not correct (using cross filter selection).

A simple Mosaic client is used for total number of records:

```javascript
export class KPI extends SvelteMosaicClient {
  // ...
  
  public query(filter = []) {
    return Query.from('flights_5m')
      .select({ total_flights: count() })
      .where(filter);
  }
  
  // ...
}
``` 

The resulting query from `mosaic-sql` looks like this:

```sql
SELECT COUNT(*)::INTEGER AS "total_flights"
FROM "flights_5m"
WHERE
  ("flight_no" IS NOT DISTINCT FROM 418) OR ("flight_no" IS NOT DISTINCT FROM 772)
  AND ("origin" IS NOT DISTINCT FROM 'DEN')
```

Above query is not correct as AND binds more tightly than OR.

This pull request fixes above issue and updates an existing tests, which also covers above case now.

### DIFF
--- a/packages/sql/src/Query.js
+++ b/packages/sql/src/Query.js
@@ -323,9 +323,11 @@ export class Query {
     // WHERE
     if (where.length) {
       const clauses = where
-        .map(String)
-        .filter(x => x)
-        .map((clause) => `(${clause})`)
+        .reduce((agg, clause) => {
+          const c = String(clause);
+          if (c) agg.push(clause.op === 'OR' ? `(${c})` : c);
+          return agg;
+        }, [])
         .join(' AND ');
       if (clauses) sql.push(`WHERE ${clauses}`);
     }
--- a/packages/sql/src/Query.js
+++ b/packages/sql/src/Query.js
@@ -322,7 +322,11 @@ export class Query {
 
     // WHERE
     if (where.length) {
-      const clauses = where.map(String).filter(x => x).join(' AND ');
+      const clauses = where
+        .map(String)
+        .filter(x => x)
+        .map((clause) => `(${clause})`)
+        .join(' AND ');
       if (clauses) sql.push(`WHERE ${clauses}`);
     }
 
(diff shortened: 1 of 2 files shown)


jheer:
Thanks! While not "elegant", it would be fine to do something like this:

```
const clauses = where.map(String).filter(x => x);
const filter = clauses.length === 1
  ? clauses[0]
  : clauses.map(c => `(${c})`).join(' AND ');
if (filter) sql.push(`WHERE ${filter}`);
```

spren9er:
This would solve it only for the case, when there is one clause.
However, when subclauses consist of a single predicate, it would still create SQL with superfluous parentheses, like

```SQL
SELECT "foo" FROM "data" WHERE (("bar" > 50)) AND (("bar" < 100))
```

I tried now with inspecting subclauses...

Not 100% sure if all possible inputs of `where` are of (or turned into) class `SQLExpression` and have an annotated `op`.

jheer:
I don't think we want to rely on `SQLExpression` here. Currently callers are allowed to pass in raw strings or even their own structured types that produce SQL predicates upon string coercion. One possibility could be to examine generated strings and add parens if those (trimmed?) strings aren't already parenthesized.

BTW I think the underlying problem exists in the `mosaic-sql` package `and` and `or` methods as well. Consider:

```js
import { and, or } from '@uwdata/mosaic-sql';
and(or('foo', 'bar'), 'baz') + '' // --> "foo" OR "bar" AND "baz" :(
```

So we need improved string generation logic there as well.

spren9er:
Ok, I see.
It seems that this requires more in depth knowledge (and overview) of the codebase of `mosaic-sql`, than I currently have.

PR #200, opened 2023-10-02
add configuration options for DuckDB class

rickiesmooth:
I need to install unsigned extensions, so when I instantiate my duckdb class I need to pass a [configuration object](https://duckdb.org/docs/sql/configuration#configuration-reference) like so:

```js
new DuckDB.Database(":memory:", {
    allow_unsigned_extensions: "true",
})
```

I also need to be able to install the arrow extension from a `custom_extension_repository` so ideally I'd be able to override the `CONFIG` as well


jheer:
Thanks!

PR #202, opened 2023-10-11
Logical operators should include parens.

jheer:
- Add parens around generated logical (AND, OR) operators when given two or more clause arguments.

This PR should serve as a fix / workaround for #191. Closes #191.


domoritz (approved):

domoritz:
Could you make a release with this or document how to make a release so I can make one?

PR #231, opened 2023-12-15
typescript: adding type definitions

calvinfo:
Hey folks,

Thanks for putting together this awesome project. I wanted to work with generated typescript types, so those will now be generated in the `dist` folders as part of the `lerna publish` step. If you'd like to verify, you can run...

```sh
$ npx lerna run prepublishOnly
$ find packages/*/dist/types
```

Happy to make other edits, let me know what you think.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,6 +16,7 @@
   "module": "src/index.js",
   "jsdelivr": "dist/mosaic-core.min.js",
   "unpkg": "dist/mosaic-core.min.js",
+  "types": "dist/types/index.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/uwdata/mosaic.git"
@@ -25,7 +26,8 @@
     "build": "node ../../esbuild.js mosaic-core",
     "lint": "eslint src test --ext .js",
     "test": "mocha 'test/**/*-test.js'",
-    "prepublishOnly": "npm run test && npm run lint && npm run build"
+    "types": "tsc",
+    "prepublishOnly": "npm run test && npm run lint && npm run build && npm run types"
   },
   "dependencies": {
     "@duckdb/duckdb-wasm": "^1.27.0",
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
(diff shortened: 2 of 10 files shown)


domoritz:
Thanks for sending the pull request. I'm curious what difference this makes over the type annotations as comments we have right now?

calvinfo:
@domoritz most of the code (as far as I can tell) doesn't have type annotations (e.g. I was trying out the `loadParquet` function). Even for the ones which do, VScode won't natively pick up the type definitions without a root type declaration. 

CleanShot 2023-12-15 at 09 54 04@2x

My thinking was that this seemed like a good catch-all to provide support for current and future types. Happy to look at other approaches for it.

domoritz:
The issue I see is that it requires compilation to get the types so as a developer of mosaic, we don't benefit from types in other internal packages. I'd rather have types directly annotated. 

The vscode issue seems orthogonal.  Is there some way we can fix the package so that the annotations are being picked up?

calvinfo:
That makes sense, for the internal code, what makes you prefer JS vs Typescript? The best option (for myself as a user) would be if it was TS from the start.

domoritz:
That's just how @jheer started. If there is no overhead for development, TS is great. I'm a huge fan and would be happy to consider moving everything over.

calvinfo:
Got it. Let me know how you two would like to proceed here. I'm happy to do a bit of work to update to TS, but otherwise having the auto-generated types seems like a very useful stopgap.

domoritz:
I'm all for ts but let @jheer weigh in.

I'm somewhat against this change right now because I don't see why it's needed. Seems like a problem with the editor rather than this package. But maybe I misunderstood why vscode isn't picking up the types from comments in your setup. 

I want to fix causes, not symptoms (with exceptions for practicality).

calvinfo:
@domoritz figured out the issue there, it comes from the `tsconfig.json` excluding `node_modules`. I can tune this to include the relevant `node_modules/@uwdata`, but it'd really be nicer from a consumption perspective if there were published types or TS-native

jheer:
Thanks! I agree with @domoritz regarding this specific PR. I would be happy to have published types! I would also be ok with porting to TS, though I will not have the bandwidth to do a conversion myself.

calvinfo:
Okay, I'll close for now and add types locally for the moment.

sandilya28:
Did anyone add types definition locally? If yes, please share it here

Issue #236, opened 2023-12-24
make data-server usable as a websocket library


mingfang:
Currently the data-server websocker standalone server is standalone.
Sometimes that's not desirable because you already have your own server, e.g. Express.
To make data-server usable in Express, we need to export `queryHandler` and `socketResponse`.

mingfang:
Here is an example use case with Express
```js
import express from 'express';
import expressWebsockets from "express-ws";
import { DuckDB, Cache, queryHandler, socketResponse } from "@uwdata/mosaic-duckdb";

const { app } = expressWebsockets(express());
const db = new DuckDB()
const CACHE_DIR = '.mosaic/cache';
const queryCache = new Cache({ dir: CACHE_DIR })
const handleQuery = queryHandler(db, queryCache)
app.ws("/duckdb-ws", (websocket, request) => {
    const res = socketResponse(websocket);
    websocket.on('message', data => handleQuery(res, data));
})

```

PR #237, opened 2023-12-24
feat: export Cache, queryHandler and socketResponse

mingfang:
fixes https://github.com/uwdata/mosaic/issues/236


jheer:
Thanks!

PR #245, opened 2023-12-29
fix: use uuid package instead of crypto for remote http jupyters

### LINKED ISSUE #244: Jupyter widget breaks on non-localhost http jupyters
If you run Jupyter either over https or on localhost, you don't get this bug. But when you're on a remote using http (or on your localhost, but aliased), the widget breaks with,

```
Uncaught (in promise) TypeError: globalThis.crypto.randomUUID is not a function
    at s (5ca03b59-285e-488c-b59a-ef82c3b57b2c:37017:33)
    at 5ca03b59-285e-488c-b59a-ef82c3b57b2c:37022:36
    at new Promise (<anonymous>)
    at Object.query (5ca03b59-285e-488c-b59a-ef82c3b57b2c:37022:14)
    at u (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8429:48)
    at a (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8409:9)
    at c (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8414:21)
    at Object.add (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8209:134)
    at Object.request (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8450:20)
    at sL.query (5ca03b59-285e-488c-b59a-ef82c3b57b2c:8527:25)
```

Some research suggests that browsers do not give access to the crypto api outside of "secure contexts" https://stackoverflow.com/questions/105034/how-do-i-create-a-guid-uuid,

> [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) is now standard on all modern browsers and JS runtimes. However, because [new browser APIs are restricted to secure contexts](https://blog.mozilla.org/security/2018/01/15/secure-contexts-everywhere/), this method is only available to pages served locally (localhost or 127.0.0.1) or over HTTPS.

and perhaps a recommended fix would be to replace the single callsite with the recommended `uuid` module from that stackoverflow answer.

frtennis1:
Closes #244

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11509,7 +11509,6 @@
       "version": "9.0.1",
       "resolved": "https://registry.npmjs.org/uuid/-/uuid-9.0.1.tgz",
       "integrity": "sha512-b+1eJOlsR9K8HJpow9Ok3fiWOWSIcIzXodvv0rQjVoOVNpWMpxf1wZNpt4y9h10odCNrqnYp1OBzRktckBe3sA==",
-      "dev": true,
       "funding": [
         "https://github.com/sponsors/broofa",
         "https://github.com/sponsors/ctavan"
@@ -12515,7 +12514,8 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@uwdata/vgplot": "^0.4.0",
-        "apache-arrow": "^13.0.0"
+        "apache-arrow": "^13.0.0",
+        "uuid": "^9.0.1"
       },
       "devDependencies": {
         "anywidget": "^0.8.0"
--- a/packages/widget/package.json
+++ b/packages/widget/package.json
@@ -17,7 +17,8 @@
   },
   "dependencies": {
     "@uwdata/vgplot": "^0.4.0",
-    "apache-arrow": "^13.0.0"
+    "apache-arrow": "^13.0.0",
+    "uuid": "^9.0.1"
   },
   "devDependencies": {
     "anywidget": "^0.8.0"
--- a/packages/widget/src/index.js
+++ b/packages/widget/src/index.js
@@ -3,6 +3,7 @@
 import { coordinator, namedPlots, parseSpec } from '@uwdata/vgplot';
 import * as arrow from 'apache-arrow';
 import './style.css';
+import { v4 as uuidv4 } from 'uuid';
 
 /**
  * @typedef Model
@@ -30,7 +31,7 @@ export async function render(view) {
    * @param {(reason?: any) => void} reject the promise reject callback
    */
   function send(query, resolve, reject) {
-    const uuid = globalThis.crypto.randomUUID();
+    const uuid = uuidv4();
 
     openQueries.set(uuid, { query, startTime: performance.now(), resolve, reject });
     view.model.send({ ...query, uuid });


domoritz (commented):
Can we use uuid only as a fallback? Or even just use a pseudo random counter? It really isn't necessary to have something complicated here as we just try to match requests.

frtennis1:
I believe the uuid package already implements the function by falling back to the native crypto implementation if it is accessible. In particular, I looked into the generated `index.js` ESM -- here is the diff that is created,

```
<     const a = Ap(e.partition).map((c) => t.maybeParam(c));
---
>     const a = $p(e.partition).map((c) => t.maybeParam(c));
30159c30159,30188
< async function ZZ(e) {
---
> let ol;
> const UZ = new Uint8Array(16);
> function qZ() {
>   if (!ol && (ol = typeof crypto < "u" && crypto.getRandomValues && crypto.getRandomValues.bind(crypto), !ol))
>     throw new Error("crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported");
>   return ol(UZ);
> }
> const be = [];
> for (let e = 0; e < 256; ++e)
>   be.push((e + 256).toString(16).slice(1));
> function jZ(e, t = 0) {
>   return be[e[t + 0]] + be[e[t + 1]] + be[e[t + 2]] + be[e[t + 3]] + "-" + be[e[t + 4]] + be[e[t + 5]] + "-" + be[e[t + 6]] + be[e[t + 7]] + "-" + be[e[t + 8]] + be[e[t + 9]] + "-" + be[e[t + 10]] + be[e[t + 11]] + be[e[t + 12]] + be[e[t + 13]] + be[e[t + 14]] + be[e[t + 15]];
> }
> const VZ = typeof crypto < "u" && crypto.randomUUID && crypto.randomUUID.bind(crypto), _2 = {
>   randomUUID: VZ
> };
> function YZ(e, t, n) {
>   if (_2.randomUUID && !t && !e)
>     return _2.randomUUID();
>   e = e || {};
>   const r = e.random || (e.rng || qZ)();
>   if (r[6] = r[6] & 15 | 64, r[8] = r[8] & 63 | 128, t) {
>     n = n || 0;
>     for (let i = 0; i < 16; ++i)
>       t[n + i] = r[i];
>     return t;
>   }
>   return jZ(r);
> }
> async function iK(e) {
30161c30190
<   const t = () => e.model.get("spec"), n = () => e.model.get("temp_indexes"), r = pn().logger(), i = /* @__PURE__ */ new Map();
---
>   const t = () => e.model.get("spec"), n = () => e.model.get("temp_indexes"), r = yn().logger(), i = /* @__PURE__ */ new Map();
30163c30192
<     const d = globalThis.crypto.randomUUID();
---
>     const d = YZ();
```

so notably the uuid package implementation gets bundled in (i.e. to my knowledge there is no perf benefit from explicitly handling the fallback.)

I defer to you on whether you prefer to do something Math.random based, but the [StackOverflow answer](https://stackoverflow.com/questions/105034/how-do-i-create-a-guid-uuid)  seemed to dissuade people from trying something like that (relevant part quoted below)

> Note: The use of any UUID generator that relies on Math.random() is strongly discouraged (including snippets featured in previous versions of this answer) for [reasons best explained here](https://bocoup.com/blog/random-numbers). TL;DR: solutions based on Math.random() do not provide good uniqueness guarantees.

Let me know what you think and I can either merge as is or implement your suggestion.

frtennis1:
I rebased the fix on top of some of the latest changes - @domoritz if you get a sec, let me know if you have thoughts on my previous message

domoritz (approved):
Thanks for the update. Looks good. I restarted the CI so hopefully that passes now (seemed to have been a random failure).

PR #247, opened 2023-12-30
feat: Generalize Nearest selection to accept array of parameters

### LINKED ISSUE #246: Let interactors update more than one parameter
Here is an example of something that I think is currently impossible: say you have a dataset like,

```
df = pd.DataFrame(itertools.product(np.arange(.2, 1.2, .01), np.arange(0, 50)), columns=['alpha', 'x'])
df['alpha'] = df['alpha'].round(2)
df['x_pow_a'] = np.power(df['x'], df['alpha'])

>>> df
      alpha   x     x_pow_a
0      0.20   0    0.000000
1      0.20   1    1.000000
...     ...  ..         ...
4998   1.19  48  100.156023
4999   1.19  49  102.643945

[5000 rows x 3 columns]
```

and want to make two line plots: one of `(x, x_pow_a)` and one of `(alpha, x_pow_a)` cross-filtered each with a Nearest selector that maps to some global filter call it `$filt$. This so far is possible. But then if you want to use the selector to make a visual representation of what is being selected (like a "ruleX" element, as in the "Stocks" Mosaic example) it becomes impossible because,

(1) You can't have two Nearest selectors on the same plot because they overwrite some event listener.
(2) You can't have a Nearest selector simultaneously update two selections (one local to the chart to draw the rule and one global for the cross-filter).

A visual representation of the plot described in words:

image

My suggestion is to change the selection member of interactors to optionally be an array. I have implemented a proof of concept for Nearest and it seems to work well. Wanted to get people's thoughts on this before I took it any further.

frtennis1:
Right now this is only implemented for Nearest (hence why I'm putting this up as a draft), but if this looks ok to people, the idea shouldn't be hard to generalize to the other interactors as well.

Closes #246

### DIFF
--- a/packages/vgplot/src/interactors/Nearest.js
+++ b/packages/vgplot/src/interactors/Nearest.js
@@ -10,7 +10,11 @@ export class Nearest {
     field
   }) {
     this.mark = mark;
-    this.selection = selection;
+    if (Array.isArray(selection)) {
+      this.selections = selection;
+    } else {
+      this.selections = [selection];
+    }
     this.clients = new Set().add(mark);
     this.channel = channel;
     this.field = field || getField(mark, [channel]);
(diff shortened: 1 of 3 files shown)


jheer:
Thanks for the PR. I'm going to close this out for now as we don't have a strategy yet for considering publishing to multiple params / selections in batch. But we could consider this in a framework-holistic way if the need proves strong enough. I opened issue #412 for this.

PR #250, opened 2024-01-01
feat: Add ability to specify arbitrary aggregations in json

### LINKED ISSUE #249: Cannot specify arbitrary aggregations in json
`vg.sql` and `vg.agg` are close cousins. They each allow arbitrary sql expressions; `vg.agg` just sets an extra aggregation property. However, best I can tell, the json spec only supports arbitrary sql expressions via `vg.sql` if you set the `"expr"` key. This makes it impossible to generate some charts via json. As an example of something I don't believe can be currently specified via json,

```
x = np.arange(100)
y = x + np.random.normal(scale=5, size=len(x))
df = pd.DataFrame({'x_val': x, 'y': y})

MosaicWidget(
    {
        "plot": [
            {
                "mark": "ruleX",
                "y1": {"agg": "avg(y) - stddev(y)/count(y)"},
                "y2": {"agg": "avg(y) + stddev(y)/count(y)"},
                "x": {"bin": "x_val"},
                "marker": "tick",
                "data": {"from": "df"},
            }
        ]
    },
    data={'df': df},
)
```

My proposal would be to special-case the "agg" key in the same way the "expr" key is special-cased.

frtennis1:
Closes #249

### DIFF
--- a/packages/vgplot/src/spec/parse-spec.js
+++ b/packages/vgplot/src/spec/parse-spec.js
@@ -1,7 +1,7 @@
 import { Param, Selection, coordinator } from '@uwdata/mosaic-core';
 import { menu, search, slider, table } from '@uwdata/mosaic-inputs';
 import {
-  sql, avg, count, max, median, min, mode, quantile, sum,
+  sql, agg, avg, count, max, median, min, mode, quantile, sum,
   row_number, rank, dense_rank, percent_rank, cume_dist, ntile,
   lag, lead, first_value, last_value, nth_value,
   dateMonth, dateMonthDay, dateDay
@@ -156,8 +156,9 @@ export class ParseContext {
   maybeTransform(value) {
     if (isObject(value)) {
       return value.expr
-        ? parseExpression(value, this)
-        : parseTransform(value, this);
+        ? parseExpression(value, this) :
+        (value.agg ? parseAggregation(value, this)
+        : parseTransform(value, this));
     }
   }
 
@@ -363,8 +364,9 @@ function parseInteractor(spec, ctx) {
   return fn(options);
 }
 
-function parseExpression(spec, ctx) {
-  const { expr, label } = spec;
+function parseExprOrAgg(spec, ctx, key, func) {
+  const { label } = spec;
+  const expr = spec[key];
   const tokens = expr.split(/(\\'|\\"|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|\$\w+)/g);
   const spans = [''];
   const exprs = [];
@@ -379,7 +381,15 @@ function parseExpression(spec, ctx) {
     }
   }
 
-  return sql(spans, ...exprs).annotate({ label });
+  return func(spans, ...exprs).annotate({ label });
+}
+
+function parseExpression(spec, ctx) {
+  return parseExprOrAgg(spec, ctx, "expr", sql);
+}
+
+function parseAggregation(spec, ctx) {
+  return parseExprOrAgg(spec, ctx, "agg", agg);
 }
 
 function parseTransform(spec, ctx) {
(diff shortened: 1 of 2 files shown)


packages/vgplot/src/spec/parse-spec.js
```diff
@@ -156,8 +156,9 @@ export class ParseContext {
   maybeTransform(value) {
     if (isObject(value)) {
       return value.expr
-        ? parseExpression(value, this)
-        : parseTransform(value, this);
+        ? parseExpression(value, this) :
```
jheer: Can we re-write using the following style?

```js
return value.expr ? parseExpression(value, this)
  : value.agg ? parseAggregation(value, this)
  : parseTransform(value, this);
```

packages/vgplot/src/spec/parse-spec.js
```diff
@@ -363,8 +364,9 @@ function parseInteractor(spec, ctx) {
   return fn(options);
 }
 
-function parseExpression(spec, ctx) {
-  const { expr, label } = spec;
+function parseExprOrAgg(spec, ctx, key, func) {
```
jheer: As a simpler alternative, perhaps we can instead leave `parseExpression` as it was, and add a new `parseAggregation` method like so:

```js
function parseAggregation(value, this) {
  return parseExpression(value, this).annotate({ aggregate: true });
}
```

We can then drop the `agg` import as well.

packages/vgplot/src/spec/to-module.js
```diff
@@ -162,8 +164,9 @@ class CodegenContext extends ParseContext {
   }
 }
 
-function parseExpression(spec, ctx) {
-  const { expr, label } = spec;
+function parseExprOrAgg(spec, ctx, key, func_name) {
```
jheer: Instead, maybe rewrite parseExpression like so:

```js
function parseExpression(spec, ctx, key = 'expr', method = 'sql') {
  const { label } = spec;
  const expr = spec[key];
  const tokens = expr.split(/(\\'|\\"|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|\$\w+)/g);
  let str = '';

  for (let i = 0; i < tokens.length; ++i) {
    const tok = tokens[i];
    if (tok.startsWith('$')) {
      str += `\${${ctx.maybeParam(tok)}}`;
    } else {
      str += tok;
    }
  }

  return `vg.${method}\`${str}\``
    + (label ? `.annotate({ label: ${JSON.stringify(label)} })` : '');
}
```

To parse an aggregation we then only need to call parseExpression with the second and third arguments both equal to `'agg'`.

packages/vgplot/src/spec/to-module.js
```diff
@@ -145,6 +145,8 @@ class CodegenContext extends ParseContext {
     if (isObject(value)) {
       return value.expr
         ? parseExpression(value, this)
+	: value.agg ?
```
jheer: Reformat as suggested above in parse-spec.js.

jheer (changes requested):
Thanks for the PR! Please see comments for requested changes.

packages/vgplot/src/spec/parse-spec.js
frtennis1: Note that parseExpression still needs the key ("expr" or "agg") so it can't be kept literally as is (but I implemented something like your suggestion in to-module.js adding the default kwarg)

frtennis1:
In testing this PR I ran into a new bug. I'm not super familiar with this code path so would appreciate thoughts. In `DataCubeIndexer.js::getIndexColumns` 

```
    switch (aggregate?.toUpperCase()) {
```
fails on 
```
TypeError: aggregate.toUpperCase is not a function
```
where `aggregate = true`. It seems there's a difference between how `vg.agg` tags aggregate (as a boolean `true`) and how `AggregateFunction` tags aggregates (as a string with the aggregate name). 

Do you know if it's safe to have the `{aggregate: True}` aggregates fall to the default case in this switch? And would you prefer tagging `vg.agg` with something like `{aggregate: "CUSTOM"}` or changing this switch statement to gracefully handle bools?

jheer (approved):

jheer:
Thank you @frtennis1! The changes in this PR look good.

Thanks also for finding that bug. I will open a separate issue where we can discuss that.

PR #260, opened 2024-01-01
feat: allow window functions in groupby by adding with statement

### LINKED ISSUE #259: Window functions do not work in query with aggregate
This is half bug half feature request. The following snippet does not currently work,

```
x = np.random.normal(scale=50, size=100)
y = x + np.random.normal(scale=5, size=len(x))
df = pd.DataFrame({'x_val': x, 'y': y})

MosaicWidget(
    {
        "plot": [
            {
                "mark": "ruleX",
                "y1": {"agg": "avg(y) - stddev(y)/sqrt(count(y))"},
                "y2": {"agg": "avg(y) + stddev(y)/sqrt(count(y))"},
                "x": {"expr": "ntile(10) over (order by x_val)"},
                "marker": "tick",
                "data": {"from": "df"},
            }
        ]
    },
    data={'df': df},
)
```

That's because the generated sql is problematic,

```
> SELECT avg(y) - stddev(y)/sqrt(count(y)) AS "y1", avg(y) + stddev(y)/sqrt(count(y)) AS "y2", ntile(10) over (order by x_val) AS "x" FROM "df" AS "source" GROUP BY "x"
Binder Error: GROUP BY clause cannot contain window functions!
```

It'd be pretty nice if the above chart snippet could work because the intent seems unambiguous to me.

A suggestion for making it work would be to modify the query generation to move the non-aggregate clause computation into a "WITH" statement where there are no restrictions on window functions, i.e.

```
WITH "__mosaicTemp" AS (SELECT *, ntile(10) over (order by x_val) AS "x" FROM "df") SELECT avg(y) - stddev(y)/sqrt(count(y)) AS "y1", avg(y) + stddev(y)/sqrt(count(y)) AS "y2", "x" FROM "__mosaicTemp" GROUP BY "x"
```

which I _think_ should not be a perf cost in cases where it's not needed because duckdb will optimize it away (?).

frtennis1:
Closes #259

### DIFF
--- a/packages/vgplot/src/marks/Density1DMark.js
+++ b/packages/vgplot/src/marks/Density1DMark.js
@@ -33,8 +33,7 @@ export class Density1DMark extends Mark {
     const [lo, hi] = this.extent = (dim === 'x' ? extentX : extentY)(this, filter);
     const bx = binField(this, dim);
     return binLinear1d(
-      markQuery(channels, table, [dim])
-        .where(filter.concat(isBetween(bx, [lo, hi]))),
+      markQuery(channels, table, filter.concat(isBetween(bx, [lo, hi])), [dim]),
       bin1d(bx, lo, hi, bins),
       this.channelField('weight') ? 'weight' : null
     );
--- a/packages/vgplot/src/marks/Mark.js
+++ b/packages/vgplot/src/marks/Mark.js
@@ -166,7 +166,8 @@ export function channelOption(c) {
 }
 
 export function markQuery(channels, table, skip = []) {
-  const q = Query.from({ source: table });
+  const q_with = Query.from({ source: table }).select('*');
+  const q = Query.from('__mosaicTemp');
   const dims = new Set;
   let aggr = false;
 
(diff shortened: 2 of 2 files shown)


jheer:
Thanks for the PR! We're going to have to think carefully whether or not we want to make this change. A few top of mind concerns that may not be reflected in the current test suite:

- Possible performance issues. For example, does DuckDB actually optimize away the subquery when not needed?
- Possible impacts on the `DataCubeIndexer`. We need to make sure that the static query analysis performed by the indexer still works as intended with changes to all generated mark queries.

I also need to think about any possible alternative approaches to achieving the desired functionality.

jheer:
Just thinking through different options, one possibility might be to include an explicit key for window expressions. So alongside `expr` (for normal expressions) and `agg` (for aggregate expressions), we might add `win` or similar for window expressions. Generally I don't like growing the surface area this way if we can avoid it, but it would enable more targeted query generation if we identify any issues with using CTEs or subqueries by default.

One smaller change might be to use a chained subquery directly, rather than a CTE using `with`. Though in theory this should only be a syntactic change to the current PR, it is possible that this could be optimized differently depending on DB internals.

frtennis1:
From having used this change locally for a bit, if nothing else, I do find that the sql queries logged to the console get uglier / harder to debug with this change. So I am particularly sympathetic to something that does not change the queries where it's not needed (even if they get optimized away fully, which in the few examples I've looked at using duckdb's `DESCRIBE` it seemed like they were).

Unfortunately, I have a poor grasp of the DataCubeIndexer so I'm not sure about the implications there -- can just lend the data point that having used this change for a few weeks, I haven't seen any wonky behavior associated with this change.

If you've thought about it more and are on-board with it I can look into adding something like your suggestion of `win`.

jheer:
Going to close out for now as we have no planned work here. Where applicable, one option is to perform window operations as part of the initial data loading, then query the results as a normal base table.

PR #274, opened 2024-01-24
mosaic-sql: typescript implementation

calvinfo:
Hey there @jheer and @domoritz. 

I took a stab at updating the `mosaic-sql` package to use typescript. Here's the implementation.

My thinking for why to start with just mosaic-sql.

- as the 'base' for more of the stuff in core, this one seemed like a good place to start
- it seemed like the most well-tested of the packages

I tried to keep all the tests the same, though there are a few key points to bring up...

- `tests/cast-test.ts`: the labels are now `avg("bar")` rather than `avg(bar)`. I _think_ this is the correct behavior, but let me know if it's not? I wasn't exactly sure what these labels were used for.
- the `package.json` for `mosaic-sql` exports `dist/index.js` as the `main` field. this allows `core` to import regular js files in testing (though better would be to use the `index.ts` down the road
- `esbuild.js` is modified so that it now takes an argument. this way the typescript generation can be configured on a package-by-package basis

What do you guys think of it? I'm not sure I have enough time to focus on some of the other packages quite yet, but wanted to get this out there as a starting point. Some of the types are quite 'loose', and I think a second pass to both simplify and unify them would be really beneficial. 

I am happy to keep this in PR form for others to pick up, but wanted to get it out there as a starting point.

-Calvin


jheer:
Thanks @calvinfo! I'll take a closer look when I have some time later on. For reference, the `label` is a text string that can be used by visualization code to generate human-friendly labels (e.g., as part of axis titles). FWIW the lack of quotation marks was intentional.

calvinfo:
Oooh got it. Okay, I can fix that. Will also fix the broken linting right now.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "noEmit": true,
+    "esModuleInterop": true,
+    "module": "CommonJS",
```
domoritz, line 8: Why commonjs? Should we use esm?

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es6",
```
domoritz: I think we can target esnext with our bundler, no?

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
```
domoritz: Why?

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "noEmit": true,
+    "esModuleInterop": true,
+    "module": "CommonJS",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "jsx": "preserve",
```
domoritz: We don't use jsx.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "incremental": true,
```
domoritz: I think we can use the default.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "downlevelIteration": true
+  },
+  "include": [
+    "**/*.ts",
```
domoritz: I usually include only the main file and let ts figure out the includes. It avoids a bunch of bugs.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+  ],
+  "exclude": [
```
domoritz: I think we can skip this.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
```
domoritz: Probably not needed if we switch the target.

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
```
domoritz, line 4: Is there an error in a lib?

packages/sql/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
```
domoritz: Do we have js?

esbuild.js
```diff
@@ -1,19 +1,20 @@
...
 
 const name = process.argv[2];
+const entrypoint = process.argv[3];
 
 function config(opt) {
   return {
-    entryPoints: ['src/index.js'],
-    target: ['esnext'],
-    format: 'esm',
+    entryPoints: [entrypoint],
```
domoritz:
```suggestion
    entryPoints: [entrypoint ?? "src/index.js"],
```

then we don't need to pass it in the build calls below

package.json
```diff
@@ -19,16 +19,19 @@
     "test": "lerna run test",
     "server": "nodemon packages/duckdb/bin/run-server.js",
     "dev": "vite",
-    "release": "npm run test && npm run lint && lerna publish"
+    "release": "npm run build && npm run test && npm run lint && lerna publish"
```
domoritz, line 22: Could we run tests with ts-node so we don't need to build?

packages/sql/package.json
```diff
@@ -10,19 +10,20 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/uwdata/mosaic.git"
   },
   "scripts": {
-    "prebuild": "rimraf dist && mkdir dist",
-    "build": "node ../../esbuild.js mosaic-sql",
-    "lint": "eslint src test --ext .js",
-    "test": "mocha 'test/**/*-test.js'",
+    "prebuild": "rimraf dist && mkdir dist && tsc",
```
domoritz: Why run tsc here? Doesn't esbuild work with ts?

package.json
domoritz: Looks like we use ts-mocha so this should not be needed, right?

domoritz (changes requested):
Very nice. Looking forward to getting this in.

packages/sql/tsconfig.json
calvinfo: Yeah, good call! Can make changes there

packages/sql/tsconfig.json
calvinfo: Can remove this

esbuild.js
calvinfo: ah great idea

packages/sql/package.json
calvinfo: good call, this is from an earlier version, will remove!

package.json
calvinfo: @domoritz the reason I'd added was because the `mosaic-core` package relies on this library for testing. my thinking was that running the `build` first would ensure that mosaic-core is referencing the `dist` file, but maybe that's not right?

calvinfo:
One other question for both of you... in terms of linting, I think we could either do...

1) add @typescript-eslint to the .eslint config for this repo. using the recommend plugins means getting rid of all `any` types (or ignoring them manually). I've considered adding an `Expression` type which would be `type Expression = SQLExpression | Ref | Param | string;` which I think would take care of the majority of cases. I want to ensure I'm understanding the typing properly... does that seem right?
2) remove linting for this package for the moment, since the typescript compilation will catch any obvious mistakes.

packages/sql/tsconfig.json
calvinfo: not now, can remove :)

calvinfo:
Also just followed up with a cleaned `tsconfig`. Let me know what you think.

package.json
domoritz: Oh I see. Maybe we should run the tests everywhere with ts-mocha so that we can use ts or js. I am worried that we compiled and raw code gets out of sync and devs get confused.

domoritz:
eslint does more than tsc so I think it's worth having both.

calvinfo:
Okay, `label` now works as desired (regarding the above comment) and `npm run lint` is fixed too. For the moment it warns on `any` types, but I think we can narrow those down over time. I could definitely benefit from both your thoughts on the typing side when it comes to making them much tighter and more intuitive.

package.json
calvinfo: yeah, that's a solid idea. I think it'd be easy enough to add too

packages/sql/package.json
```diff
@@ -10,8 +10,9 @@
   "license": "BSD-3-Clause",
   "author": "Jeffrey Heer (http://idl.cs.washington.edu)",
   "type": "module",
-  "main": "src/index.js",
-  "module": "src/index.js",
+  "types": "dist/mosaic-sql.d.ts",
+  "main": "dist/mosaic-sql.js",
+  "module": "src/index.ts",
```
domoritz, line 15: Is this ok? I thought we would need to always point to js files here.

packages/sql/src/aggregates.ts
```diff
@@ -0,0 +1,161 @@
+import { SQLExpression, parseSQL, sql } from "./expression";
```
domoritz, line 1: Is that any way you can mark the files as moved? Maybe rename first and commit and then change the content. It would make it easier to see the differences.

packages/sql/package.json
calvinfo: You're right, will address.

packages/sql/src/aggregates.ts
calvinfo: I agree, I'm a little time-strapped for the next few days but will see if I can do some sort of rebase + fix commit log to make it easier to compare the diff.

domoritz: Maybe the easiest is to make a new branch, do the rename, commit, paste in the new files, commit again. Replace this branch.

However, now that I think about it, I wonder whether renames will be tracked correctly when we squash.

domoritz: I'm kind of surprised that GitHub doesn't show the files as renamed. Are there so many changes?

calvinfo: @domoritz I have a cleaned up version here: https://github.com/uwdata/mosaic/compare/main...calvinfo:calvinfo/sql/ts-2?expand=1. Lmk what you think. I'm happy to make a PR to that if you think it looks good.

domoritz: Thanks. It still looks like some of the files are not tracked as renamed like packages/sql/src/datetime.js.

calvinfo: Yeah, not sure what to tell you there. I separated it out this time into two explicit rename commits (no functionality change)

https://github.com/uwdata/mosaic/commit/37a2b9259caa06bd44ee673785e675d5ef75b65e
https://github.com/uwdata/mosaic/commit/bc4d19ce05ce281f15a8e4c7c01f4b0de19581a8

In the case of `datetime.js`, I think a high enough percentage of the lines changed to trigger a "re-write" vs a "rename" in the github UI.

calvinfo:
Tried using this in my own packages, and I think there are definitely some quirks when packaging for isomorphic use (and I think I got some of the typing wrong even though the tests pass). Definitely don't merge yet unless you want to take a better stab at updating the package.json

packages/sql/src/aggregates.ts
domoritz: Thanks for separating the changes. Once this works, let's merge two commits. One with all the renames, and one with the code changes. That should be good!

domoritz:
@calvinfo I'd love to merge this to avoid future conflicts. I think it's okay if the types are not fully correct yet but we need clean support for having js and ts packages and an easy way to have changes to the ts sources propagate. Could I help you push this over the finish line?

calvinfo:
@domoritz please! Any help is super appreciated, feel free to tweak this and run with it. I can put together a list of issues I've noticed trying to pull in this package elsewhere. I'm a little swamped over the next week to make changes and don't want to block the new refactors y'all are trying to push through.

eknowles:
Is it intentional to delete/create new files rather than rename with git? I feel there's a bit of git history being lost there.

domoritz:
> Is it intentional to delete/create new files rather than rename with git? I feel there's a bit of git history being lost there.

https://github.com/uwdata/mosaic/pull/274#discussion_r1467601173

domoritz:
Closing for now since we have https://github.com/uwdata/mosaic/pull/355 with much better ts support.

calvinfo:
Thanks @domoritz! Excited to give it a shot when I pick up the project again.

willium:
I put [these partial declarations](https://gist.github.com/willium/15e61146761c69c7eb50d01dbe005852) together for my own needs while waiting for #355 to land

PR #280, opened 2024-01-31
Overhaul: new spec parser, refactor packages, add spatial support

jheer:
This PR provides multiple major updates.

**Update declarative spec parser and code generator**
- **Breaking**: Remove spec parser and code generator from `vgplot`. The `parseSpec` method is no longer exported by vgplot. Instead, use the new methods from the `@uwdata/mosaic-spec` package.
- **Breaking**: Remove support for client-side data loading. Previously this was used to load GeoJSON and TopoJSON data. The DuckDB spatial extension is now used to load and manage such data directly in the database.
- **Breaking**: The `expr` construct has been removed from declarative specs, instead use `sql`. This change better aligns the specs with the underlying vgplot API (where the `sql` tagged template is used).
- Add `spec` package with updated spec parser and AST framework.
- Add extension loading to spec processing.

**Refactor packages**
- Add `plot` package for plot, mark, and interactor component classes. 
- Re-focus `vgplot` package on exported API for Mosaic use.
- Add top-level `specs` folder for shared access to example specs.

**Add spatial support**
- Add support and examples for the DuckDB `spatial` extension.
- Add `loadSpatial` helper and spatial methods (e.g., `centroidX`, `centroidY`) to `mosaic-sql`.
- Remove client-side GeoJSON/TopoJSON loading, route through DuckDB instead.
- Add mapping and spatial analysis examples.

**Other changes**
- Add optional vgplot API contexts to permit multiple coordinators or namedPlot sets on the same page.
- Add sql `loadExtension()` utility.
- Add array support to `Coordinator.exec()`.
- Fix vgplot `margin()` attribute directive.
- Fix line/area marks: do not apply M4 optimization to unoriented marks.
- Replace `selections` with `params` in widget, include all params, and index params by name.


packages/widget/src/index.js
```diff
@@ -125,3 +126,9 @@ export async function render(view) {
     reset();
   };
 }
+
+async function instantiateSpec(spec) {
+  const ast = parseSpec(spec);
+  const dom = await astToDOM(ast);
```
jheer: @domoritz The new `astToDOM()` method returns an object that includes the DOM `element` but also a `Map` of name-Param pairs (including all Param and Selection instances in the spec). This might be handy for better returning Param and Selection values from the browser back to the Jupyter kernel. Feel free to push updates to the widget package on this branch if you want to incorporate this!

domoritz: Awesome. Done in f03de439ef9e37.

PR #287, opened 2024-02-02
feat: Mosaic server in Python

domoritz:
- [x] working websocket
- [x] create and load budles
- [x] basic unit tests 
- [x] http
- [x] cache to disk


jheer:
This is looking great. Thank you!!

Would it help to name the package folder “python-server” or “python-duckdb-server” to better indicate what the package provides?

domoritz:
Totally unscientific but the Python server seems to perform pretty well. Here are the logs of me brushing back and forth a bit. I tried a version of the python server with https://websockets.readthedocs.io/ and it wasn't as fast as the one with socketify but again I didn't do any clean experiments. 

Node server

Screenshot 2024-02-03 at 10 36 06

Python server

Screenshot 2024-02-03 at 10 36 40

domoritz:
> Would it help to name the package folder “python-server” or “python-duckdb-server” to better indicate what the package provides?

I prefer not to have the programming language in the package name. But we could name it `duckdb-server`. However, we may extend what the server can do at some point so it might make sense to keep it generic. 

I was also thinking we would replace the existing server completely. Is there a strong reason to keep the node sever?

jheer:
> I prefer not to have the programming language in the package name. But we could name it `duckdb-server`. However, we may extend what the server can do at some point so it might make sense to keep it generic.

I'd rather error on the side of being helpfully specific, so maybe `duckdb-server`?

> I was also thinking we would replace the existing server completely. Is there a strong reason to keep the node server?

I definitely want to keep the DuckDB node client class in the `mosaic-duckdb` package. I'm already using it in other experiments (for example, to instantiate a running Coordinator in Node that talks directly to a DuckDB node instance). It's less clear to me what we want to do with the server in the long term. I think we should keep it for now and reassess once we see (1) experiences with a Python-based server, and (2) what fixes/issues we see with DuckDB+Arrow+Node in DuckDB v10+.

domoritz:
@jheer The server should be feature complete now. There are a bunch of more things we can do but I think this is a good starting point for more experiments.

PR #288, opened 2024-02-02
chore: clean up widget formatting

domoritz:
We use to format two directories up which isn't correct. I also switched linting to eslint so we just lint the js.


domoritz:
Ups, forgot that I already sent https://github.com/uwdata/mosaic/pull/284

PR #290, opened 2024-02-07
docs: Add link for "M4 Optimization"

manzt:
Found this link somewhere else in the docs.


jheer:
Thanks!

domoritz:
[A comment in the code](https://github.com/uwdata/mosaic/blob/5081d7b787effe8142b53cbd628e4807c92f3d42/packages/vgplot/src/marks/ConnectedMark.js#L41-L47) references http://www.vldb.org/pvldb/vol7/p797-jugel.pdf and https://arxiv.org/pdf/2306.03714.pdf. I think the observable document is a good summary, though.

PR #293, opened 2024-02-13
Update server and widget

jheer:
- Update server and widget to use DuckDB 9.2.
- Fix server to pass exceptions to logger.


domoritz (approved):

domoritz:
Not sure why the CI isn't running but this looks good.

PR #299, opened 2024-02-18
feat: allow Menu to select a sql expression

### LINKED ISSUE #298: Allow for Menu to select a sql expression
The [Presidential Opinion example](https://uwdata.github.io/mosaic/examples/presidential-opinion.html#specification) has something that looks a lot like "select a column/sql-expression via this menu and it gets plotted on the chart." This is a really neat feature to have.

However, the way it is implemented (where the Parameter takes on a value of 1 or -1 and a sql expression does a multiplication with the column values) doesn't generalize well to more than two options.

It is already possible to get the "choose between many options" behavior in javascript by setting the options of the menu to the object produced by `vg.sql`. However, to my knowledge, this is not possible to do via a spec. In a spec, the best you can do is make the options strings, but the strings get quoted as literals, and are not interpreted as column names.

I thought of two ways this could be implemented to work with specs:
1. The Menu could special-case some key-namespace to interpret as sql (probably in the form of objects that look like `{sql: "y_column"}`)
2. The spec parser could have some syntax for producing a `vg.sql` object. I went down this route at first, but it's relatively tricky to get it to work because the menu options are parsed via the completely generic `parseOptions` creating a `LiteralNode` with the list of options. So to get this feature would require special-casing in code that is fully generic right now (and the `LiteralNode` which right now holds an opaque list would have to be further decomposed into nodes).

frtennis1:
Closes #298 

Example usage of this feature,

```
df = pd.DataFrame({"x_val": np.arange(20)}).assign(
    y_val=lambda df: df["x_val"] * df["x_val"],
    noise=np.random.normal(scale=5, size=20),
)

d = {
    "vconcat": [
        {
            "input": "menu",
            "label": 'yCol',
            "options": [
                {"label": "Pure signal", "value": {"sql": "y_val"}},
                {"label": "Small noise", "value": {"sql": "y_val+noise"}},
                {"label": "Medium noise", "value": {"sql": "y_val+2*noise"}},
                {"label": "Large noise", "value": {"sql": "y_val+5*noise"}},
            ],
            "as": "$yCol",
        },
        {
            "plot": [
                {"mark": "lineY", "x": "x_val", "y": {"sql": "$yCol"}, "data": {"from": "df"}}
            ],
        },
    ],
}


MosaicWidget(d, data={'df': df})
```

image

image

### DIFF
--- a/packages/inputs/src/Menu.js
+++ b/packages/inputs/src/Menu.js
@@ -1,10 +1,21 @@
 import { MosaicClient, isParam, isSelection } from '@uwdata/mosaic-core';
-import { Query, eq, literal } from '@uwdata/mosaic-sql';
+import { Query, eq, literal, sql as sqlFunc} from '@uwdata/mosaic-sql';
 import { input } from './input.js';
 
 const isObject = v => {
   return v && typeof v === 'object' && !Array.isArray(v);
 };
+
+const parseOption = value => {
+  if (!isObject(value)) {
+    return { value };
+  }
+  const { value: { sql }, ...options } = value;
+  if (sql === undefined) {
+    return value;
+  }
+  return {value: sqlFunc`${sql}`, ...options}
+}
 
 export const menu = options => input(Menu, options);
 
@@ -36,7 +47,7 @@ export class Menu extends MosaicClient {
 
     this.select = document.createElement('select');
     if (options) {
-      this.data = options.map(value => isObject(value) ? value : { value });
+      this.data = options.map(parseOption);
       this.update();
     }
     value = value ?? this.selection?.value ?? this.data?.[0]?.value;


jheer:
We don't plan to support expressions as menu options at the moment. While I'm sure it's just meant to be illustrative, the specific example above can be achieved by having the input provide a numerical value (noise factor) and then using it that in a sql expression elsewhere via a param reference.

I can also imagine people wanting to have string-valued menu options that correspond to database column names. As of this comment, this is not supported in declarative specs using params because a param string value will be interpreted as a string literal when referenced in downstream SQL expressions. Feel free to open a new issue to request some mechanism to instead interpret strings as column names or expressions (possibly as a type hint as part of a param definition?).

PR #300, opened 2024-02-20
Respect scale transforms when binning.

jheer:
This PR adds support for non-linear scales across binning methods, both within gridded marks like `raster`, and for the `bin` transform. This PR also renames some of the options for density/contour/raster marks and adjusts their defaults. This simplifies some naming and usage, and better aligns with the underlying Observable Plot library.

- _binType_ is now _interpolate_. The `"normal"` option value is now `"none"`.
- _binWidth_ is now _pixelSize_.
- _binsX_ is now _width_.
- _binsY_ is now _height_.

Changes:

- **Breaking**: Rename options for density/contour/raster marks, change defaults.
- Add support for non-linear scales (log, symlog, sqrt) to pixel-level binning methods.
- Add support for non-linear scales (log, symlog, sqrt) to the `bin` transform.
- Add SQL scale transform utilities to the `mosaic-sql` package.
- Add richer option support to `raster` mark, including gridded ordinal data alongside numeric data.
- Add `heatmap` mark, a `raster` mark with convenient defaults for density estimation.
- Update mark implementations to use improved binning.
- Update `density1d` example spec to use a log-scaled domain.
- Update examples, including new Observable latency example.
- Update documentation.
- Fix range transformation bug in `DataCubeIndexer`.
- Fix filter range bug in `DensityLineMark`.
- Fix M4 optimization bug in `ConnectedMark` (line and area).


packages/plot/src/marks/ConnectedMark.js
```diff
@@ -47,13 +40,13 @@ export class ConnectedMark extends Mark {
  * an efficient version with a single scan and the aggregate function
  * argmin and argmax, following https://arxiv.org/pdf/2306.03714.pdf.
  */
-function m4(input, bx, x, y, lo, hi, width, cols = []) {
-  const bins = sql`FLOOR(${width / (hi - lo)}::DOUBLE * (${bx} - ${+lo}::DOUBLE))::INTEGER`;
+function m4(input, bin, x, y, cols = []) {
+  const pixel = sql`FLOOR(${bin})::INTEGER`;
```
domoritz, line 50: I know we want to have ints in the end since dealing with decimal is annoying but I think we can omit the integer casting for every sub expression at some point and let the db figure out what the best representation of the types is.

domoritz (approved):
Great to have better scale support.

packages/plot/src/marks/ConnectedMark.js
jheer: In many of these cases the integer casting is helpful. We are binning to a space of pixels, so we have some bounds on the numbers produced. If DuckDB was smarter about generating more compact representations (like small ints) that would be great. But it typically defaults to using long ints which are bigger than we need and also get mapped to BigInts by default by the Arrow lib, which also complicates things. So I suspect we'll want to keep these casts for a while.

PR #313, opened 2024-03-04
Observable Plot attribution

mbostock:
Hi @jheer! We noticed that you’re including portions of Observable Plot’s source code in Mosaic. You’re welcome to do so, but you must include a copy of our LICENSE. (This PR combines it into your top-level LICENSE, but you could put the license within the relevant source files if you prefer, as you do [for jstat](https://github.com/uwdata/mosaic/blob/9e788e6dc5241fa1c54967a25fd9599f97da1a41/packages/plot/src/marks/util/stats.js#L4).) We also recommend including links back to the original source so that it’s easier to check if things have changed.

Note that you can import the spatial interpolators from Plot directly if you don’t want to copy the source (but it appears you’ve changed the signatures slightly so perhaps that won’t work for you):

https://github.com/observablehq/plot/blob/41a63e372453d2f95e7a046839dfd245d21e7660/src/index.js#L25

And we improved the `isColor` routine to support CSS Color Module Level 5 in https://github.com/observablehq/plot/pull/1454:

https://github.com/observablehq/plot/blob/41a63e372453d2f95e7a046839dfd245d21e7660/src/options.js#L464-L477


jheer:
Thank you @mbostock!!

jheer:
In general I'd prefer reusing Plot's interpolators, but wanted to pass in a pre-allocated raster array (hence the signature change). Also FWIW, I noted that in the original implementation the barycentric interpolator can still leave some pixels as `NaN` (after an initial `NaN` fill). This became apparent as floating point issues in downstream code. I'm planning to file an issue for Plot but probably won't have time to put together a standalone repro for at least a week or so.

PR #322, opened 2024-03-10
feat: publish python packages together with node packages

### LINKED ISSUE #319: Heatmap not working
The heatmap examples are not working in jupyter. I believe it is because heatmap.js is missing from the marks folder.

BTW. Thanks for this great library. I'm looking forward to trying this out on some large time series datasets.

domoritz:
Also synchronizes the package versions. To successfully publish packages, make sure that you have a working Python setup and that you are authenticated with pypi.

Fixes #319


jheer:
@domoritz I merged this and tried it out but unfortunately it's not fully working. The version numbers appear to update correctly, but lerna skips the `publish` step -- I think because the Python package.json files include the "private" flag? So I still had to manually publish them.

PR #332, opened 2024-03-20
feat: rust server

domoritz:
* DuckDB server written in Rust.
  * Supports http/1, http/2, and websockets.
  * Caching and bundling support.
* Fixes bundle logic in demo page.
* Rewrite priority queue as a class.
* Uses DuckDB-rs 1.0 which was released recently.

After this is merged, my plan is to use the http/2 support for parallel queries from the browser.


domoritz:
I looked into using rust from python and it's pretty straight forward but probably doesn't make as much sense for the widget since it's so simple already and I don't know how we would allow duckdb to query pandas through rust.

jheer (approved):

jheer:
Overall this looks good to me. However, I noticed that the Rust tests appear to take a very long time. It's took over 20 minutes, apparently due to lots of download and compilation steps. This does not make for a great developer experience here on GitHub. Any ideas for how to address this?

domoritz:
It looks like we are compiling twice right now. I'll check what's up with that. I might also be able to add caching to reduce build times.

domoritz:
Took some experimentation but got it down to 37s.

mhkeller:
Thanks again for your work in implementing this!

domoritz:
You're welcome. How are you using it?

mhkeller:
I've been working on an app that queries postgres and, among other things, allows for custom mosaic visualizations off of the query result.

PR #334, opened 2024-03-20
feat: use columnar data within vgplot

domoritz:
This defers chasing the arrow tables to arrays of objects and optimizes some of the computation to use columnar processing (e.g. in nearest). 

There are possibly more things to optimize (e.g. Plot itself https://github.com/observablehq/plot/issues/191) and we can remove some of the special handling when Arrow supports Decimals better.


packages/plot/src/marks/util/grid.js
```diff
@@ -1,13 +1,12 @@
...
 import {
-  convertArrowArrayType,
   convertArrowColumn,
   isArrowTable
 } from '@uwdata/mosaic-core';
 
 function arrayType(values, name = 'density') {
   return isArrowTable(values)
-    ? convertArrowArrayType(values.getChild(name).type)
+    ? values.getChild(name).type.ArrayType
```
domoritz: This is not fully equivalent since we previously used Float64Array in many cases and now use Arrow's logic for array types.

packages/core/src/util/convert-arrow.js
```diff
@@ -16,19 +12,12 @@ export function isArrowTable(values) {
...
-export function convertArrowArrayType(type) {
-  switch (type.typeId) {
-    case INTEGER:
-    case FLOAT:
-    case DECIMAL:
-      return Float64Array;
-    default:
-      return Array;
-  }
+export function needsIntervention(type) {
```
domoritz: Better naming suggestions welcome. Maybe `needsSpecialHandling`?

packages/plot/src/marks/GeoMark.js
```diff
@@ -22,12 +23,17 @@ export class GeoMark extends Mark {
 
     // parse GeoJSON strings to JSON objects
     const geom = this.channelField('geometry')?.as;
+
+    if (isArrowTable(data)) {
+      this.data = data.toArray();
```
domoritz: Needed since we are modifying the array. May need some more thinking to change this if we want to work with arrow.

packages/plot/src/marks/RegressionMark.js
```diff
@@ -44,7 +44,7 @@ export class RegressionMark extends Mark {
   }
 
   queryResult(data) {
-    this.modelFit = toDataArray(data);
+    this.modelFit = toObjectArray(data);
```
domoritz: We could rewrite the logic to handle arrow but doesn't seem necessary since the data is so small.

domoritz:
Hmm, this breaks the observable latency example. Need to check what's up.

jheer:
I like the goal here, but some of these changes probably won’t help much (and others will hurt performance) until Plot also supports optimized columnar data.

Using Arrow’s automatic row object generation (via Proxy) has a non-trivial overhead cost. That’s why I have our own faster object generation method (arrowToObjects) to avoid costly unpacking within Plot. So for now we should still use that even if no type conversions are needed.

mbostock:
Just chiming in here — Plot already supports optimized columnar data, it just doesn’t support _shorthand_ for that data. I added an example to https://github.com/observablehq/plot/issues/191#issuecomment-2010986851 that shows how to pass in columnar data. Please let me know if we can help!

domoritz:
> Just chiming in here — Plot already supports optimized columnar data, it just doesn’t support shorthand for that data. I added an example to https://github.com/observablehq/plot/issues/191#issuecomment-2010986851 that shows how to pass in columnar data. Please let me know if we can help!

Something like [`62e72fa` (#334)](https://github.com/uwdata/mosaic/pull/334/commits/62e72fa53114deaea20ced51cfbc11296eaafb0e#diff-f9b3058a7c646c6a45376588bb428e8ad1244d60c12e20c2a5a674711a07de08R29-R43) ?

mbostock:
That looks right to me, @domoritz! 👍

packages/plot/src/interactors/Toggle.js
```diff
@@ -52,8 +58,9 @@ export class Toggle {
   init(svg, selector, accessor) {
     const { mark, channels, selection } = this;
     const { data } = mark;
+    const method = isArrowTable(data) ? 'get' : 'at';
```
domoritz: This will not be needed after https://github.com/apache/arrow/pull/40712.

jheer (approved):

packages/core/src/util/convert-arrow.js
domoritz: Not needed anymore.

PR #336, opened 2024-03-21
Add initial projection support for interactors.

jheer:
This PR explores possible brushing support over projections instead of scales. It requires changes to Observable Plot to expose projections and sufficient information to perform inversion. This initial version supports brushing over projections that maintain an orthogonal separation between longitude (e.g., -> x) and latitude (e.g., -> y), including planar (identity, reflect-y), equirectangular, and mercator projections.

The code assumes that Plot projections are instead constructed in Plot's `projection.js` like so:

```js
return {
  width: dx,
  height: dy,
  offset: [marginLeft + insetLeft, marginTop + insetTop],
  translate: [tx, ty],
  scale: k,
  invert: projection.invert,
  stream: (s) => projection.stream(transform.stream(clip(s)))
};
```

And then later exposed in `plot.js` similar to this:

```js
figure.projection = () => context.projection;
```

If/when Plot issue https://github.com/observablehq/plot/issues/1191 is addressed, we can visit this!


Fil:
We can't send an inverse projection function to the backend, but instead of sending separate bounds for lon and for lat, can we send a ST_geometry, and query with a filter `ST_Within(datum, geometry)`? Or would that be too slow anyway?

The geometry would have to correspond to "everything inside the brush, in source coordinates", which is (as a first approximation [1]) the inverse projection of the brush rectangle.

[1] It's more complicated than that since you might brush over points that are outside of the sphere, but that's a secondary issue.

jheer:
Thanks @Fil. For the kind of brushing you're describing, we could probably get by processing the selected geometry in browser and creating a predicate over ids (that is, selecting by item rather than over intervals). That's on our TODO list.

This PR is focused on exploring support for interval selections. If we project in database perhaps we can just get by with "implicit" planar projections (where Plot still generates and exposes x/y scales). Nevertheless, this PR tries to push further for cases where explicit projections are defined and Plot does not expose any scales.

PR #338, opened 2024-03-21
Use columnar data within vgplot.

jheer:
This builds on the earlier Arrow PR to better leverage columnar data. Rather than propagate Arrow tables throughout the code, we convert both input arrays-of-objects and Arrow tables to a simple, internally standardized columnar data representation. We then use this to provide columnar data directly to Plot to minimize data copying.

For Arrow data, our columnar representation will reuse the same in-memory arrays where possible (e.g., for ints and numbers), otherwise it performs a one-time conversion (e.g., to extract strings, dates, decimals, etc.).

By having a single internal standard, we also avoid special casing for Arrow vs. non-Arrow data in many places.


domoritz:
Looks good. The standard layout makes it easier than using arrow everywhere. The conversions here mostly are really fast.

PR #339, opened 2024-03-22
fix: only activate client in intervals when the pointer is up

### LINKED ISSUE #337: Disable activation of charts while the mouse is pressed
When moving a brush, one might move the cursor over another chart which activates it, which then creates indexes and lag. We should disable activation while the cursor is pressed. 

https://github.com/uwdata/mosaic/assets/589034/0d0ce691-bd4e-4e7d-9c22-06517a0f94a8

domoritz:
fixes #337

The cursor isn't correct while you hover over another chart and we also don't activate a client when the pointer goes up over a chart but those are just small things that would require more substantial changes.

### DIFF
--- a/packages/plot/src/interactors/util/pointerIsDown.js
+++ b/packages/plot/src/interactors/util/pointerIsDown.js
@@ -1,4 +1,6 @@
 export let pointerIsDown = false;
 
-document.body.addEventListener('pointerdown', () => pointerIsDown = true);
-document.body.addEventListener('pointerup', () => pointerIsDown = false);
+if (typeof document !== 'undefined') {
+  document.body.addEventListener('pointerdown', () => pointerIsDown = true);
+  document.body.addEventListener('pointerup', () => pointerIsDown = false);
+}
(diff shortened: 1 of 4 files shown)


packages/plot/src/interactors/util/pointerIsDown.js
```diff
@@ -0,0 +1,6 @@
+export let pointerIsDown = false;
+
+if (typeof document !== 'undefined') {
```
jheer, line 3: Hmm. On my TODO list is to make Mosaic work with server-side rendering, where we can pass in a document as part of the API context. I'd like to avoid random references that assume `document` is defined in the global scope only. I'll think about alternatives...

MouseEvent is way easier as it includes which buttons are down.

jheer:
I'm closing out this PR in favor of #341, which is simpler (checks mouse buttons only) and covers all interactors. At worse, it provides a clean partial solution. However, #341 may also be a pretty complete solution as it is less common for touch to range to other parts of the screen (compared to mouse dragging) and touch events have different rules around target capture/release which may work in our favor.

Issue #345, opened 2024-03-31
PanZoom not working with RasterMark


jonrmayer:
Very simple example - I did try all types of configuration - but the following would not work:
hconcat: 
  - vconcat:
    - plot:
      - mark: frame
      - { mark: raster, data: { from: summary_data}, x: px,y: py, fill: density  } 
      - { select: panZoom }

jonrmayer:
Apologies, I should have mentioned that the x and y scales were changing but the raster image itself remained static.

jheer:
Thanks for the issue! Right now you can zoom a raster display via overview + detail interactions (as in the Line Density chart example), but the PanZoom interactor is not well supported (sorry!). I'll flag this as both a documentation and enhancement issue for now.

jonrmayer:
Thank you!. I am a geospatial data guy - hence the obsession with panning & zooming ;-)  This is a great little library with a lot of current/future potential and I am putting together an demo for internal consumption. Hopefully, this will encourage some of my colleagues ( senior JS developers ) to get involved. This might take some time. Thanks again for your research and work on this project.

jheer:
Actually, upon additional inspection I discovered a bug. Your spec _should_ work (though the performance of raster pan/zoom will depend on the data set size, raster size, and database performance - and so range from very smooth to very chunky. I've update this issue to reflect the bug and will try to get a fix in place soon.

jheer:
Fixed by [d3e521a](https://github.com/uwdata/mosaic/commit/d3e521a69876732609a8119959b8078392dd5859).

PR #353, opened 2024-04-01
docs: add types for setPlot and plot attributes

domoritz:
I was hoping it would help detect things like #345 but we would need even stronger types.


jheer:
LGTM. Should we merge into #355 first?

PR #355, opened 2024-04-01
Add improved TypeScript support.

jheer:
This PR starts to provide better TypeScript support, but stops well short of complete typings. Sufficient changes were made to prevent the widget test from failing on the latest TS version. Future todos include more coverage of jsdoc-based types in the core Mosaic packages.

**Changelog**:
- Add and publish TypeScript types for declarative JSON specs.
- Update some jsdoc typings throughout Mosaic packages.
- Add test TypeScript files for all specs, run `tsc` on them as part of `@uwdata/mosaic-spec` tests.
- Refactor `handleParam` mark utility for better clarity and type-friendliness.
- Bump to latest version of TypeScript.


packages/core/src/Coordinator.js
```diff
@@ -8,8 +8,7 @@ let _instance;
 
 /**
  * Set or retrieve the coordinator instance.
- *
- * @param {Coordinator} instance the coordinator instance to set
+ * @param {Coordinator} [instance] the coordinator instance to set
```
domoritz, line 11: What do the `[]` do here?

domoritz: Oh, it means optional.

jheer: That's jsdoc syntax to indicate an optional argument, akin to `instance?` in proper TypeScript. Without that, I believe calling the method as a getter will fail type checks.

jheer:
@domoritz FYI it looks like your type updates now cause type check failures, probably due to interactions with my changes?

packages/core/src/Coordinator.js
domoritz: Yeah, I noticed that when I tried `?` for an optional argument. I'm not quite used to jsdocs types yet.

jheer:
> I fixed the issues in my changes so they pass now. I also just enabled type checks for the specs in js. Feel free disable that again but it looks like we have some discrepancies that this finds. E.g. `SpecNode` is a thing in js but not in d.ts.

Yes, thank you! I've been focusing first on types for the declarative spec itself, and then will look more at the AST node classes. My ideal would likely be to use jsdoc for the classes and `d.ts` for the JSON, but I'm not sure such a hybrid typing approach is supported out of the box (and I haven't had time to dig into that yet).

domoritz:
Disabled for now. Might actually be worth switching to Typescript at this point so we type check the implementation against the spec as well. But that can happen later.

domoritz (approved):
having both .d.ts and JSDocs with types is confusing and seems to break VSCode at least (and tsk if we type check the js files). The problem is that it doesn't know whether to go to the js file or the d.ts file to get the types for a method etc. It would maybe be easier to either go all in on ts or to move all types into the d.ts files.

Having said that, this is still a good step forwards so I am supportive of merging it.

domoritz:
What's your main concern with going all ts? Is it that it would mean that we have to explicitly opt out of typing rather than opt in for all code or that we would need to compile any dependency before we can use it in a dependent package?

jheer:
I'm not dead set against going all TS at some point. It's just that it's a big porting lift and not a top priority at the moment.

PR #356, opened 2024-04-03
Use alternative typing approach.

jheer:
Instead of using static `.d.ts` files alongside `.js`, this PR uses `.ts` files for pure type definitions for the Mosaic JSON spec, but otherwise keeps to jsdoc within JS. This requires a build step where we run `tsc` to process both the JS and TS files to produce `.d.ts` types. However, the resulting code organization is cleaner and avoids duplicate and/or incorrect content across parallel `.js` and static `.d.ts` files.


jheer:
@domoritz I think this is a better starting point for typings. If you agree, please approve. Then I will merge into the existing types branch to update our approach.

domoritz (approved):
That works better.

PR #414, opened 2024-05-29
refactor: rewrite QueryManager to be a class

domoritz:


packages/core/src/Coordinator.js
```diff
@@ -24,8 +24,9 @@ export class Coordinator {
   constructor(db = socketConnector(), options = {}) {
     const {
       logger = console,
-      manager = QueryManager()
+      manager = new QueryManager()
     } = options;
+    /** @type QueryManager */
```
jheer: I don't mind changing to a class, but this changes the interface and its types, as "internal" properties and methods like `next()` and `enqueue()` are now part of the exposed instance and the class type signature. If someone wants to create their own query manager (unlikely I guess, but still...) they now have to deal with type compatibility issues beyond the minimum necessary interface.

domoritz: Makes sense. In the future, we could define a proper Query Manager Interface that our default implantation implements. For now, I removed this line.

Issue #399, opened 2024-05-23
Support Databases other than DuckDB


domoritz:
Create new connectors, for example to Postgres, Clickhouse, or Snowflake. Identify breakage points for Mosaic’s current SQL query generation and design solutions for translating from an internal "canonical" query language to different DB dialects.

A good starting point would be to leverage [multi database support in DuckDB](https://duckdb.org/2024/01/26/multi-database-support-in-duckdb.html).

weifanwu:
working on this issue!

derekperkins:
We're interested in BigQuery + StarRocks adapters. Once there's an adapter pattern, we'd be happy to help contribute SQL interface translations.

MarcSkovMadsen:
I'm interested in MS SQL, Azure SQL and similar databases.

domoritz:
https://github.com/tobymao/sqlglot could be an interesting experiment to build a rest server that proxies queries to another database. I'm sure it's still tricky because of data loading but worth a try.

weifanwu:
Wrote documentation for how to connect (PostgreSQL, MySQL, SQLite) using DuckDB extensions:

https://github.com/uwdata/mosaic/pull/430

domoritz:
You're amazing! I have to dig deeper into your pull request but it looks very exciting.

jheer:
Wow, still need to review in detail but this looks really promising!

domoritz:
Thanks for sharing. DuckDB wasm is already a dependency so wasm isn't an issue. The trickiest thing is probably having the right connector and data loading which would need to be written per database. There probably are major differences. The playground for polyglot says it works but I couldn't find any docs in Postgres for this. 

<img width="1349" height="356" alt="Image" src="https://github.com/user-attachments/assets/37e7e95c-9c63-4912-bf3c-59efce1f0410" />

dzole0311:
I did a small experiment around Mosaic with a non-DuckDB backend: https://github.com/dzole0311/mosaic-datafusion-zarr-deckgl (I have a deployment set up here: https://mosaic-datafusion-zarr-deckgl.netlify.app/)

https://github.com/user-attachments/assets/f2f70381-a2e6-4d57-8b3d-8c5b1863e66f

[...]

PR #430, opened 2024-06-10
Create documentation for supporting multi database

weifanwu:
Wrote the documentation to support multiple database connections (PostgreSQL, MySQL, SQLite) with DuckDB extensions. 

TODO:
- [ ] (Implement feature that supports other databases don't have DuckDB extensions yet such as  BigQuery, StarRocks or SnowFlakes)

Issue: #399


docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
...
+
+const password = "YOUR_PASSWORD";
+
+const postgres_connection_string = `dbname=postgres user=postgres password=${password} host=127.0.0.1`;
+
+await vg.coordinator().exec([
+  "INSTALL postgres;",   // Install the postgres extension 
+  "LOAD postgres;",   // Load postgres extension manually, 
+  `ATTACH '${postgres_connection_string}' AS postgres_db (TYPE POSTGRES);`,   // Attach PostgreSQL databases
+  "SET search_path TO postgres_db;",   // Set the schema search path for the current session
```
jheer: After the `ATTACH`, might the [`USE` statement](https://duckdb.org/docs/sql/statements/use.html) be a preferred way to target the postgres db instead of the more DuckDB-specific `SET`?

docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
...
+
+clear();
+
+const password = "YOUR_PASSWORD";
+
+const postgres_connection_string = `dbname=postgres user=postgres password=${password} host=127.0.0.1`;
+
+await vg.coordinator().exec([
+  "INSTALL postgres;",   // Install the postgres extension 
+  "LOAD postgres;",   // Load postgres extension manually,
```
jheer: The `postgres` extension is one of the [official extensions and is listed as autoloadable](https://duckdb.org/docs/extensions/official_extensions). Are the `INSTALL` and `LOAD` statements actually needed here?

docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
+# Connect DuckDB to other Databases: 
+
+Mosaic utilizes DuckDB, which [can supports multiple databases](https://duckdb.org/2024/01/26/multi-database-support-in-duckdb.html) such as MySQL, PostgreSQL, and SQLite, by using [DuckDB extensions](https://duckdb.org/docs/extensions/overview.html). This capability allows data to be seamlessly read into DuckDB and transferred between systems, enhancing flexibility and interoperability in data management.
+
+## Examples:
+### Connecting to `PostgreSQL`:
+
+``` js
+import { vg, clear } from "http://localhost:5173/dev/setup.js";
```
jheer: This import is only valid for the Mosaic dev setup. We should use a normal import (`from '@uwdata/vgplot';`) instead and we should not include `clear` here.

docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
...
+  )
+);
+
+```
+
+
+### Connecting to `MySQL`:
+
+``` js
+import { vg, clear } from "http://localhost:5173/dev/setup.js";
```
jheer: As above: This import is only valid for the Mosaic dev setup. We should use a normal import (`from '@uwdata/vgplot';`) instead and we should not include `clear` here.

docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
...
+clear();
+
+const password = "YOUR_PASSWORD";
+
+const databaseName = "YOUR_DATABASE_NAME";
+
+const mysql_connection_string = `host=localhost password=${password} user=root port=0 database=${databaseName}`;
+
+await vg.coordinator().exec([
+  "INSTALL mysql;",   // Install MySQL extension
```
jheer: I don't think the semi-colons are necessary here when using an array argument to `exec`?

docs/api/duckdb/multi-database-support.md
```diff
@@ -0,0 +1,157 @@
...
+await vg.coordinator().exec([
+  "INSTALL sqlite;",   // Install SQLite extension
+  "LOAD sqlite;",   // Load SQLite extension
+  `ATTACH 'sakila.db' (TYPE SQLITE);`,   // Attach SQLite database
+  "USE sakila;",   // Switch to the database called sakila
+  vg.loadParquet("ca55", "http://localhost:5173/data/ca55-south.parquet"), // Load the dataset into the SQLite database
+]);
+
+// Create interactive visualizations for your dataset
+const $interp = vg.Param.value("random-walk");
```
jheer: I don't think we need to repeat the specification for all three databases. I think including it once (the first time) should be plenty. Then the rest of the example could just be a comment like `// Mosaic specification goes here` or similar.

jheer (changes requested):
Thanks! Please see the various comments. Once addressed, this should be ready to merge.

In addition, I think this documentation should reside under `/api/core/` instead of `/api/duckdb/` (as the latter is specific to Mosaic's Node.js DuckDB server package).

jheer (approved):

jheer:
Thanks @weifanwu!

PR #438, opened 2024-06-20
Asynchronous Querying Implemented

kzhang31415:
Still not sure whether the queries come back in the right order, but queries are now sent without waiting for the previous queries to return. Please let me know if any changes need to be made to the code.


kzhang31415:
Pull request has been cleaned up by the most recent commit, changes in all other files besides QueryManager.js have been reverted

jheer:
Thanks! A few quick notes:

- 4/5 files in this PR have unnecessary changes. Only QueryManager needs to be edited.
- Main TODOs for issuing multiple queries in parallel is (1) to ensure correctness of the results and (2) measure to what degree, if any, doing so improves performance. So we would want to get tests and possible some performance measures in place before we merge a PR that eliminates query serialization.

package-lock.json
domoritz, line 1: Please undo this change.

package-lock.json
```diff
@@ -14985,4 +14985,4 @@
       }
     }
   }
-}
+}
```
domoritz: Please undo this change.

package.json
```diff
@@ -45,4 +45,4 @@
   "workspaces": [
     "packages/*"
   ]
-}
+}
```
domoritz: Please undo this change.

packages/core/src/QueryManager.js
```diff
@@ -18,10 +18,24 @@ export class QueryManager {
...
+    const requests = [];
+    const results = [];
+    while (!this.queue.isEmpty()) {
+      const { request, result } = this.queue.next();
+      requests.push(request);
+      results.push(result);
+    }
+    this.pending = this.submitBatch(requests, results);
+    this.pending.finally(() => { this.pending = null; this.next(); });
+    /* */
```
domoritz:
```suggestion
```

packages/core/src/util/convert-arrow.js
```diff
@@ -142,4 +142,4 @@ function decimalToNumber(v, scale) {
   }
 
   return x * scale;
-}
+}
```
domoritz: Please undo this change.

packages/core/src/util/selection-types.ts
```diff
@@ -134,4 +134,4 @@ export interface SelectionClause {
    * by creating pre-aggregated data cubes when applicable.
    */
   meta?: ClauseMetadata;
-}
+}
```
domoritz: Please undo this change.

packages/core/src/QueryManager.js
```diff
@@ -18,10 +18,24 @@ export class QueryManager {
   }
 
   next() {
+    /*
     if (this.pending || this.queue.isEmpty()) return;
     const { request, result } = this.queue.next();
     this.pending = this.submit(request, result);
     this.pending.finally(() => { this.pending = null; this.next(); });
+    */
+    /* */
```
domoritz: Please don't comment out code in a pull request but remove it so we can see the diff.

domoritz (changes requested):
Thanks for the pull request. Please make sure only necessary changes are included in the future. 

Please add an explanation for the changes you made in the pull request description. 

Also, we will need to evaluate the actual performance improvements and also make sure that the order or queries is preserved. I think a combination of unit tests and a semi-format proof would be best to show that.

kzhang31415:
[Query_Parallelization_Writeup.pdf](https://github.com/user-attachments/files/16182401/Query_Parallelization_Writeup.pdf)
Here is a writeup including a proof of performance increase due to changes in this pull request, some benchmarks on the performance gains, and pseudocode of the new and old code. I've also just added unit tests in the coordinator-test.js file. I'm still unable to remove package-lock.json from the changed files for some reason, despite there being no changes in that file.

packages/core/test/coordinator-test.js
```diff
@@ -1,8 +1,11 @@
 import assert from 'node:assert';
 import { Coordinator, coordinator } from '../src/index.js';
+import { QueryResult } from '../src/util/query-result.js';
+import { resolve } from 'node:path';
```
domoritz: bad imports

packages/core/src/QueryManager.js
```diff
@@ -49,7 +53,7 @@ export class QueryManager {
       if (cache) {
         const cached = this.clientCache.get(sql);
         if (cached) {
-          this._logger.debug('Cache');
+          this._logger.debug(`Cache`);
```
domoritz: please undo

packages/core/src/QueryManager.js
```diff
@@ -62,13 +66,46 @@ export class QueryManager {
...
+          }
+        }
+        const t0 = performance.now();
+        if (this._logQueries) {
+          this._logger.debug('Query', { type, sql, ...options });
+        }
+        this.db.query({ type, sql, ...options }).then(data => {
+          if (cache) this.clientCache.set(sql, data);
+          this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
+          results[i].fulfill(data);
```
domoritz: I don't think this correctly sorts the results yet, no? The request is fulfilled in the order of the `then` in line 96.

packages/core/test/coordinator-test.js
```diff
@@ -1,8 +1,11 @@
...
+import { send } from 'vite';
 
 describe('coordinator', () => {
-  it('has accessible singleton', () => {
+  it('has accessible singleton', async () => {
```
domoritz:
```suggestion
  it('has accessible singleton', () => {
```

packages/core/test/coordinator-test.js
```diff
@@ -11,4 +14,26 @@ describe('coordinator', () => {
...
+  it('query results returned in correct order', async () => {
+
+    const connector = {
+      uid : 0,
+
+      async query(query) {
+        this.uid++;
+        let id = this.uid;
+        //Earlier queries are resolved later so that they are purposely returned out of order
+        let result = new Promise((resolve, reject) => setTimeout(() => resolve(id), 200*(10-id)));
```
domoritz: Let's not use timeout in tests because it slows down the tests. Instead figure out a way to resolve the promises out of order.

domoritz (changes requested):

kzhang31415:
Updated writeup with proof of correct
[Query_Parallelization_Writeup.pdf](https://github.com/user-attachments/files/16339287/Query_Parallelization_Writeup.pdf)
 results ordering

domoritz:
This still doesn't seem to be correct. 

with this test

```
async function wait() {
  return new Promise(setTimeout);
}

describe("coordinator", () => {
  it("query results returned in correct order", async () => {
    const promises = [];

    const connector = {
      async query() {
        const promise = new QueryResult();
        promises.push(promise);
        return promise;
      },
    };

    const coord = new Coordinator(connector);

    const r0 = coord.query();
    const r1 = coord.query();
    const r2 = coord.query();
    const r3 = coord.query();

    // queries have not been sent yet
    assert.equal(promises.length, 0);

    await wait();

    // all queries should have been sent to the connector
    assert.equal(promises.length, 4);

    // resolve promises in reverse order
    promises.at(-1).fulfill();
    assert.equal(r0.pending, true);

    promises.at(-2).fulfill();
    assert.equal(r1.pending, true);

    promises.at(-3).fulfill();
    assert.equal(r2.pending, true);

    // promises are only fulfilled after the last request to the coordinator resolves
    promises.at(-4).fulfill();
    assert.equal(r0.pending, false);
    assert.equal(r1.pending, false);
    assert.equal(r2.pending, false);
    assert.equal(r3.pending, false);
  });
});

```

and changing query result to 

```
/**
 * A query result Promise that can allows external callers
 * to resolve or reject the Promise.
 */
export class QueryResult extends Promise {
  /**
   * Create a new query result Promise.
   */
  constructor() {
    let resolve;
    let reject;
    super((r, e) => {
      resolve = r;
      reject = e;
    });
    this._resolve = resolve;
    this._reject = reject;
    this._pending = true;
  }

  /**
   * Resolve the result Promise with the provided value.
   * @param {*} value The result value.
   * @returns {this}
   */
  fulfill(value) {
    this._pending = false;
    this._resolve(value);
    return this;
  }

  /**
   * Rejects the result Promise with the provided error.
   * @param {*} error The error value.
   * @returns {this}
   */
  reject(error) {
    this._pending = false;
    this._reject(error);
    return this;
  }

  /**
   * Whether this promise is still pending.
   * @returns {boolean}
   */
  get pending() {
    return this._pending;
  }
}

// necessary to make Promise subclass act like a Promise
QueryResult.prototype.constructor = Promise;
```

it looks like the responses are resolved before all the requests return from the connector

domoritz:
I think this needs a bit more work to be correct but also address some missing features.

* support canceling queries
* support limiting the number of queries sent
* support caching and not sending the same query multiple times (while also supporting canceling)

I'm taking a stab at another approach in a separate branch and will share what I'm learning.

domoritz:
Superseded by #475.

PR #440, opened 2024-06-21
Add selection relays to publish clauses to multiple selections.

### LINKED ISSUE #246: Let interactors update more than one parameter

jheer:
- Add `Selection.relay(...)` method to let a selection publish (relay) clauses to downstream selections.
- Add `relay` property to top-level of declarative specifications.
- Update `athletes` example to filter athlete auto-complete, use relays, and highlight hovered table row.

This relay mechanism allows a single interactor / clause source to populate more than one downstream selection, while still ensuring a "primary" selection is provided as the interactor's output (e.g., for two-way value binding). Close #246.

### Relay Syntax

The `relay` syntax in JavaScript takes one or more target selections (or arrays thereof) as arguments:

```js
sourceSelection.relay(targetSelection1, targetSelection2, ...)
```

The `relay` syntax in a declarative spec uses the source selection name as a key (mirroring `params` definitions) and one or more target selection references (with `$` prefix, mirroring other param references in a spec):

```yaml
relay:
  sourceSelection: [$targetSelection1, $targetSelection2]
```

### Other Options to Consider

The syntax added in this PR separates selection definition from relay network specification. An alternative approach could be to include relay specification as part of selection definition within declarative specs. For example, if defined relative to the source selection:

```yaml
params:
  sourceSelection: { select: intersect, relay: [$targetSelection1, $targetSelection2] }
```

This has the advantage of keeping selection specification more consolidated, but also has the perhaps unintuitive requirement that logically prior "source" selection definitions must reference downstream selections as part of their definition.

A flipped approach would be for target selections to instead specify source selections, like so:

```yaml
params:
  targetSelection: { select: intersect, includes: $sourceSelection }
```

This feels cleaner to me, but also cuts against the grain of the underlying JS implementation, in which targets are registered with sources. If we ultimately prefer this target->source style of specification, we could update the external JS  API to use a consistent style.

Comments / opinions welcomed!

### DIFF
--- a/docs/public/specs/yaml/athletes.yaml
+++ b/docs/public/specs/yaml/athletes.yaml
@@ -1,25 +1,32 @@
 meta:
   title: Olympic Athletes
   description: >
-    An interactive dashboard of athlete statistics.
-    The input menus and searchbox filter the display and are automatically populated by backing data columns.
+    An interactive dashboard of athlete statistics. The menus and searchbox
+    filter the display and are automatically populated by backing data columns.
 data:
   athletes: { file: data/athletes.parquet }
+params:
+  category: { select: intersect }
+  query: { select: intersect }
+  hover: { select: intersect, empty: true } # select nothing when empty
+relay:
+  category: $query # relay clauses from category selection to query selection
 hconcat:
 - vconcat:
   - hconcat:
     - input: menu
       label: Sport
-      as: $query
+      as: $category
       from: athletes
       column: sport
     - input: menu
       label: Sex
-      as: $query
+      as: $category
       from: athletes
       column: sex
     - input: search
       label: Name
+      filterBy: $category
       as: $query
       from: athletes
       column: name
@@ -41,6 +48,14 @@ hconcat:
     - select: intervalXY
       as: $query
       brush: { fillOpacity: 0, stroke: black }
+    - mark: dot
+      data: { from: athletes, filterBy: $hover }
+      x: weight
+      y: height
+      fill: sex
+      stroke: currentColor
+      strokeWidth: 1
+      r: 3
     xyDomain: Fixed
     colorDomain: Fixed
     margins: { left: 35, top: 20, right: 1 }
@@ -52,5 +67,6 @@ hconcat:
     maxWidth: 570
     height: 250
     filterBy: $query
+    as: $hover
     columns: [name, nationality, sex, height, weight, sport]
     width: { name: 180, nationality: 100, sex: 50, height: 50, weight: 50, sport: 100 }
--- a/docs/examples/athletes.md
+++ b/docs/examples/athletes.md
@@ -5,7 +5,7 @@
 
 # Olympic Athletes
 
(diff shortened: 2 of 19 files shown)


frtennis1:
This is sweet, thank you!

docs/public/specs/yaml/athletes.yaml
```diff
@@ -1,25 +1,32 @@
...
+    An interactive dashboard of athlete statistics. The menus and searchbox
+    filter the display and are automatically populated by backing data columns.
 data:
   athletes: { file: data/athletes.parquet }
+params:
+  category: { select: intersect }
+  query: { select: intersect }
+  hover: { select: intersect, empty: true } # select nothing when empty
+relay:
+  category: $query # relay clauses from category selection to query selection
```
domoritz, line 13: Is the relay needed in this case? When I tried to only use `$query` and `$hover` I feel the example behaves the same. Am I missing something? If not, it would be nice to have an example where the relay is needed. It could be something like the example from https://github.com/uwdata/mosaic/issues/246.

domoritz (approved):
What I like the about the relay syntax is that it lets you specify the selections separate from how they influence each other. This feels more composable and I feel it would make it easier to write code that composes multiple client specifications (with selections) together. The biggest disadvantage to me is that you now have to read the relay section to fully understand what influences a selection. 

If we put the relay definition into the selections, I also prefer the second approach that puts the sources into the target. It lets me read a selection and know what selections affect it. Having to read all the other selections feels cumbersome. Having said that, to know what affects a selection, you already have you have to already read the client specifications so that makes the relay specification syntax not that much more complicated. 

Overall, I prefer the syntax in this pull request because of the composability aspect and because the definition in the selection doesn't make them that much easier to read.

jheer:
Closing in favor of #488.

PR #462, opened 2024-07-15
feat: streaming responses from rust server

domoritz:
WIP


domoritz:
Should switch to https://github.com/duckdb/duckdb-rs/pull/373 once it's released.

PR #475, opened 2024-08-02
feat: parallel queries

### LINKED ISSUE #388: Parallel Queries
Right now, Mosaic sends one query after another to the server. This means that it always has to wait for a result to return before a next query is sent which means the server is not used effectively. As a first step, try to send a batch of queries. By batching we still maintain the order or queries which avoids potential bugs with out-of-order responses. In the next step, try building a HTTP server that processes requests in parallel and returns results out of order without leading to bugs in the UI.

domoritz:
Updated version of #438

Rather than sending requests in series, send them in parallel but serialize the order before returning them from the query manager. 

Fixes #388


domoritz:
Some quick experiments on localhost. 

Before

node:
Screenshot 2024-08-05 at 10 42 07

python:

Screenshot 2024-08-05 at 10 46 02

After

node:
Screenshot 2024-08-05 at 10 40 35

python:

Screenshot 2024-08-05 at 10 45 31

domoritz:
I didn't do an in-depth analysis but in some quick comparisons it looks like very results are resolved more quickly. The network manager also shows queries running in parallel.

Here is an example with 2G network simulation in Firefox

Before: 
https://github.com/user-attachments/assets/bad87032-e756-49b0-8171-5ad3df0c0aac

After:
https://github.com/user-attachments/assets/9fc46638-c536-4e56-aaec-ef6ad0a374d9

packages/core/test/coordinator-test.js
jheer, line 1: This file changes the default quote from single to double. Can we make this consistent with other files?

.vscode/launch.json
jheer, line 1: Don't see why this file is part of the PR.

eslint.config.js
jheer, line 1: Don't see why this file is part of the PR.

packages/core/src/util/query-result.js
```diff
@@ -15,15 +22,41 @@ export class QueryResult extends Promise {
...
    */
   fulfill(value) {
-    this._resolve(value);
+    if (this._value !== undefined) {
+      if (value !== undefined) {
+        throw Error('Promise has prepared and provided value');
+      }
+      this._resolve(this._value);
+    } else if (value === undefined) {
+      throw Error('Promise has has neither prepared nor provided value');
```
jheer: Typo here.

packages/core/src/util/query-result.js
```diff
@@ -15,15 +22,41 @@ export class QueryResult extends Promise {
...
+    return this;
+  }
+
+  /**
+   * Prepare to resolve the result Promise with the provided value.
+   * @param {*} value The result value.
+   * @returns {this}
+   */
+  prepare(value) {
+    this._state = QueryState.prepared;
```
jheer: `QueryState.ready` might be a reasonable alternative to `prepared`?

packages/core/src/QueryManager.js
```diff
@@ -49,8 +84,9 @@ export class QueryManager {
       if (cache) {
         const cached = this.clientCache.get(sql);
         if (cached) {
+          const data = await cached;
           this._logger.debug('Cache');
-          result.fulfill(cached);
+          result.prepare(type === 'exec' ? null : data);
```
jheer: It's a separate issue, but this makes me think we should double check caching interactions with exec queries. I don't think we want to be going to the cache for any exec queries...

jheer (changes requested):
LGTM, once comments are addressed.

.vscode/launch.json
domoritz: Moved to [#489](https://github.com/uwdata/mosaic/pull/489)

eslint.config.js
domoritz: Pulled out into [#490](https://github.com/uwdata/mosaic/pull/490)

packages/core/src/util/query-result.js
domoritz: Sounds good. Should the method still be called `prepare` or become `ready`?

packages/core/src/QueryManager.js
domoritz: It looks like [cache is set to false by default](https://github.com/uwdata/mosaic/blob/76e4874f05eacb27452e85f49c38d5adddd3dfbf/packages/core/src/QueryManager.js#L40) and `exec` [makes a query without cache](https://github.com/uwdata/mosaic/blob/76e4874f05eacb27452e85f49c38d5adddd3dfbf/packages/core/src/Coordinator.js#L118) so it should always be off.

jheer: Good, sounds like younger me did think it through :)

packages/core/src/util/query-result.js
jheer: I'd rename it to `ready`

domoritz: done

domoritz:
I tested the (at least casually) the flight 200k and 10m examples on

* node sever socket and rest
* python server socket and rest
* wasm

There is a problem when I brush on one chart and move the cursor over another chart because mosaic then tries to run the query to build the index and query the index at the same time (and the latter fails since the index is not there yet).

I guess this problem (index not ready yet) can also happen if the index creation is slow and you start interacting without the index being there. I think the solution here is to always prioritize exec queries and don't send any queries until the exec is done. I'll implement that unless you have another suggestion. We may still want to prevent activation of chart while the mouse is down. At some point it might be nice to have an actual dependency model for queries but that's a lot more invasive.

jheer:
> I guess this problem (index not ready yet) can also happen if the index creation is slow and you start interacting without the index being there. I think the solution here is to always prioritize exec queries and don't send any queries until the exec is done. I'll implement that unless you have another suggestion. We may still want to prevent activation of chart while the mouse is down. At some point it might be nice to have an actual dependency model for queries but that's a lot more invasive.

I think blocking on `exec` queries probably makes sense. There are other specs with derived data sets on load that also assume serialization -- though in some cases multiple exec queries may be issued as a single DB request, I don't think this is always the case.

I agree that a full dependency model is a bigger lift. That said, checking (and if necessary, blocking) on data cube table construction within the data cube indexer might not be so bad as a stop-gap.

Either way, we should make sure we have accompanying unit tests for these cases.

domoritz:
Added logic to prevent multiple parallel exec requests and also added a unit test.

packages/core/src/QueryManager.js
```diff
@@ -1,29 +1,64 @@
...
+    if (request.type === 'exec') this.pendingExec = true;
+
+    this.submit(request, result).finally(() => {
+      // return from the queue all requests that are ready
+      while (this.pendingResults.length && this.pendingResults[0].state !== QueryState.pending) {
+        const result = this.pendingResults.shift();
+        if (result.state === QueryState.ready) {
+          result.fulfill();
+        } else if (result.state === QueryState.done) {
+          console.warn('Found resolved query in pending results.');
```
jheer: Shouldn't this call route through the logger?

packages/core/src/QueryManager.js
```diff
@@ -49,8 +89,9 @@ export class QueryManager {
       if (cache) {
         const cached = this.clientCache.get(sql);
         if (cached) {
+          const data = await cached;
           this._logger.debug('Cache');
-          result.fulfill(cached);
+          result.ready(type === 'exec' ? null : data);
```
jheer: This line looks harmless, but raises a question: do we ever see exec calls hit the cache? I don't think we should... they should be issued with `cache: false | undefined`. And if we didn't shouldn't the cache result resolve to `undefined` anyway? So I don't think we need this conditional.

packages/core/src/QueryManager.js
```diff
@@ -60,10 +101,16 @@ export class QueryManager {
...
+      const promise = this.db.query({ type, sql, ...options });
+      if (cache) this.clientCache.set(sql, promise);
+
+      const data = await promise;
+
       if (cache) this.clientCache.set(sql, data);
+
       this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
-      result.fulfill(data);
+      result.ready(type === 'exec' ? null : data);
```
jheer, line 113: Same comment as above.

packages/duckdb-server-rust/src/app.rs
```diff
@@ -42,7 +42,7 @@ async fn handle_post(
 
 pub fn app() -> Result<Router> {
     // Database and state setup
-    let db = ConnectionPool::new(":memory:", 16)?;
+    let db = ConnectionPool::new(":memory:", 1)?; // TODO: we can only use one connection since temp tables are scoped per connection
```
jheer, line 45: Once #519 lands data cube indexes will no longer be temp tables. But other Mosaic exec calls may make temp tables, so I don't think we can expand the connection pool...

jheer (approved):
Overall looks good to me, pending my review comments. _But please do not merge until after #519 lands and I've cut a v0.11.0 release_. I plan to cut a release without parallel queries and then merge this PR for the subsequent release, in part to permit more careful testing.

In addition, we need to check all of the database connectors. I think `rest` and `wasm` should be fine. I didn't re-check the Jupyter widget but I think it will be ok for parallel use, too. However, the `socket` connector includes its own blocking, forcing sequential query execution. We need to revisit that.

packages/core/src/QueryManager.js
domoritz: Done

packages/core/src/QueryManager.js
domoritz: Right, we don't cache exec requests but actually could benefit from it. Below, we add the promise for pending requests to the cache so we don't submit the same request twice.

This makes me think that we maybe want to have a separate map of pending requests that we put requests in no matter whether their results should be cached or not.

jheer: I don’t think it is safe to cache exec queries, as they can modify DB state. For example if I add a table, drop it, then add it again (which could maybe happen for a data cube index table), I don’t want the last exec call ignored due to caching.

domoritz: Good example. A common request of creating an index would be nice to cache, though. I guess for now we can not cache and hopefully improve the parallelization when we have full query dependency tracking.

packages/duckdb-server-rust/src/app.rs
domoritz: Yes, will leave this at 1 for now. 

What are your thoughts on not using temp tables at all anymore? We could change all the examples to write tables into a separate namespace so they are still easy to clean up.

packages/core/src/QueryManager.js
domoritz: We need this here so that we don't err in QueryResult.fulfill.

domoritz:
Updated the code to address your comments. 

> However, the socket connector includes its own blocking, forcing sequential query execution. We need to revisit that.

I tested with the ws connector and it works fine. We are unnecessarily trying to parallelize queries that are then executed in series by the connector, yes. I could add a flag to not parallelize in the query manager but that introduces an extra code path so I'd be in favor of merging as is.

jheer:
> I tested with the ws connector and it works fine. We are unnecessarily trying to parallelize queries that are then executed in series by the connector, yes. I could add a flag to not parallelize in the query manager but that introduces an extra code path so I'd be in favor of merging as is.

I agree that we don't need another flag. My point was just to note that we need to update the socket connector to actually get benefits of parallel queries there.

domoritz:
Yes, that would be nice. I don't see an obvious way right now to do that, though. When we get a result on the socket, we just send binary so we can't easily change the order of responses. We would need to add some control signal that's sent before a result to send query results out of order. Should we do this in this pull request or a follow up (I can file an issue for it)?

domoritz:
I filed https://github.com/uwdata/mosaic/issues/547 as a follow up.

PR #479, opened 2024-08-12
fix: Don't consolidate aggregated and non-aggregated queries without a groupby

### LINKED ISSUE #478: Query consolidator can break when queries have non-grouped aggregations
Try to run the following queries:
- `SELECT x FROM table`
- `SELECT COUNT(*) FROM table`

The `QueryConsolidator` combines these to `SELECT "x" AS "col0", COUNT(*)::INTEGER AS "col1" FROM "table" AS "source"` which obviously fails.

```js
it('should not consolidate aggregated and non-aggregated queries', async () => {
        const qs = [];
        const c = consolidator(q => qs.push(q.request.query.toString()), voidCache(), () => {/*ignore*/});
        c.add({request: { type: 'arrow', query: Query.from({ source: 'table' }).select({ x: 'x' }) }}, Priority.Normal);
        c.add({request: { type: 'arrow', query: Query.from({ source: 'table' }).select({ x: count() }) }}, Priority.Normal);
        await new Promise(resolve => setImmediate(resolve));
        assert.equal(qs.length, 2);
    });
```

andyrooger:
In the `QueryConsolidator`, make sure `SELECT x FROM table` is not consolidated with `SELECT count() FROM table` since the aggregation in the latter query implies an implicit grouping which the former doesn't.

Implemented the fix by adding a GROUP BY * to the query's consolidation key where aggregates are present in the select.

Fixes #478

### DIFF
--- a/packages/core/src/QueryConsolidator.js
+++ b/packages/core/src/QueryConsolidator.js
@@ -108,6 +108,10 @@ function consolidationKey(query, cache) {
       // @ts-ignore
       q.$groupby(groupby.map(e => (e instanceof Ref && map[e.column]) || e));
     }
+    // @ts-ignore
+    else if(query.select().some(({ expr }) => expr.aggregate)) {
+      q.$groupby('*');
+    }
 
     // key is just the transformed query as SQL
     return `${q}`;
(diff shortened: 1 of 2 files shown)


andyrooger:
Thanks @jheer . Would you like me to switch the `GROUP BY *` in the PR to use `GROUP BY ALL` to match the code you mentioned? Seems to be a better fit like that.

jheer:
Thanks! I added the `ALL` (vs. `*`), added a comment and made the formatting the consistent.

PR #480, opened 2024-08-13
Replace Arrow JS with Flechette.

jheer:
This PR removes the `apache-arrow` dependency in favor of the smaller/faster/more flexible [`@uwdata/flechette`](https://github.com/uwdata/flechette). The change should improve performance a bit and simplifies Mosaic internals, as it obviates the need for internal custom Arrow data conversion steps. The DuckDB-WASM connector now queries for a raw Arrow IPC byte buffer and performs its own decoding.

The biggest concern here is the breaking change in Arrow data format. While the API for Flechette tables mirrors many aspects of the Arrow JS reference implementation, it is not a perfect one-to-one replacement. Some Mosaic users may need to adjust query result processing in response.

- **Breaking**: Replace `apache-arrow` with `@uwdata/flechette` to access Apache Arrow IPC data.


packages/core/src/connectors/wasm.js
```diff
@@ -1,4 +1,19 @@
...
+function getArrowIPC(con, query) {
+  return new Promise((resolve, reject) => {
+    con.useUnsafe(async (bindings, conn) => {
+      try {
+        const buffer = await bindings.runQuery(conn, query);
+        resolve(buffer);
+      } catch (error) {
+        reject(error);
+      }
+    });
```
domoritz, line 15: From https://github.com/duckdb/duckdb-wasm/issues/267#issuecomment-2252749509

packages/core/src/connectors/rest.js
```diff
@@ -20,7 +20,7 @@ export function restConnector(uri = 'http://localhost:3000/') {
       });
 
       return query.type === 'exec' ? req
-        : query.type === 'arrow' ? tableFromIPC(req)
+        : query.type === 'arrow' ? decodeIPC(await (await req).arrayBuffer())
```
domoritz, line 23: It would be nice to support streams here (at some point) so the client can process data while it's still being sent. See https://jakearchibald.com/2016/streams-ftw/.

domoritz (approved):
Looks good overall. Excited to see the performance differences.

PR #522, opened 2024-09-14
refactor: use simpler HMR setup for anywidget

domoritz:
building on https://anywidget.dev/blog/anywidget-02/#native-hot-module-replacement-hmr

CC @manzt


manzt:
Sorry, just catching up on this. Looks great! esbuild is definitely our preferred setup now with anywidget's built-in HMR.

PR #524, opened 2024-09-16
docs(widget): Clarify multi-dataframe/Arrow support for DuckDB

manzt:
The Mosaic widget not only supports Pandas data frames but any Arrow object (including other dataframes like polars).

re: https://github.com/manzt/quak/issues/54#issuecomment-2353416672

ref: https://duckdb.org/docs/api/python/data_ingestion#directly-accessing-dataframes-and-arrow-objects


domoritz (approved):
Awesome. Thank you for also adding the types.

Issue #312, opened 2024-03-04
Add TypeScript Support


maneetgoyal:
Any plans for adding types?

domoritz:
See #274

jheer:
Yes, I would like to add types! I think my preferred route is not to port to TypeScript, but rather provide type files included along side the JavaScript source files. (Similar to how [Observable Plot](https://github.com/observablehq/plot/tree/main/src) does this.)

domoritz:
Can't wait for https://tc39.es/proposal-type-annotations/. 🙏

maneetgoyal:
> Yes, I would like to add types! I think my preferred route is not to port to TypeScript, but rather provide type files included along side the JavaScript source files. (Similar to how [Observable Plot](https://github.com/observablehq/plot/tree/main/src) does this.)

Yeah, makes sense. Can avoid/defer a rewrite/migration to TS.

domoritz:
https://github.com/NiGhTTraX/ts-monorepo seems like a good solution to supporting typescript in a monorepo (without the need to rebuild everything to get the latest changes in downstream packages).

domoritz:
https://github.com/uwdata/mosaic/pull/355 gets us pretty close so I'll close this for now.

domoritz:
We still have a lot of places without types.

willium:
I think an underrated reason to port to TypeScript is that it could make contributing back to Mosaic easier for the uninitiated like myself. Currently, reading the code involves jumping between a lot of files, and reasoning about how variables are mutated, wrapped, and unwrapped at different points.

manzt:
I completely understand reasons to avoid adopting TypeScript, but I also agree with @willium that having types would greatly benefit contributions and upgrades. For example, with my client library (quak), types would give me much more confidence when upgrading to new releases.

I recently tried upgrading to v0.12 and ran into runtime errors I haven’t had time to debug yet. While I appreciate the detailed changelogs, it’s challenging to piece together how everything in mosaic fits. Types would provide a concrete guide for resolving these issues and maintaining compatibility.

>  I think my preferred route is not to port to TypeScript

Could I ask if this preference is based on avoiding repo complexity (e.g., a monorepo with many TypeScript packages and compilation steps) or concerns about type-checking the internals of the codebase itself? For instance, is it more about avoiding the overhead of TypeScript as a tool, or the constraints it might impose on how mosaic is implemented?

[...]

PR #544, opened 2024-10-02
ts-ignore on import in selection-types.ts

jordanhboxer:
This needs to be added otherwise I have to disable `noImplicitAny` in my tsconfig.json which can hide other errors.
This wouldn't be needed either if the file were a compiled declaration file (.d.ts) because of `skipLibCheck` in the tsconfig.json so this seems like the simplest solution

### DIFF
--- a/packages/core/src/util/selection-types.d.ts
+++ b/packages/core/src/util/selection-types.d.ts
@@ -1,5 +1,4 @@
 import { SQLExpression } from '@uwdata/mosaic-sql';
-// @ts-ignore Could not find a declaration file for module '../MosaicClient.js'.
 import { MosaicClient } from '../MosaicClient.js';
 
 /**
--- a/packages/core/src/util/selection-types.ts
+++ b/packages/core/src/util/selection-types.ts
@@ -1,4 +1,5 @@
 import { SQLExpression } from '@uwdata/mosaic-sql';
+// @ts-expect-error Could not find a declaration file for module '@uwdata/mosaic-core'.
 import { MosaicClient } from '../MosaicClient.js';
 
 /**
--- a/packages/core/src/util/selection-types.ts
+++ b/packages/core/src/util/selection-types.ts
@@ -1,5 +1,5 @@
 import { SQLExpression } from '@uwdata/mosaic-sql';
-// @ts-expect-error Could not find a declaration file for module '@uwdata/mosaic-core'.
+// @ts-ignore Could not find a declaration file for module '../MosaicClient.js'.
 import { MosaicClient } from '../MosaicClient.js';
 
 /**


kwonoh:
I get the below error from a typescript project with `"noImplicitAny": true` while this repo has `"noImplicitAny": false`. 

```
[TypeScript] Could not find a declaration file for module '../MosaicClient.js'. '<path>/@uwdata/mosaic-core/src/MosaicClient.js' implicitly has an 'any' type.
<path>/@uwdata/mosaic-core/src/util/selection-types.ts:2:30
    1 | import { SQLExpression } from '@uwdata/mosaic-sql';
  > 2 | import { MosaicClient } from '../MosaicClient.js';
      |                              ^^^^^^^^^^^^^^^^^^^^
    3 |
    4 | /**
    5 |  * Selection clause metadata to guide possible query optimizations.
 ```

Adding `// @ts-ignore` resolve the above issue. 

However, if `selection-types.ts` will only contain type definitions, I think it should be named as `selection-types.d.ts` so the library users can use [`skipLibCheck`](https://www.typescriptlang.org/tsconfig/#skipLibCheck) for addressing this kind of issue.

domoritz (changes requested):
Can you enable `skipLibCheck` in your config? Wouldn't that ignore type errors in the library? 

If we want to make a change here, I would suggest that we enable `noImplicitAny` to check everything and avoid potential regressions and instead of ts-ignore, should we add the missing declarations?

domoritz:
I see. I'm okay with moving to .d.ts files for everything but iirc there was some reason @jheer used ts files. 

Id love to have a setup that lets is gradually transition to ts. If you could make a proposal that would be awesome. 

Happy to merge this pull request if it's a stopgap but you would probably need a release, no?

domoritz:
If you can, go ahead and try.

kwonoh:
Changing `selection-types.ts` to `selection-types.d.ts` seems to work locally. I didn't see any errors from build/lint/test/docs:build. Waiting for the repo's workflow to kick in.

domoritz:
Support supportive of going ts but we want to make sure it's done in a way where we don't need to re-compile a sub-project to get the updates in another. Ideally, we figure out some way to do gradual transitions, that would be best and then send follow up PRs to change individual models. 

Related: https://github.com/uwdata/mosaic/issues/312

jordanhboxer:
@jheer That sounds like the ideal solution given that that will produce the correct output in the built package rather than using a declaration file in the source code. 

To give you some more context, our build pipeline runs `tsc` on all `.ts` files and since it hasn't been compiled to a declaration file we can't ignore it `skipLibCheck`. Setting `"noImplicitAny": false` is not ideal for us either as that will hide potential issues.

domoritz:
Like this https://github.com/uwdata/mosaic/pull/545?

jordanhboxer:
I believe so!

domoritz:
Should be fixed by https://github.com/uwdata/mosaic/pull/545.

PR #549, opened 2024-10-03
fix: correct connector types

domoritz:
Not needed when the request is a bundle request but we do need other parameters then.

CC https://github.com/alexkreidler


jheer (approved):
LGTM

Issue #553, opened 2024-10-08
Return dataServer to allow for closing


mhkeller:
In a simple example running the node duckdb data server, I didn't see a way to close the server since the `dataServer` function doesn't return. I think you would need `return app.listen(port)` [here](https://github.com/uwdata/mosaic/blob/bce92a7e6a92c71859357542b7da656646d97113/packages/duckdb/src/data-server.js#L22).

Happy to start a PR with this if useful and I'm not missing something simple that would allow me to run `app.close()` after instantiating the server.

edit: that `return` example above was just illustrative. In practice, you would want the return to follow the logs, obviously

```js
const server = app.listen(port);
console.log(`Data server running on port ${port}`);
if (rest) console.log(`  http://localhost:${port}/`);
if (socket) console.log(`  ws://localhost:${port}/`);
return server;
```

jheer:
Thanks! A PR adding this change would be much appreciated.

domoritz:
Done in #554

PR #554, opened 2024-10-08
feat: return dataServer on creation (#553)

mhkeller:
As discussed in https://github.com/uwdata/mosaic/issues/553, this returns the server so that the user can subsequently do things with it such as close it. 

I didn't see an existing test for booting up the server where it would make sense to add this scenario but if I missed something, let me know.


domoritz:
Is this a fix or a feat?

mhkeller:
I can swap it to a feat if that's better.

PR #555, opened 2024-10-10
feat: remove mentions of data cubes

domoritz:
The materialized views of pre-aggregated data (indexes) are not data cubes (which would contain multiple levels of aggregates) so to be more precise, this change renames properties, methods, and updates comments.


packages/core/src/Indexer.js
```diff
@@ -7,42 +7,44 @@ import { fnv_hash } from './util/hash.js';
...
  */
 
 /**
- * Build and query optimized indices ("data cubes") for fast computation of
- * groupby aggregate queries over compatible client queries and selections.
- * A data cube contains pre-aggregated data for a Mosaic client, subdivided
- * by possible query values from an active selection clause. These cubes are
- * realized as as database tables that can be queried for rapid updates.
+ * Build and query optimized pre-aggregated materaialized views, which we refer
+ * to as *indexes* (not to be confused with indexes in databasees), for fast
```
jheer: Typo "databasees".

jheer (commented):
I think we should make this change, or something similar. My only hesitation is to think through the naming scheme a bit more. Looking across code and comments, we now have "preaggregation", "index"/"indexer", and "materialized views". I think we should settle on a primary name (ideally not *too* verbose) and use it consistently throughout.

I'm thinking we should move away from "index", as these aren't really indexes in the normal sense (b-tree, inverted index, etc). They are "projections" of a data cube (similar to the `PROJECTION` SQL construct in some DBs), though I don't like how this conflicts with the projection operator of relational algebra.

domoritz:
I agree that more consolidation would be good. When making these changes, it felt like I needed something short in many places. My favorite so far is "preaggregate". What do you think?

jheer:
@domoritz I streamlined the naming and expanded the updates to files that were missed in the first pass.

Unfortunately something seems to have gone wrong after I rebased this PR on main... prior commits are now showing up as part of the PR and CI tests are not starting for me. Locally, everything looks good and works. I just have to diagnose what went wrong with the branch...

domoritz:
Hmm, maybe just try a normal merge of main into this branch.

jheer:
Closing in favor of #566.

PR #566, opened 2024-10-24
Rename data cube indexing to pre-aggregation.

jheer:
- **Breaking**: Rename `DataCubeIndexer` class to `PreAggregator`.
- **Breaking**: Rename coordinator `dataCubeIndexer` property to `preaggregator`.
- **Breaking**: Rename coordinator `indexes` constructor option to `preagg`.
- **Breaking**: Rename pre-aggregator `index` method to `request`.
- **Breaking**: Rename pre-aggregator `dropIndexTables` method to `dropSchema`.
- Update documentation, internal methods, and comments to use the terminology of "pre-aggregation" and "pre-aggregated materialized views".


domoritz (approved):

domoritz (changes requested):
There are still a few mentions of indexes (e.g. `filterIndexable`)

domoritz:
Made some more changes in [`bd1d783` (#566)](https://github.com/uwdata/mosaic/pull/566/commits/bd1d78359cf7fc1fa713e77d39421ec5b39b907e). The name `optimizable` is maybe a bit too generic.

jheer:
> There are still a few mentions of indexes (e.g. `filterIndexable`)

Yeah, I had chosen not to change that at this point. While currently we only use that property for pre-aggregation, it does provide a slightly more general cue. But I'm open to picking a better name... maybe `filterStable` or `groupbyStable` or similar?

domoritz:
Renamed to filterStable

PR #571, opened 2024-10-30
Add region interactor, improve selection handling.

jheer:
- Add `region` interactor for 2D selection of individual point values.
- Add SQL `isIn` function.
- Add use of SQL IN operator for single-value point selections.
- Add `channels` mark property to spec types.
- Fix empty selection resolution to use false literal.
- Fix interactor facet handling and datum extraction.
- Refactor BrushStyles types to dedicated file.
- Update documentation and example specs.

### DIFF
--- a/packages/plot/src/interactors/util/intersect.js
+++ b/packages/plot/src/interactors/util/intersect.js
@@ -7,7 +7,7 @@ import { parsePath } from './parse-path.js';
  * @param {[[number, number], [number, number]]} extent Rectangular brush
  *  extent within which to select elements. The extent is represented as
  *  upper-left and bottom-right (x, y) coordinates.
- * @returns {SVGElement[]} A list of intersecting SVG elements.
+ * @returns {Element[]} A list of intersecting SVG elements.
  */
 export function intersect(svg, el, extent) {
   // svg origin in viewport coordinates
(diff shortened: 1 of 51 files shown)


packages/spec/src/spec/marks/Marks.ts
```diff
@@ -432,6 +432,12 @@ export interface MarkOptions {
     | (TipOptions & { pointer?: TipPointer })
     | ParamRef;
 
+  /**
+   * Additional named channels, for example to include in a tooltop.
```
domoritz:
```suggestion
   * Additional named channels, for example to include in a tooltip.
```

packages/plot/src/interactors/util/intersect.js
```diff
@@ -30,7 +30,7 @@ export function intersect(svg, el, extent) {
       list.push(child);
     }
   }
-  return list;
+  return /** @type {SVGElement[]} */(list);
```
domoritz: Might be easier/clearer to move this up to line 18?

domoritz (approved):
Looks great.

Screenshot 2024-10-30 at 13 43 36

packages/spec/src/spec/marks/Marks.ts
jheer: Hahaha. Good catch.

PR #573, opened 2024-10-30
ci: setup build/publish with uv

domoritz:


packages/widget/pyproject.toml
```diff
@@ -17,29 +13,33 @@ dependencies = [
...
-dev = [
-  "jupyterlab",
+[tool.uv]
+dev-dependencies = [
+  "jupyterlab>=4.2.5",
   "pandas",
   "pyyaml",
-  "watchfiles"
+  "ruff>=0.6.2"
 ]
```
manzt: There is now a standard for this kind of dependencies and the latest version of uv supports the standard: https://x.com/charliermarsh/status/1849897753612288396

```suggestion
[dependency-groups]
dev = [
  "jupyterlab>=4.2.5",
  "pandas",
  "pyyaml",
  "ruff>=0.6.2"
]
````

File: packages/duckdb-server/pyproject.toml [hunk 2]
```diff
@@ -23,36 +19,36 @@ duckdb-server = "pkg.__main__:serve"
...
-  "pytest",
-  "pytest-cov",
+[tool.uv]
+dev-dependencies = [
+    "watchdog[watchmehatch]",
+    "mypy>=1.11.1",
+    "pytest>=7.4.4",
+    "ruff>=0.6.1",
+    "types-ujson",
 ]
```
manzt: See commend below

```suggestion
[dependency-groups]
dev = [
    "watchdog[watchmehatch]",
    "mypy>=1.11.1",
    "pytest>=7.4.4",
    "ruff>=0.6.1",
    "types-ujson",
]
```

packages/duckdb-server/package.json
```diff
@@ -3,8 +3,10 @@
   "version": "0.11.0",
   "private": true,
   "scripts": {
-    "prepublishOnly": "rimraf dist && mkdir dist && hatch run test:cov && hatch fmt --check && hatch build",
-    "publish": "hatch publish --user __token__",
-    "release": "npm run prepublishOnly && npm run publish"
+    "prepublishOnly": "uv run pytest && uv run ruff check && uv run ruff format --check",
+    "publish": "uv build && uvx twine upload --skip-existing dist/*",
```
manzt, line 7: Don't mind this at all! We do something similar in anywidget. The funny thing is that uv is so flexible that you could probably publish a package to PyPI that just reads scripts from somewhere in the `pyproject.toml` and calls it (e.g., a python package called `project-scripts` that fulfills the following):

```
uvx project-scripts <script-name>
```

domoritz: Haha, that's a nice hack. I hope uv will support scripts/shortcuts/commands/dev-scripts (whatever they will call it) eventually to clean this up.

manzt: Agreed. IIRC, there has been some discussion around this, but I don't remember the thread or where I saw it.

domoritz:
The only thing left here is maybe using a workspace. Thoughts?

domoritz:
I switched to a workspace for the top-level so we can use the same dependencies and simplify re-using packages across the workspace. @manzt can you take a final look before merging?

manzt:
ha - I was just checking our the branch to poke around. This definitely seems like the right setup for a workspace (the shared lock file is ideal). I'm thinking you could also probably move some of the ruff config eventually here and run it from the root of the repo.

One thing I wasn't sure about with workspaces is the support for an "anonymous" root project, that is a `pyprojec.toml` without any `project` metadata. All the [uv workspace docs](https://docs.astral.sh/uv/concepts/workspaces/) only seem to discuss a hybrid of packages + root ... but I just checked out the branch and it's really nice!

.github/workflows/test.yml
```diff
@@ -41,29 +41,29 @@ jobs:
...
           cd packages/duckdb-server
-          hatch build
-          hatch fmt --check
-          hatch run test:cov
+
+          uv build
+          uv run ruff check
+          uv run ruff format --check
+          uv run mypy
+          uv run --with pytest-cov pytest --cov-report=term-missing --color=yes --cov=pkg
```
manzt: Depending on how you want to split out these, I believe you can run:

```
uv run ruff check
uv run ruff format --check
```

from the root of the repo. 

I  usally have a separate workflow for "Python Lint" e.g.,:


```sh
 python_lint:
    name: Lint Python
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - uses: astral-sh/setup-uv@v3
        with:
          python-version-file: ".python-version"
          cache: "pip"
          cache-dependency-path: "**/pyproject.toml"

      - name: Ruff
        run: |
          uv run ruff check
          uv run ruff format --check
          
  python_test:
    name: Test Python
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - uses: astral-sh/setup-uv@v3
        with:
          python-version-file: ".python-version"
          cache: "pip"
          cache-dependency-path: "**/pyproject.toml"

      - name: Widget
        run: |
          cd packages/widget
          uv build

      - name: Server
        run: |
          cd packages/duckdb-server
          uv run mypy
          uv run --with pytest-cov pytest --cov-report=term-missing --color=yes --cov=pkg
```

manzt (approved):
LGTM! The workspaces is really nice. Just a small comment about CI and moving around ruff stuff but I think up for debate/ for a separate PR

domoritz:
> One thing I wasn't sure about with workspaces is the support for an "anonymous" root project, that is a pyprojec.toml without any project metadata. All the [uv workspace docs](https://docs.astral.sh/uv/concepts/workspaces/) only seem to discuss a hybrid of packages + root ... but I just checked out the branch and it's really nice!

Yeah, I feel most of the time I don't want a root project and I'm glad uv is happily doing this. The docs could probably be updated to describe the anonymous workspace approach.

> I'm thinking you could also probably move some of the ruff config eventually here and run it from the root of the repo.

Done in [9314395](https://github.com/uwdata/mosaic/pull/573/commits/93143950b7e9555d634bfbeb252d5b8fa0ba102f)

manzt:
> Yeah, I feel most of the time I don't want a root project and I'm glad uv is happily doing this. 

Agreed, nice to know it works!

.github/workflows/test.yml
domoritz: I just moved out up. I didn't set up a separate workflow since the python one is plenty fast. 

setup-uv doesn't seem to like the python version file. 

Screenshot 2024-10-31 at 10 43 57

manzt: > setup-uv doesn't seem to like the python version file.

Yeah, uv is a python installer and by default i think picks a "managed" mode where it will determine and install interpreters rather than relying on the system. If there is a `.python-version` root of the project (like we've done in this PR) it will respect that when running various uv commands. You can specify the version with `UV_PYTHON=` in the env if you want to test multple python versions in matrix.

manzt: I think you could `setup-python` with a specific version and then `setup-uv` and run uv commands with `--system` to use the "system python" (i.e., whatever `setup-python` did), but i find the uv-only workflow most simple.

manzt: https://github.com/manzt/anywidget/blob/478e742eaaf85d8fab68b006210680232863a488/.github/workflows/ci.yml#L56-L69

For example, I should change this to:

```yaml
  TestPython:
    name: Python / Test
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version:
          - "3.8"
          - "3.9"
          - "3.10"
          - "3.11"
          - "3.12"
    steps:
      - uses: actions/checkout@v4
      - uses: pnpm/action-setup@v4
        with:
          run_install: true
      - uses: astral-sh/setup-uv@v3
        with:
          version: "0.4.x"
      - name: Run tests
        run: uv run --with pytest-cov pytest ./tests --color=yes --cov anywidget --cov-report xml
        env:
          UV_PYTHON: ${{ matrix.python-version }}
 ```

PR #585, opened 2024-11-08
feat: graph component

### LINKED ISSUE #391: Network/Graph Visualization Components
Add new Mosaic-powered interface components for visualizing network/graph data, for example as node-link diagrams or matrix views. The components should accept selections for filtering nodes and/or edges. One design consideration is how to best support visual encoding definitions for nodes (size, shape, color, …?) and/or edges (color, width, …?). A node-link component might generate Plot specifications to create an initial SVG rendering, then modify those SVG elements to perform layout (e.g., using [d3-force](https://github.com/d3/d3-force)). In addition, one might consider data transformations, for example aggregating nodes and edges akin to [PivotGraph](http://hint.fm/papers/pivotgraph.pdf).

BoolKang222:
Fixes #391 

I’ve added a client.js file to adapt the graph component using Cosmograph's api.

### DIFF
--- a/packages/graph_component/client.js
+++ b/packages/graph_component/client.js
@@ -1,0 +1,201 @@
+import { 
+    Cosmograph, 
+    CosmographHistogram, 
+    CosmographSearch, 
+    CosmographTimeline 
+  } from '@cosmograph/cosmograph';
+  import { throttle } from './util/throttle.js';
+  
+  export class CosmographClient {
+    constructor(targetElement, histogramElement, searchElement, timelineElement) {
+      // Create containers for all components
+      this._element = targetElement || document.createElement('div');
+      this._histogramElement = histogramElement || document.createElement('div');
+      this._searchElement = searchElement || document.createElement('div');
+      this._timelineElement = timelineElement || document.createElement('div');
+  
+      document.body.appendChild(this._element);
+      document.body.appendChild(this._histogramElement);
+      document.body.appendChild(this._searchElement);
+      document.body.appendChild(this._timelineElement);
+  
+      // Initialize the main Cosmograph instance
+      this._cosmograph = new Cosmograph(this._element, {
+        nodeColor: (node) => node.color || '#b3b3b3',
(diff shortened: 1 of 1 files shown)


domoritz (changes requested):
* make this a real npm package (add package.json etc), model after https://github.com/uwdata/mosaic/tree/main/packages/inputs for example
* add `@cosmograph/cosmograph` as a dependency to the package
* either add this client to vgplot and make an example spec with this client **or**
* make an example in your new package that puts the graph on a website using something like this

```
const el = document.querySelector("#app")

const graphWidget = new GraphComponent({element: el})
```

packages/graph_component/client.js
```diff
@@ -0,0 +1,201 @@
...
+  } from '@cosmograph/cosmograph';
+  import { throttle } from './util/throttle.js';
+  
+  export class CosmographClient {
+    constructor(targetElement, histogramElement, searchElement, timelineElement) {
+      // Create containers for all components
+      this._element = targetElement || document.createElement('div');
+      this._histogramElement = histogramElement || document.createElement('div');
+      this._searchElement = searchElement || document.createElement('div');
+      this._timelineElement = timelineElement || document.createElement('div');
```
domoritz, line 15: let's not do this yet until we have the graph fully working

packages/graph_component/client.js
```diff
@@ -0,0 +1,201 @@
...
+    CosmographHistogram, 
+    CosmographSearch, 
+    CosmographTimeline 
+  } from '@cosmograph/cosmograph';
+  import { throttle } from './util/throttle.js';
+  
+  export class CosmographClient {
+    constructor(targetElement, histogramElement, searchElement, timelineElement) {
+      // Create containers for all components
+      this._element = targetElement || document.createElement('div');
```
domoritz, line 12: I think we should fail when there is no element provided

domoritz:
Closing for now until we redo it.

PR #586, opened 2024-11-08
Rewrite SQL builders, make substantial project updates.

jheer:
- **Breaking**: Rewrite `mosaic-sql` package to use a complete AST formulation. Add additional SQL helper methods.
- **Breaking**: Drop `agg` SQL expression helper. We now perform automatic analysis to determine if an expression is an aggregate expression.
- **Breaking**: Remove automatic internal casts, such as casting count() aggregates to 32-bit integers. As we dropped Arrow-JS in favor of [Flechette](https://github.com/uwdata/flechette), Arrow parsing now gracefully handles 64-bit integers, decimals, and other type conversions.
- Move SQL transforms for binning and M4 to `mosaic-sql` package in a new `transforms` folder.
- Update pre-aggregation optimizations to support arbitrary aggregate expressions, not just individual aggregate function calls. We use the SQL facilities to rewrite expressions for more complex optimized aggregations.
- Update pre-aggregation optimizations to support aggregates with FILTER clauses.
- Refactor pre-aggregation optimizations, moving them to the new `mosaic-core/preagg` folder.
- Update plots, inputs, and spec parsing to use new SQL utilities. SQL expressions no longer provide column dependency lists, param listeners, or idiosyncratic annotations. Instead, needed metadata is now extracted using visitor methods that walk the SQL expression AST. All query generator methods now create full, analyzable ASTs using our updated SQL builders.
- Improved TypeScript types and tests in `mosaic-sql` and `mosaic-core`.
- Basic documentation updates (though more extensive additions needed later).


domoritz:
Sweet. I'll take a deeper look soon.

package-lock.json
```diff
@@ -4184,6 +4184,10 @@
       "resolved": "packages/sql",
       "link": true
     },
+    "node_modules/@uwdata/mosaic-sql-old": {
```
domoritz: Is this right?

packages/sql/src/visit/visitors.js
```diff
@@ -0,0 +1,103 @@
+import { AGGREGATE, COLUMN_REF, FRAGMENT, PARAM, VERBATIM, WINDOW } from '../constants.js';
+import { aggregateNames, AggregateNode } from '../ast/aggregate.js';
+import { ColumnRefNode } from '../ast/column-ref.js';
+import { SQLNode } from '../ast/node.js';
+import { walk } from './walk.js';
+
+const aggrRegExp = new RegExp(`^(${aggregateNames.join('|')})`);
+const funcRegExp = /(\\'|\\"|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|\$\w+|\w+\()/g;
```
domoritz: Could be more readable with comments.

```suggestion
const funcRegExp = /(
    \\'               | // Matches an escaped single quote (e.g., \')
    \\"               | // Matches an escaped double quote (e.g., \")
    "(?:\\"|[^"])*"   | // Matches a double-quoted string, allowing for escaped quotes (e.g., "text with \"escaped\" quotes")
    '(?:\\'|[^'])*'   | // Matches a single-quoted string, allowing for escaped quotes (e.g., 'text with \'escaped\' quotes')
    \$\w+             | // Matches a variable starting with $ followed by word characters (e.g., $variableName)
    \w+\(               // Matches a function name, capturing any word characters followed by an opening parenthesis (e.g., functionName()
)/g;
```

packages/sql/src/visit/walk.js
```diff
@@ -0,0 +1,30 @@
...
+  const result = visit(node);
+  if (result) return result;
+
+  const props = recurse[node.type];
+  const n = props?.length ?? 0;
+  for (let i = 0; i < n; ++i) {
+    const value = node[props[i]];
+    if (Array.isArray(value)) {
+      const m = value.length;
+      for (let j = 0; j < m; ++j) {
```
domoritz, line 21: Could be a for...of

packages/sql/src/transforms/m4.js
domoritz, line 1: Very nice to have this outside of vgplot now

domoritz (approved):

package-lock.json
jheer: No, will fix!

packages/sql/src/visit/walk.js
jheer: It could, but I wanted to keep the loops tight in cases of larger queries/expressions to walk. Using iterators adds non-trivial overhead. Maybe wouldn't really matter in this case, but I prefer to default to the more performant case for internals like this.

PR #592, opened 2024-11-15
Update top-level plot attributes.

jheer:
- Update top-level Plot attributes: remove `label`; add `clip`, `ariaLabel`, `ariaDescription`


domoritz:
Nice. This is good for accessibility.

PR #595, opened 2024-11-18
Support param-valued data tables.

jheer:
- Add Param-value support for `from` table names in vgplot marks and table views. This change permits dynamic swapping of backing data tables at runtime, including among differently sized samples.


domoritz (commented):
Very useful.

PR #607, opened 2024-11-23
Add selection activation for inputs.

jheer:
- Add selection activation for inputs. Inputs trigger selection activation upon keyboard focus or pointer enter (when a drag is not in progress).


domoritz (approved):
Nice improvement to make updates feel faster.

PR #609, opened 2024-11-26
chore: update to vite 6

domoritz:


jheer (approved):
LGTM, assuming you tested locally and everything's running as expected.

domoritz:
Yep, I tested it and worked as before.

PR #611, opened 2024-11-26
Fix dynamic column handling.

### LINKED ISSUE #610: Preaggregation fail for param referenced columns with brushed selection
Was playing around with the very nice new vg.column option but noticed that the preaggregation fails sometimes for plots with a brushed selection and regression line after updating the column. Also, can you specify somewhere what should happen with the selection when the columns are updated?  The selected area jumps around with column updates but would nice to have it reset most of the time.

- Select a xy interval
- Change column
- When the mouse enters the plot again the preaggregation fails (sometimes)

Using vgplot in observable framework loading in the flights-200k dataset:
````
```js
import * as vgplot from "npm:@uwdata/vgplot";

const db = await DuckDBClient.of({
  flights: FileAttachment("../data/flights-200k.parquet")
});

const coordinator = new vgplot.Coordinator();
const vg = vgplot.createAPIContext({coordinator});
coordinator.databaseConnector(vgplot.wasmConnector({duckdb: db._db}));

const $x = vg.Param.value("delay");
const $y = vg.Param.value("distance");

const $query = vg.Selection.intersect();

const regChart = vg.vconcat(
  vg.hconcat(
    vg.menu({
      label: "Y",
      options: ["delay", "distance", "time"],
      as: $y
    }),
    vg.menu({
      label: "X",
      options: ["delay", "distance", "time"],
      as: $x
    })
  ),
  vg.vspace(10),
  vg.plot(
    vg.heatmap(
      vg.from("flights"),
      {
        x: vg.column($x), 
        y: vg.column($y), 
        fill: "density"
      }
    ),
    vg.regressionY(
      vg.from("flights", {filterBy: $query}),
      {
        x: vg.column($x), 
        y: vg.column($y)
      }
    ),
    vg.intervalXY({as: $query, brush: {fillOpacity: 0, stroke: "currentColor"}}),
  )
);
```

${regChart}
````

Example of console errors:

image

### PR DESCRIPTION
**Pull request by jheer**
- Fix dynamic column handling. Ensure column extraction includes column params. Within vgplot marks, do not use column name as alias if it is a dynamic param.

Close #610.

### DIFF
--- a/packages/plot/src/marks/Mark.js
+++ b/packages/plot/src/marks/Mark.js
@@ -1,5 +1,5 @@
 import { isParam, MosaicClient, toDataColumns } from '@uwdata/mosaic-core';
-import { Query, SelectQuery, collectParams, column, isAggregateExpression, isColumnRef, isNode, isParamLike } from '@uwdata/mosaic-sql';
+import { Query, SelectQuery, collectParams, column, isAggregateExpression, isColumnParam, isColumnRef, isNode, isParamLike } from '@uwdata/mosaic-sql';
 import { isColor } from './util/is-color.js';
 import { isConstantOption } from './util/is-constant-option.js';
 import { isSymbol } from './util/is-symbol.js';
@@ -15,7 +15,7 @@ const isFieldObject = (channel, field) => {
 const fieldEntry = (channel, field) => ({
   channel,
   field,
-  as: isColumnRef(field) ? field.column : channel
+  as: isColumnRef(field) && !isColumnParam(field) ? field.column : channel
 });
 const valueEntry = (channel, value) => ({ channel, value });
 
--- a/packages/sql/src/ast/column-param.js
+++ b/packages/sql/src/ast/column-param.js
@@ -1,6 +1,15 @@
 import { COLUMN_PARAM } from '../constants.js';
 import { ColumnRefNode } from './column-ref.js';
 
+/**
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/sql/src/ast/column-param.js [hunk 1]
```diff
@@ -1,6 +1,15 @@
 import { COLUMN_PARAM } from '../constants.js';
 import { ColumnRefNode } from './column-ref.js';
 
+/**
+ * Check if a value is a dynamic column reference node.
+ * @param {*} value The value to check.
+ * @returns {value is ColumnParamNode}
+ */
+export function isColumnParam(value) {
+  return value instanceof ColumnParamNode;
```
domoritz, line 10: Where is this used? Below, you are checking `node.type === COLUMN_PARAM`.

jheer: It’s used within the vgplot Mark class.

packages/plot/src/marks/Mark.js
```diff
@@ -15,7 +15,7 @@ const isFieldObject = (channel, field) => {
 const fieldEntry = (channel, field) => ({
   channel,
   field,
-  as: isColumnRef(field) ? field.column : channel
+  as: isColumnRef(field) && !isColumnParam(field) ? field.column : channel
```
domoritz, line 18: Ah, here.

PR #612, opened 2024-11-26
Add density 1D normalize option.

jheer:
- Add density 1D `normalize` option. If set to `true, scales density estimates by the total (weighted) point mass.

See also #606.

### DIFF
--- a/packages/spec/src/spec/marks/Density.ts
+++ b/packages/spec/src/spec/marks/Density.ts
@@ -46,10 +46,12 @@ export interface Density1DOptions {
   bins?: number | ParamRef;
 
   /**
-   * Flag indicating if density estimates should be normalized by dividing
-   * by the total point mass. Defaults to false.
+   * Normalization method for density estimates. If `false` or `'none'` (the
+   * default), the density estimates are smoothed weighted counts. If `true`
+   * or `'sum'`, density estimates are divided by the sum of the total point
+   * mass. If `'max'`, estimates are divided by the maximum smoothed value.
    */
-  normalize?: boolean | ParamRef;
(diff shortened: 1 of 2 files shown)


packages/spec/src/spec/marks/Density.ts
```diff
@@ -44,6 +44,12 @@ export interface Density1DOptions {
    * Defaults to 1024.
    */
   bins?: number | ParamRef;
+
+  /**
+   * Flag indicating if density estimates should be normalized by dividing
+   * by the total point mass. Defaults to false.
+   */
+  normalize?: boolean | ParamRef;
```
domoritz: I wonder whether we should make this an enum so that we can also support normalization by the max instead of the sum for example. 

```suggestion
  normalize?: 'sum' | 'max' | ParamRef;
```

domoritz (approved):

packages/spec/src/spec/marks/Density.ts
jheer: I’m ok doing this. I wonder if we should still support `true` as an alias for `”sum”`.

More generally `max` could lead to misinterpretation, but FWIW ggplot and other tools also support it. (For example to make violin plots where each violin has the same width.)

domoritz: true as an alias for sum sounds reasonable (since it's the better default) to me but maybe minor and adds two ways to do the same thing

jheer: I've updated the PR to include `max` normalization. The difference is importantly subtle: `sum` is applied using the unsmoothed grid (total point mass, as smoothing may push mass outside the visualized domain), while `max` is applied relative to the smoothed values.

domoritz (approved):
Looks good.

PR #622, opened 2024-12-01
Support multiple 1D density estimates.

jheer:
- Add groupby and stack support to `densityX` and `densityY` marks.
- Add `groupby` argument to `binLinear1D` SQL function.
- Add `density-groups` example.
- Expand param typing to accept `null` value.
- Fix `menu` input to support `null` values.


domoritz (approved):
Very nice and powerful feature.

PR #625, opened 2024-12-04
feat: export queryFieldInfo and jsType

kwonoh:
export two utils functions `queryFieldInfo` and `jsType`

### DIFF
--- a/packages/core/src/Coordinator.js
+++ b/packages/core/src/Coordinator.js
@@ -143,11 +143,10 @@ export class Coordinator {
   query(query, {
     type = 'arrow',
     cache = true,
-    persist = false,
     priority = Priority.Normal,
     ...options
   } = {}) {
-    return this.manager.request({ type, query, cache, persist, options }, priority);
+    return this.manager.request({ type, query, cache, options }, priority);
   }
 
   /**
--- a/packages/core/src/Coordinator.js
+++ b/packages/core/src/Coordinator.js
@@ -135,7 +135,7 @@ export class Coordinator {
    * @param {object} [options] An options object.
    * @param {'arrow' | 'json'} [options.type] The query result format type.
    * @param {boolean} [options.cache=true] If true, cache the query result.
-   * @param {boolean} [options.persist=false] If true, cache the query result.
+   * @param {boolean} [options.persist=false] If true, persist cached query result in the server.
    * @param {number} [options.priority] The query priority, defaults to
    *  `Priority.Normal`.
    * @returns {QueryResult} A query result promise.
--- a/packages/core/src/Coordinator.js
+++ b/packages/core/src/Coordinator.js
@@ -134,7 +134,7 @@ export class Coordinator {
    *  or a SQL string.
    * @param {object} [options] An options object.
    * @param {'arrow' | 'json'} [options.type] The query result format type.
-   * @param {boolean} [options.cache=true] If true, cache the query result.
+   * @param {boolean} [options.cache=true] If true, cache the query result in the client (QueryManager).
    * @param {boolean} [options.persist=false] If true, persist cached query result in the server.
    * @param {number} [options.priority] The query priority, defaults to
    *  `Priority.Normal`.
--- a/packages/core/src/Coordinator.js
+++ b/packages/core/src/Coordinator.js
@@ -27,7 +27,7 @@ export function coordinator(instance) {
 }
 
 /**
- * @typedef {import('@uwdata/mosaic-sql').Query | string} QueryType
+ * @typedef {import('@uwdata/mosaic-sql').Query | import('@uwdata/mosaic-sql').DescribeQuery | string} QueryType
  */
 
 /**
@@ -135,6 +135,7 @@ export class Coordinator {
    * @param {object} [options] An options object.
    * @param {'arrow' | 'json'} [options.type] The query result format type.
    * @param {boolean} [options.cache=true] If true, cache the query result.
+   * @param {boolean} [options.persist=false] If true, cache the query result.
    * @param {number} [options.priority] The query priority, defaults to
    *  `Priority.Normal`.
    * @returns {QueryResult} A query result promise.
@@ -142,10 +143,11 @@ export class Coordinator {
   query(query, {
     type = 'arrow',
     cache = true,
+    persist = false,
     priority = Priority.Normal,
     ...options
   } = {}) {
-    return this.manager.request({ type, query, cache, options }, priority);
+    return this.manager.request({ type, query, cache, persist, options }, priority);
   }
 
   /**
--- a/packages/core/src/util/field-info.js
+++ b/packages/core/src/util/field-info.js
@@ -10,10 +10,24 @@ export const Stats = { Count, Nulls, Max, Min, Distinct };
 
 /**
  * @typedef {Count | Distinct | Max | Min | Nulls} Stat
+ * 
+ * @typedef {{
+ *   table: string | import('@uwdata/mosaic-sql').TableRefNode,
+ *   column: (string | import('@uwdata/mosaic-sql').ColumnRefNode) & { aggregate?: boolean },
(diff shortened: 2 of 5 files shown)


jheer:
Thanks @kwonoh. This looks like a reasonable and straightforward modification. Can you share a bit more about why you'd like to have these methods exported? (It helps us to keep track of various use cases!)

jheer (changes requested):
Overall looks good. If we're going to expose the `queryFieldInfo` method, we should also add jsdoc comments and types before merging.

kwonoh:
@jheer I have added the jsdoc and type definitions. Please review the types closely. Some notes and questions:
1. I am not sure `MosaicClientField` is a good name for the return type of `MosaicClient.fields` and the type of `fields` param of `queryFieldInfo`
2. The params of `summarize` were all string, and I have changed that to be `string` or `(Table|Column)RefNode` based on my understanding of the code. Please confirm this is correct typing.
2. `Stats` in `field-info.js` seems not being used. What's the plan for this variable?

kwonoh:
@jheer I have added the jsdoc and type definitions. Now, it passes the tests, but please review the types closely. Some notes and questions:
1. I am not sure `FieldInfoRequest` is a good name for the return type of `MosaicClient.fields` and the type of `fields` param of `queryFieldInfo`
2. Could you explain where [`column.aggregate` is set in the `getFieldInfo` function?](https://github.com/uwdata/mosaic/blob/57580776bcf69a37b6c3c0724a970d4db7d8a7fd/packages/core/src/util/field-info.js#L51) I have added `& { aggregate?: boolean }` part in `FieldInfoRequest` for this usecase, but I couldn't find where it's coming from.
3. The params of `summarize` has been updated to use `FieldInfoRequest`.
4. `Stats` in `field-info.js` seems not being used. What's the plan for this variable?

packages/core/src/Coordinator.js
```diff
@@ -135,17 +135,19 @@ export class Coordinator {
    * @param {object} [options] An options object.
    * @param {'arrow' | 'json'} [options.type] The query result format type.
    * @param {boolean} [options.cache=true] If true, cache the query result.
+   * @param {boolean} [options.persist=false] If true, persist cached query result in the server.
```
kwonoh, line 138: `persist` option was missing in the type definition

packages/core/src/Coordinator.js
```diff
@@ -27,7 +27,7 @@ export function coordinator(instance) {
 }
 
 /**
- * @typedef {import('@uwdata/mosaic-sql').Query | string} QueryType
+ * @typedef {import('@uwdata/mosaic-sql').Query | import('@uwdata/mosaic-sql').DescribeQuery | string} QueryType
```
kwonoh, line 30: `DescribeQuery` was missing as a `QueryType` causing an error

packages/core/src/util/field-info.js
```diff
@@ -9,7 +9,25 @@ export const Distinct = 'distinct';
...
 
 /**
- * @type {Record<string, (column: string) => AggregateNode>}
+ * @typedef {Count | Distinct | Max | Min | Nulls} Stat
+ * 
+ * @typedef {{
+ *   table: string | import('@uwdata/mosaic-sql').TableRefNode,
+ *   column: (string | import('@uwdata/mosaic-sql').ColumnRefNode) & { aggregate?: boolean },
+ *   stats?: Stat[] | Set<Stat>
+ * }} FieldInfoRequest
```
kwonoh, line 18: Not sure `FieldInfoRequest` is a good name for a type that represents the return type of `MosaicClient.fields` and the type of `fields` param of `queryFieldInfo`

packages/core/src/util/field-info.js
```diff
@@ -9,7 +9,25 @@ export const Distinct = 'distinct';
...
+ * 
+ * @typedef {{
+ *   table: FieldInfoRequest["table"],
+ *   column: string,
```
kwonoh, line 22: While `FieldInfoRequest.column` is `string | ColumnRefNode`, both `getFieldInfo` and `getTableInfo` return only `string` type for `column` field.

packages/core/src/Coordinator.js
```diff
@@ -134,18 +134,20 @@ export class Coordinator {
    *  or a SQL string.
    * @param {object} [options] An options object.
    * @param {'arrow' | 'json'} [options.type] The query result format type.
-   * @param {boolean} [options.cache=true] If true, cache the query result.
+   * @param {boolean} [options.cache=true] If true, cache the query result in the client (QueryManager).
+   * @param {boolean} [options.persist=false] If true, persist cached query result in the server.
```
kwonoh, line 138: `persist` option was missing in the type definition causing an error

packages/core/src/Coordinator.js
```diff
@@ -134,18 +134,20 @@ export class Coordinator {
...
    * @param {number} [options.priority] The query priority, defaults to
    *  `Priority.Normal`.
    * @returns {QueryResult} A query result promise.
    */
   query(query, {
     type = 'arrow',
     cache = true,
+    persist = false,
```
kwonoh: The default value of `persist` differs depending on the server implementation
- [`duckdb-server-rust`: `true`](https://github.com/uwdata/mosaic/blob/57580776bcf69a37b6c3c0724a970d4db7d8a7fd/packages/duckdb-server-rust/src/query.rs#L18) cc @domoritz 
- [`duckdb-server` (Python): `false`](https://github.com/uwdata/mosaic/blob/57580776bcf69a37b6c3c0724a970d4db7d8a7fd/packages/duckdb-server/pkg/query.py#L24)
- [`duckdb` (Node.js): `false`](https://github.com/uwdata/mosaic/blob/57580776bcf69a37b6c3c0724a970d4db7d8a7fd/packages/duckdb/src/Cache.js#L52)

packages/core/src/Coordinator.js
jheer: Can you say more about where the error was coming from? I believe `persist` should be part of the aggregated `...options`, not broken out individually.

jheer (changes requested):

packages/core/src/Coordinator.js
kwonoh: > Can you say more about where the error was coming from?

The type error was from [`{ persist: true }` in`getFieldInfo`](https://github.com/uwdata/mosaic/blob/71cf64e074ab3c72401f599fd256317eec26cd3b/packages/core/src/util/field-info.js#L67). Currently, there is no `persist` field in the `options` type info of `Coordinator.query` here.

> I believe persist should be part of the aggregated ...options, not broken out individually.

~~`persist` type info is added as a member of `options` object (`options.persist` not `persist`), not as a separate field. Do you want to add a new interface type definition for `options` separately?~~

~~or did you mean there should be no default value assignment for `persist`?~~

packages/core/src/Coordinator.js
```diff
@@ -134,18 +134,20 @@ export class Coordinator {
...
     priority = Priority.Normal,
     ...options
   } = {}) {
-    return this.manager.request({ type, query, cache, options }, priority);
+    return this.manager.request({ type, query, cache, persist, options }, priority);
```
kwonoh, line 150: > I believe persist should be part of the aggregated ...options, not broken out individually.
@jheer did you mean persist should be passed as a member of `options` here?

packages/core/src/Coordinator.js
kwonoh: I now see what you mean. Updated the code.

packages/core/src/Coordinator.js
kwonoh: `persist = false,` has been removed as it needs to be part of `...options`.
But that means the default value of `persist` should be the same across the servers

jheer (approved):

jheer:
Thanks!

PR #629, opened 2024-12-14
Add headless render tests using JSDOM.

jheer:
- Add `pending` result property to `MosaicClient` to aid status tracking.
- Add headless render tests using JSDOM to `vgplot` package.


packages/vgplot/test/render.test.js
```diff
@@ -0,0 +1,58 @@
...
+  const htmlPath = resolve(cwd, `output/${name}.html`);
+  const [expected, { default: run }] = await Promise.all([
+    readFile(htmlPath, { encoding: 'utf8' }),
+    import(specPath)
+  ]);
+  const mc = new Coordinator(nodeConnector(), { logger: null });
+  const el = await run(createAPIContext({ coordinator: mc }));
+
+  await clientsReady(el);
+  expect(el.outerHTML).toEqual(expected);
```
domoritz: Might be easier to use `toMatchSnapshot` (https://vitest.dev/guide/snapshot) especially if we want to add more tests like this in the future.

domoritz (approved):

packages/vgplot/test/render.test.js
jheer: Ooh neat. This looks like the right method to use in the future: https://vitest.dev/guide/snapshot#file-snapshots

PR #631, opened 2024-12-16
feat: add $schema to JSON specs

domoritz:
This enables autocomplete and add explicit versions to the examples.


bin/prepare-examples.js
```diff
@@ -11,6 +11,8 @@ import { parse } from 'yaml';
 // - YAML, non-parsed JSON, and ESM code written to /docs/public/specs
 // - Example Markdown pages written to /docs/examples
 
+import pgk from '../packages/spec/package.json' with { type: "json" };
```
jheer: Why `pgk` instead of `pkg`?

jheer (commented):
I have mixed feelings about pegging all test specs to a specific version, as this creates an extra point of potential error, requiring regeneration of all examples with every version change. If we still want to include a schema, an alternative option would be to point to `latest`.

On the other hand, I can see why it might be nice to have a fixed schema in the published JSON on the website. Not sure that outweighs the other issues here, and we could potentially set schema separately solely for the online published examples.

domoritz:
Fixed the typo, tests, and added an explicit version variable. 

Post 1.0, we could for example switch to `const version = pkg.version.split('.').slice(0, 2).join('.');` so that the version only tracks the minor version but not the patch version. I could also change the tests to not test the `$schema` property if you prefer.

I don't feel strongly either way. Having auto complete in the specs in the repo is nice.

Screenshot 2024-12-16 at 12 39 26

jheer:
> Fixed the typo, tests, and added an explicit version variable.
> 
> Post 1.0, we could for example switch to `const version = pkg.version.split('.').slice(0, 2).join('.');` so that the version only tracks the minor version but not the patch version. I could also change the tests to not test the `$schema` property if you prefer.
> 
> I don't feel strongly either way. Having auto complete in the specs in the repo is nice.
> Screenshot 2024-12-16 at 12 39 26

FWIWI, the JSON specs are auto-generated from the YAML, so you probably shouldn't be editing those files anyway. (And for other files, you can of course add `$schema` if you like.) The YAML specs already have the schema and autocomplete applied if you are using VS Code, as the repo includes [Code settings](https://github.com/uwdata/mosaic/blob/main/.vscode/settings.json) to apply the schema to the YAML specs.

domoritz:
Oh neat. I added a setting so we get autocomplete for json as well: https://github.com/uwdata/mosaic/pull/633. Happy to defer to your preference for this pull request.

jheer:
Closing out in favor of #633 and to avoid maintenance issues around stale version numbers.

PR #649, opened 2025-01-09
feat: add address arg for duckdb-server-rust

kwonoh:
Add `--address` as part of CLI arguments. The default is `127.0.0.1` and if parsing the given value fails, it falls back to `Ipv4Addr::LOCALHOST`

```
> ./duckdb-server --help                       
DuckDB Server for Mosaic.

Usage: duckdb-server [OPTIONS] [DATABASE]

Arguments:
  [DATABASE]  Path of database file (e.g., "database.db". ":memory:" for in-memory database) [default: :memory:]

Options:
  -a, --address <ADDRESS>                            HTTP Address [default: 127.0.0.1]
  -p, --port <PORT>                                  HTTP Port [default: 3000]
      --connection-pool-size <CONNECTION_POOL_SIZE>  Max connection pool size [default: 10]
      --cache-size <CACHE_SIZE>                      Max number of cache entries [default: 1000]
  -h, --help                                         Print help
  -V, --version                                      Print version
```

### DIFF
--- a/packages/duckdb-server-rust/src/main.rs
+++ b/packages/duckdb-server-rust/src/main.rs
@@ -23,8 +23,8 @@ struct Args {
     database: String,
 
     /// HTTP Address
-    #[arg(short, long, default_value = "127.0.0.1")]
-    address: String,
+    #[arg(short, long, default_value_t = Ipv4Addr::LOCALHOST.into())]
+    address: IpAddr,
 
     /// HTTP Port
     #[arg(short, long, default_value_t = 3000)]
@@ -75,12 +75,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     // Listenfd setup
-    let addr = SocketAddr::new(
-        args.address
-            .parse::<IpAddr>()
-            .unwrap_or(Ipv4Addr::LOCALHOST.into()),
-        args.port,
-    );
+    let addr = SocketAddr::new(args.address, args.port);
     let mut listenfd = ListenFd::from_env();
     let listener = match listenfd.take_tcp_listener(0)? {
         // if we are given a tcp listener on listen fd 0, we use that one
--- a/packages/duckdb-server-rust/src/main.rs
+++ b/packages/duckdb-server-rust/src/main.rs
@@ -3,7 +3,7 @@ use axum_server::tls_rustls::RustlsConfig;
 use clap::Parser;
 use listenfd::ListenFd;
 use std::net::TcpListener;
-use std::{net::Ipv4Addr, net::SocketAddr, path::PathBuf};
+use std::{net::IpAddr, net::Ipv4Addr, net::SocketAddr, path::PathBuf};
 use tokio::net;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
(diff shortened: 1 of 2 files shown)


packages/duckdb-server-rust/src/main.rs
```diff
@@ -71,7 +75,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     // Listenfd setup
-    let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), args.port);
+    let addr = SocketAddr::new(
+        args.address
+            .parse::<IpAddr>()
+            .unwrap_or(Ipv4Addr::LOCALHOST.into()),
```
domoritz: Will this actually fall back since you define a default in line 26? I realize this might be a question for all the defaults here. It would be nice to define them only in one place.

domoritz (approved):

packages/duckdb-server-rust/src/main.rs
kwonoh: The above default value (`127.0.0.1`) will be used if nothing was passed. 
The value in `unwrap_or` (`Ipv4Addr::LOCALHOST.into()`) will be used if `parse` fails. For example, if an invalid IP address was passed like `--address 127.0`

kwonoh: > It would be nice to define them only in one place.

I agree. Do you think using global constant variables a good way to achieve this?

domoritz: I didn't know about the parse failure case you described. I wonder whether we just want to fail when the parsing of provided or default values fails. That might make things simpler.

kwonoh: I just made a change for that. Now, if `--address` is not provided, it will use default value (only defined in one place), if invalid IP address is provided, it will output the error, not using the default value.

```
> ./target/release/duckdb-server --address 0.0.0.0 --port 5002
2025-01-09T19:30:34.343971Z  INFO duckdb_server: Creating database in ':memory:'
2025-01-09T19:30:34.351430Z  WARN duckdb_server: No keys for HTTPS found.
2025-01-09T19:30:34.351448Z  INFO duckdb_server: DuckDB Server listening on http://0.0.0.0:5002 and ws://0.0.0.0:5002.
^C

> ./target/release/duckdb-server --port 5002 
2025-01-09T19:30:38.984626Z  INFO duckdb_server: Creating database in ':memory:'
2025-01-09T19:30:38.991219Z  WARN duckdb_server: No keys for HTTPS found.
2025-01-09T19:30:38.991237Z  INFO duckdb_server: DuckDB Server listening on http://127.0.0.1:5002 and ws://127.0.0.1:5002.
^C

> ./target/release/duckdb-server --address 999.999.999.999 --port 5002
error: invalid value '999.999.999.999' for '--address <ADDRESS>': invalid IP address syntax

For more information, try '--help'.
```

domoritz: Nice. Should we do the same for the app.rs file (e.g. https://github.com/uwdata/mosaic/blob/79e021829497d856d8b8ed8c564fd6eb380af42b/packages/duckdb-server-rust/src/app.rs#L52)?

kwonoh: Updated. Added the default values as global constants.

PR #650, opened 2025-01-09
feat: improve restConnector error handling

kwonoh:
The Rust server sends error messages to the client if the query fails (e.g., SQL syntax error, incorrect column name).
However, the current `restConnector` does not handle the error message from the server and tries to parse the response just based on the request type.

This PR adds a check whether the response was successful. If the response is not [`ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), then it throws an error with the response text that contains the error messages from the server.

With this change, `queryError` receives an Error instance with the error message from the server.

The error message format could be improved.

### DIFF
--- a/packages/core/src/connectors/rest.js
+++ b/packages/core/src/connectors/rest.js
@@ -21,9 +21,16 @@ export function restConnector(uri = 'http://localhost:3000/') {
         body: JSON.stringify(query)
       });
 
+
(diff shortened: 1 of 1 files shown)


domoritz (approved):
Thank you. Looks good and throwing an error will reject the promise.

PR #653, opened 2025-01-18
chore: vitest workspaces

domoritz:
Uses https://vitest.dev/guide/#workspaces-support to give a single entry point for running tests. The previous way of running tests in a package still works. 

Screenshot 2025-01-17 at 21 29 14

Needs https://github.com/uwdata/mosaic/pull/654/


jheer (approved):
LGTM

PR #677, opened 2025-02-04
feat: makeClient API

### LINKED ISSUE #394: Improved Documentation for Building Clients
Create documentation for developers who want to write their own client with React, Svelte, or plain JavaScript. Create example projects that they could use to get started.

domoritz:
A contribution from Apple (via @donghaoren and @domoritz).

Add a `makeClient` API to make it easier to build Mosaic views with modern frameworks such as React or Svelte.

Fixes #394 and https://github.com/uwdata/mosaic/discussions/482. 

Here is how one might use it in Svelte 5 syntax. This examples shows a simple sum of a given field.

```svelte
<script>
let { table, field, filter, coordinator } = $props();

let currentSum = $state(null);

$effect(() => {
    // Create the Mosaic client, and re-create it whenever any of the table, field,
    // filter, coordinator props change. Note that for the filter, a "change" here means
    // the filter is set to a new instance of Selection, rather than a change of the
    // current selection's value.
    let client = makeClient({
        coordinator: coordinator,
        selection: filter,
        prepare: async () => {
            // Obtain any information needed before running queries.
            let infos = await coordinator.query(...); // query directly.
            let fieldInfos = await coordinator.queryFieldInfos(...); // or use the field info API for commonly used information.
            // Now you can set Svelte state or variables that the query method use.
        },
        query: (predicate) => {
            return Query.from(table)
                .select({ "sum": sql`SUM(${column(field)})` })
                .where(predicate);
        },
        queryResult: (data) => {
            currentSum = data[0].sum;
        },
        queryError: (error) => {
            console.error("query error", error);
        }
    });

    // Once the client is created, you may call the following on it:
    // client.requestUpdate(); // Tell the client to run a new query. This is throttled.
    // client.initialize(); // Re-initialize the client (prepare will be called again, and then an update is requested).

    return () => {
        // Destroy the previous client before the effect updates.
        client.destroy();
    }
});
</script>

<p>Sum of {field} is: {currentSum}</p>
```

Video of the example in the `svelte-example` folder.

https://github.com/user-attachments/assets/f82a17d4-f4fc-41eb-8404-c12b813b18d9

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -34,7 +34,8 @@ export default defineConfig({
           { text: 'Mosaic vgplot', link: '/vgplot/' },
           { text: 'Mosaic Spec', link: '/spec/' },
           { text: 'Mosaic Server', link: '/server/' },
-          { text: 'Jupyter Widget', link: '/jupyter/' }
+          { text: 'Jupyter Widget', link: '/jupyter/' },
+          { text: 'Developing Clients', link: '/web-frameworks/' }
         ]
       },
       { text: 'Examples', link: '/examples/' },
@@ -62,7 +63,8 @@ export default defineConfig({
             { text: 'Mosaic vgplot', link: '/vgplot/' },
             { text: 'Mosaic Spec', link: '/spec/' },
             { text: 'Mosaic Server', link: '/server/' },
-            { text: 'Jupyter Widget', link: '/jupyter/' }
+            { text: 'Jupyter Widget', link: '/jupyter/' },
+            { text: 'Developing Clients', link: '/web-frameworks/' }
           ]
         },
         { text: 'Examples', link: '/examples/' },
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -35,7 +35,7 @@ export default defineConfig({
           { text: 'Mosaic Spec', link: '/spec/' },
           { text: 'Mosaic Server', link: '/server/' },
           { text: 'Jupyter Widget', link: '/jupyter/' },
-          { text: 'Developing Clients', link: '/web-frameworks/' }
+          { text: 'Web Clients', link: '/web-clients/' }
         ]
       },
       { text: 'Examples', link: '/examples/' },
@@ -64,7 +64,7 @@ export default defineConfig({
             { text: 'Mosaic Spec', link: '/spec/' },
             { text: 'Mosaic Server', link: '/server/' },
             { text: 'Jupyter Widget', link: '/jupyter/' },
-            { text: 'Developing Clients', link: '/web-frameworks/' }
+            { text: 'Web Clients', link: '/web-clients/' }
           ]
         },
         { text: 'Examples', link: '/examples/' },
--- a/docs/api/core/client.md
+++ b/docs/api/core/client.md
@@ -5,7 +5,7 @@ Clients are responsible for publishing their data needs and performing data proc
 
 If you are interested in creating your own Mosaic clients, see the [Mosaic GitHub repository](https://github.com/uwdata/mosaic).
 For concrete examples, start with the source code of [Mosaic inputs](https://github.com/uwdata/mosaic/tree/main/packages/inputs/src).
-If you are creating clients for web frameworks such as React, Svelte, or Vue, see the [web frameworks](/web-frameworks) documentation.
+If you are creating clients for web frameworks such as React, Svelte, or Vue, see the [web clients](/web-clients) documentation.
 
 ## constructor
 
--- a/docs/api/core/client.md
+++ b/docs/api/core/client.md
@@ -5,6 +5,7 @@ Clients are responsible for publishing their data needs and performing data proc
 
 If you are interested in creating your own Mosaic clients, see the [Mosaic GitHub repository](https://github.com/uwdata/mosaic).
 For concrete examples, start with the source code of [Mosaic inputs](https://github.com/uwdata/mosaic/tree/main/packages/inputs/src).
+If you are creating clients for web frameworks such as React, Svelte, or Vue, see the [web frameworks](/web-frameworks) documentation.
 
 ## constructor
 
--- a/docs/web-clients/index.md
+++ b/docs/web-clients/index.md
@@ -1,4 +1,4 @@
-# Developing Clients for Web Frameworks
+# Clients for Web Frameworks
 
 [Clients](/core/#clients) are responsible for publishing their data needs and performing data processing tasks&mdash;such as rendering a visualization.
 The [`MosaicClient` class](/api/core/client.html) provides a base class for client implementation.
--- a/docs/web-clients/index.md
+++ b/docs/web-clients/index.md
@@ -119,7 +119,7 @@ export function Count(props) {
         // Preparation work before the client starts.
         // Here we get the total number of rows in the table.
         const result = await coordinator.query(
-          Query.from(tableName).select({ count: count() })
+          Query.from(table).select({ count: count() })
         );
         setTotalCount(result.get(0).count);
       },
@@ -127,7 +127,7 @@ export function Count(props) {
         // Returns a query to retrieve the data.
         // The `predicate` is the selection's predicate for this client.
         // Here we use it to get the filtered count.
-        return Query.from(tableName)
+        return Query.from(table)
           .select({ count: count() })
           .where(predicate);
       },
--- a/docs/web-frameworks/index.md
+++ b/docs/web-frameworks/index.md
@@ -1,7 +1,11 @@
-# Working with Web Frameworks
+# Developing Clients for Web Frameworks
+
+[Clients](/core/#clients) are responsible for publishing their data needs and performing data processing tasks&mdash;such as rendering a visualization.
+The [`MosaicClient` class](/api/core/client.html) provides a base class for client implementation.
 
 Many applications are written in web frameworks such as [React](https://react.dev), [Svelte](https://svelte.dev), and [Vue](https://vuejs.org).
-Mosaic has a `makeClient` API that makes it smoother to integrate with such frameworks.
+These frameworks have their own state management and lifecycle management systems which make it difficult to use the `MosaicClient` class directly.
+Mosaic has a `makeClient` API that makes it easier to integrate with such frameworks.
 
 ## Svelte Example
 
@@ -15,7 +19,7 @@ Here is an example `Count` component written in Svelte, with `$effect`:
   import { makeClient } from "@uwdata/mosaic-core";
   import { count, Query } from "@uwdata/mosaic-sql";
 
-  let { coordinator, table, selection } = $props();
+  const { coordinator, table, selection } = $props();
 
   let totalCount = $state(null);
   let filteredCount = $state(null);
@@ -97,29 +101,29 @@ import { useState, useEffect } from "react";
 /** Show the number of rows in the table.
  * If a `selection` is provided, show the filtered number of rows as well. */
 export function Count(props) {
-  let { coordinator, table, selection } = props;
+  const { coordinator, table, selection } = props;
 
-  let [totalCount, setTotalCount] = useState(null);
-  let [filteredCount, setFilteredCount] = useState(null);
-  let [isError, setIsError] = useState(false);
-  let [isPending, setIsPending] = useState(false);
+  const [totalCount, setTotalCount] = useState(null);
+  const [filteredCount, setFilteredCount] = useState(null);
+  const [isError, setIsError] = useState(false);
+  const [isPending, setIsPending] = useState(false);
 
   useEffect(() => {
     // Capture tableName so Svelte keeps track of it in the effect.
     // When `table` changes, Svelte will re-run the effect and cause the old client
     // be destroyed and a new client be created.
-    let tableName = table;
+    const tableName = table;
 
     // Note that the identity of `selection` is also captured below.
     // If it is replaced with a new instance of Selection, the client will get recreated as well.
 
-    let client = makeClient({
+    const client = makeClient({
       coordinator,
       selection,
       prepare: async () => {
         // Preparation work before the client starts.
         // Here we get the total number of rows in the table.
-        let result = await coordinator.query(
+        const result = await coordinator.query(
           Query.from(tableName).select({ count: count() })
         );
         setTotalCount(result.get(0).count);
--- a/docs/web-frameworks/index.md
+++ b/docs/web-frameworks/index.md
@@ -109,13 +109,8 @@ export function Count(props) {
   const [isPending, setIsPending] = useState(false);
 
   useEffect(() => {
-    // Capture tableName so Svelte keeps track of it in the effect.
-    // When `table` changes, Svelte will re-run the effect and cause the old client
-    // be destroyed and a new client be created.
-    const tableName = table;
-
-    // Note that the identity of `selection` is also captured below.
-    // If it is replaced with a new instance of Selection, the client will get recreated as well.
+    // Note that the identity of `table` and `selection` is captured below.
+    // If they are replaced with a new instances, the client will get recreated as well.
 
     const client = makeClient({
       coordinator,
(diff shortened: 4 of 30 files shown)


packages/svelte-example/src/react/Count.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+/** Show the number of rows in the table.
+ * If a `selection` is provided, show the filtered number of rows as well. */
+export function Count(props) {
+  let { coordinator, table, selection } = props;
+
+  let [totalCount, setTotalCount] = useState(null);
+  let [filteredCount, setFilteredCount] = useState(null);
+  let [isError, setIsError] = useState(false);
+  let [isPending, setIsPending] = useState(false);
```
kwonoh: I think most of these variables can be `const`

packages/svelte-example/src/react/Count.jsx
```diff
@@ -0,0 +1,88 @@
...
+        setIsError(false);
+      },
+      queryError: () => {
+        // There is an error running the query.
+        setIsPending(false);
+        setIsError(true);
+      },
+    });
+
+    return () => {
```
kwonoh, line 62: I believe clearing selection should be added here as well.
```js
      // Clear the selection before destroying the client.
      selection.update({
        source: client,
        value: null,
        predicate: null,
      });
```

docs/web-frameworks/index.md
```diff
@@ -0,0 +1,167 @@
...
+        isError = false;
+      },
+      queryError: () => {
+        // There is an error running the query.
+        isPending = false;
+        isError = true;
+      },
+    });
+
+    return () => {
```
kwonoh, line 71: I think clearing selection should be added in the doc examples as well.

docs/web-frameworks/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+/** Show the number of rows in the table.
+ * If a `selection` is provided, show the filtered number of rows as well. */
+export function Count(props) {
+  let { coordinator, table, selection } = props;
+
+  let [totalCount, setTotalCount] = useState(null);
+  let [filteredCount, setFilteredCount] = useState(null);
+  let [isError, setIsError] = useState(false);
+  let [isPending, setIsPending] = useState(false);
```
kwonoh: Most of these variables in this React example can be `const`. 
I believe [using `const` is the convention in React world](https://react.dev/reference/react/useState).

docs/web-frameworks/index.md
```diff
@@ -0,0 +1,167 @@
...
+        setIsError(false);
+      },
+      queryError: () => {
+        // There is an error running the query.
+        setIsPending(false);
+        setIsError(true);
+      },
+    });
+
+    return () => {
```
kwonoh, line 153: I think clearing selection should be added in the doc examples as well.
```js
      // Clear the selection before destroying the client.
      selection.update({
        source: client,
        value: null,
        predicate: null,
      });
```

packages/svelte-example/src/react/Count.jsx
domoritz: Updated

docs/web-frameworks/index.md
domoritz: Updated the examples and docs.

docs/.vitepress/config.js
```diff
@@ -34,7 +34,8 @@ export default defineConfig({
           { text: 'Mosaic vgplot', link: '/vgplot/' },
           { text: 'Mosaic Spec', link: '/spec/' },
           { text: 'Mosaic Server', link: '/server/' },
-          { text: 'Jupyter Widget', link: '/jupyter/' }
+          { text: 'Jupyter Widget', link: '/jupyter/' },
+          { text: 'Developing Clients', link: '/web-frameworks/' }
```
jheer: Maybe `Web Clients` would be a clearer title than "Developing Clients"? Or `Use in Web Frameworks`?

packages/core/src/MosaicClient.js
```diff
@@ -78,6 +78,13 @@ export class MosaicClient {
     return this;
   }
 
+  /**
+   * Prepare the client before start running queries. You may query additional
```
jheer: Typo. Could change comment to state "Prepare the client before the query() method is called."

packages/core/src/MosaicClient.js
```diff
@@ -78,6 +78,13 @@ export class MosaicClient {
...
+   * information in this function.
+   */
+  async prepare() {
```
jheer, line 84: Would love to hear a bit more about why this method is needed. It seems similar to the fieldInfo stage, but more general. This makes me wonder if a different mechanism might handle "preparatory" work without having two different routes.

packages/core/src/make-client.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+export function makeClient(options) {
+  const coordinator = options.coordinator ?? defaultCoordinator();
+  const client = new DynamicQueryClient({ ...options, coordinator });
+  coordinator.connect(client);
+  return client;
+}
+
+/** An internal class used to implement the makeClient API */
+class DynamicQueryClient extends MosaicClient {
```
jheer: "DynamicQuery" has a different meaning in data visualization. I wonder if a clearer name might be something like "ProxyClient".

packages/core/test/client.test.js
```diff
@@ -60,6 +60,8 @@ describe('MosaicClient', () => {
     coord.connect(client1);
     coord.connect(client2);
 
+    await Promise.all([client1.pending, client2.pending]);
```
jheer, line 63: This seems unrelated to other changes in this PR. Is it needed?

packages/svelte-example/README.md
```diff
@@ -0,0 +1,5 @@
+# Mosaic React / Svelte Example
```
jheer, line 1: I would prefer not to populate the "packages" folder with examples. (I've wanted to move or remove the Vega-Lite example for some time, but just haven't gotten around to it.)

Instead, I'd like to move examples to a separate mono-repo folder. For example, maybe we should create a top-level `examples` folder and put all example packages there instead.

jheer (changes requested):

packages/core/test/client.test.js
domoritz: It's needed because the additional asynchronous prepare call in the client initialization broke the tests.

packages/svelte-example/README.md
domoritz: Good idea. Can we do that as a follow up to keep this pull request focused on the new API?

docs/web-frameworks/index.md
```diff
@@ -0,0 +1,171 @@
...
+  const [totalCount, setTotalCount] = useState(null);
+  const [filteredCount, setFilteredCount] = useState(null);
+  const [isError, setIsError] = useState(false);
+  const [isPending, setIsPending] = useState(false);
+
+  useEffect(() => {
+    // Capture tableName so Svelte keeps track of it in the effect.
+    // When `table` changes, Svelte will re-run the effect and cause the old client
+    // be destroyed and a new client be created.
+    const tableName = table;
```
kwonoh: Since React's `useEffect` requires explicit dependencies, I don't think we need to introduce a new variable `tableName` here.

btw, the comment says `Svelte` not `React`

docs/web-frameworks/index.md
```diff
@@ -0,0 +1,171 @@
...
+        setIsPending(false);
+        setIsError(true);
+      },
+    });
+
+    return () => {
+      // Destroy the client on effect cleanup.
+      client.destroy();
+    };
+  }, [coordinator, table, selection]);
```
kwonoh: Overall, this looks great and is very similar to what I created for using Mosaic in React applications.

One thing to discuss: this example uses `useEffect` to create a new client whenever dependencies change. While this particular example has three dependencies that don’t change frequently, I’ve encountered more complex scenarios where additional dependencies affect `query` or `query*` callbacks in `makeClient`. 

For some cases, I created a custom hook that subscribes to and updates Mosaic's `Param`, allowing `query` to return different queries without creating a new client with `makeClient`. I have an example of this approach in React with a custom hook, which I’ll add to the docs or examples after this PR.

However, in some other cases, I had to store the created client as a `ref` in React (using [`useRef`](https://react.dev/reference/react/useRef)) and explicitly update `query` or `query*` callbacks, as frequently recreating the client led to unexpected rendering behaviors sometimes (I guess it was related to repeatedly calling `Coordinator.connect` and `Coordinator.disconnect` in a short time span, but not sure). I think it would be helpful to include documentation or examples about updating states that would change `query` or `query*` callbacks later.

kwonoh (approved):
Thanks for adding this. Overall, looks great! I have added a few comments.

packages/core/src/MosaicClient.js
donghaoren: Yes, the idea is a more general `fieldInfo` stage where you can run arbitrary queries to gather information before the client's first `query` is issued. This stage is a bit tedious to implement outside the client. Here is an example how you'd do it without a `prepare` stage:

```js
// Without "prepare" we need a nested $effect:
$effect(() => {
  let tableName = table;

  // Define median as a state
  let median = $state(null);

  coordinator.query(
    Query.from(tableName).select({ median: median() })
  ).then((r) => {
    median = r.get(0).median;
  })

  $effect(() => {
    let client = makeClient({
      coordinator,
      selection,
      query: (predicate) => {
        return Query.from(tableName)
          .select({ avg: column("value") })
          .where(predicate, gt(column("value"), median));
      },
      queryResult: (data) => {
        // update states for display
      },
    });
    return () => {
      client.destroy();
    };   
  });
});
```

As you can see a nested `$effect` is needed. With the `prepare` function we can avoid the nested effect:

```js
// With "prepare" we can use a single $effect:
$effect(() => {
  let tableName = table;

  let median;

  let client = makeClient({
    coordinator,
    selection,
    prepare: async () => {
      let r = await coordinator.query(
        Query.from(tableName).select({ median: median() })
      );
      median = r.get(0).median;
    },
    query: (predicate) => {
      // Query is called after prepare is done, so `median` is guaranteed to have its value now.
      return Query.from(tableName)
        .select({ avg: column("value") })
        .where(predicate, gt(column("value"), median));
    },
    queryResult: (data) => {
      // update states for display
    },
  });
  return () => {
    client.destroy();
  };   
});
```

> This makes me wonder if a different mechanism might handle "preparatory" work without having two different routes.

Perhaps we can call `queryFieldInfos` from `prepare` (and remove the fieldInfo stage)?

```js
prepare: async () => {
    fieldInfos = await queryFieldInfos(...)
}
```

It will be a API-breaking change though.

docs/web-frameworks/index.md
donghaoren: > scenarios where additional dependencies affect query or query* callbacks in makeClient.

Yes this can be tricky. In this PR we did not consider how to preserve the client instance while allowing query to be changed based on a state / param.

I think directly using a state / param to make `query` return different queries can potentially lead to problems --- if the state / param is changed after `query` is called but before `queryResult`, the code in `queryResult` might see a mismatch between the the state / param and the query's result. To address this the developer would need to write additional code. I'm wondering if there can be some first-class API to support this kind of usage?

docs/web-frameworks/index.md
domoritz: Thanks. Updated!

packages/core/src/make-client.js
domoritz: Renamed

packages/core/src/MosaicClient.js
domoritz: Updated

docs/.vitepress/config.js
domoritz: Updated and moved the docs

docs/web-frameworks/index.md
domoritz: I think this is the last open thread before this is ready to be merged. So you think we should make any changes now?

kwonoh: @domoritz I think the current PR is good to merge. We can this discuss the best practices for such scenarios later.

> I think directly using a state / param to make query return different queries can potentially lead to problems --- if the state / param is changed after query is called but before queryResult, the code in queryResult might see a mismatch between the the state / param and the query's result.

@donghaoren I have used the pending state with `queryPending` to determine whether the currently available data is up to date or if newer data (or an error) is pending. Even if `query` doesn't change, there will still be a pending state the first time.

> I'm wondering if there can be some first-class API to support this kind of usage?

Having a canonical way to handle such scenarios would be ideal. I will share a few React examples after this PR is merged and deployed.

docs/api/core/client.md
```diff
@@ -5,6 +5,7 @@ Clients are responsible for publishing their data needs and performing data proc
 
 If you are interested in creating your own Mosaic clients, see the [Mosaic GitHub repository](https://github.com/uwdata/mosaic).
 For concrete examples, start with the source code of [Mosaic inputs](https://github.com/uwdata/mosaic/tree/main/packages/inputs/src).
+If you are creating clients for web frameworks such as React, Svelte, or Vue, see the [web frameworks](/web-frameworks) documentation.
```
kwonoh: @domoritz Could you double check the links for the new documentation page?

The links in the `docs/api/core/client.md` and `docs/core/index.md` are `/web-frameworks` but the links in the `docs/.vitepress/config.js` are `/web-clients` and also the new doc file is under `docs/web-clients/index.md`

docs/web-clients/index.md
```diff
@@ -0,0 +1,166 @@
...
+    // If they are replaced with a new instances, the client will get recreated as well.
+
+    const client = makeClient({
+      coordinator,
+      selection,
+      prepare: async () => {
+        // Preparation work before the client starts.
+        // Here we get the total number of rows in the table.
+        const result = await coordinator.query(
+          Query.from(tableName).select({ count: count() })
```
kwonoh: `tableName` -> `table`

docs/web-clients/index.md
```diff
@@ -0,0 +1,166 @@
...
+        );
+        setTotalCount(result.get(0).count);
+      },
+      query: (predicate) => {
+        // Returns a query to retrieve the data.
+        // The `predicate` is the selection's predicate for this client.
+        // Here we use it to get the filtered count.
+        return Query.from(tableName)
```
kwonoh: `tableName` -> `table`

kwonoh (changes requested):
Overall looks great! But I found a typo in the React example in the documentation. Also, one question about the documentation links

docs/web-clients/index.md
domoritz: Fixed. Do we need to bind table in https://github.com/uwdata/mosaic/pull/677/files/9277b7db9d716cdd2faff77725d3d26fbd6ed6df#diff-e438f5221b2978c32ff11040bb5b2598ec57d36b13a3b33a651316df79125639R116 or is that good as is now?

docs/api/core/client.md
domoritz: Fixed

docs/web-clients/index.md
domoritz: Fixed

domoritz:
@jheer addressed the comments and this looks good to go. We can iterate on the API a bit more in follow up prs (@kwonoh had some ideas).

packages/core/src/MosaicClient.js
jheer: Thanks. I think we can add this stage for now, but I suspect we'll want to make a breaking change in short order (via a future PR) to consolidate around a general `prepare` step, within which clients might choose to use the `fieldInfo` methods as a convenience.

jheer (approved):
LGTM. Thanks!!

docs/web-clients/index.md
kwonoh: I'm not sure which line you're referring to in the link. The highlighted line is 116 from the link, which contains `coordinator`. However, the React example looks good.

kwonoh (approved):

PR #678, opened 2025-02-05
fix: don't crash client when the coordinator has been removed

domoritz:
When the client is disconnected, the coordinator does not exist anymore but queries may still come back. We should try not to crash.


domoritz:
I was able to prevent this crash from happening so closing for now. Need to understand a bit better when this happens.

PR #681, opened 2025-02-07
feat: experiment - multi database support via Ibis

peter-gy:
> [!IMPORTANT]  
> The sole purpose of this pull request is to serve as a concrete point of discussion regarding issue #399 and to evaluate whether using [SQLGlot](https://github.com/tobymao/sqlglot) through [Ibis](https://github.com/ibis-project/ibis) to achieve multi-database support is a viable option _while introducing minimal differences in both the number of files and lines of code_.

## What problem do we aim to address?

As outlined in #399, it would be beneficial for Mosaic to support backends other than DuckDB. Currently, the only non-DuckDB backends supported by Mosaic are those accessed through DuckDB's `ATTACH` statement, namely MySQL, PostgreSQL, and SQLite.

## Proposal Overview

This pull request demonstrates how [Ibis](https://github.com/ibis-project/ibis) allows the use of backends like ClickHouse, PostgreSQL, DataFusion, and even Polars, which is a purely dataframe-based backend, while preserving 100% native support for the default DuckDB backend.

The main idea is to translate the queries from Mosaic clients into Ibis expressions, allowing Ibis to act as a bridge between Mosaic and approximately 20 backends. This approach eliminates the need to maintain adapters or connectors on a per-backend basis within Mosaic's codebase.

[CleanShot 2025-02-07 at 19 50 25@2x](https://ibis-project.org/)

Under the hood, Ibis converts its native internal representation (similar to Mosaic's SQL AST) into backend-specific queries using SQLGlot expressions. It also supports converting arbitrary SQL into Ibis expressions, enabling flows such as:

```mermaid
graph LR
    A[SQL in DuckDB dialect] --> B[Ibis Expression]
    B --> C[SQL in ClickHouse dialect]
    B --> D[SQL in PostgreSQL dialect]
    B --> E[SQL in DataFusion dialect]
    B --> F[Polars expression]
```

In fact, this is precisely the flow implemented in this pull request, as Mosaic clients express their data needs using the Mosaic SQL AST, which in turn gets converted into SQL in DuckDB dialect, resulting in:

```mermaid
graph LR
    X[Mosaic Client Data Need] --> Y[Mosaic SQL AST]
    Y --> A

    subgraph Ibis
        A[SQL in DuckDB dialect] --> B[Ibis Expression]
        B --> C[SQL in ClickHouse dialect]
        B --> D[SQL in PostgreSQL dialect]
        B --> E[SQL in DataFusion dialect]
        B --> F[Polars expression]
    end
```

More on Ibis internals: https://ibis-project.org/posts/does-ibis-understand-sql/

### Backend Setup via Docker

**PostgreSQL**

```shell
docker run --rm \
  --name mosaic-postgres \
  -e POSTGRES_PASSWORD=mosaic \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_DB=postgres \
  -p 5432:5432 \
  postgres:15
```

**ClickHouse**

```shell
docker run --rm \
  --name mosaic-clickhouse \
  -p 8123:8123 \
  clickhouse/clickhouse-server:latest
```

1. Switch to this branch.
2. Run `uv sync` to pull core Ibis dependencies and database.
3. Start the Vite development server using `npm run dev`, and choose `Socket` or `REST` connector.
4. Start the server with a specific backend specified as the connection string.

- PostgreSQL: `npm run server -- postgres://postgres:mosaic@localhost:5432/postgres`
- ClickHouse: `npm run server -- clickhouse://`
- DataFusion: `npm run server -- datafusion://`
- Polars: `npm run server -- polars://`
- DuckDB: `npm run server -- duckdb://`

> [!NOTE]  
> Due to the limitations outlined below, many of the example specs will not be fully functional when using backends other than DuckDB

The following screencast demonstrates how various backends can be used to serve Mosaic clients with data.

https://github.com/user-attachments/assets/3431ce60-85ce-4992-9255-687875115dc9

### Expression Compatibility

This pull request presents a quick and preliminary proof of concept. Many specifications still only work with DuckDB, as clients express their data needs using DuckDB-specific SQL syntax that is unsupported by SQLGlot, or they use native DuckDB functions that SQLGlot does not recognize. To fully leverage Ibis and the operations it supports (https://ibis-project.org/backends/support/matrix), the Mosaic AST should ideally be translated into a representation that allows for direct reconstruction of Ibis expressions on the Python side. This would simplify the process to:

```mermaid
graph LR
    X[Mosaic Client Data Need] --> Y[Mosaic SQL AST]
    Y --> B[Ibis Expression]

    subgraph Ibis
        B --> C[SQL in ClickHouse dialect]
        B --> D[SQL in PostgreSQL dialect]
        B --> E[SQL in DataFusion dialect]
        B --> F[Polars expression]
    end
```

In other words, the current flow from `Mosaic SQL AST` to `DuckDB` to `Ibis Expression` to the respective backend—where the `DuckDB to Ibis` segment is managed by SQLGlot—is fragile. Directly converting from `Mosaic SQL AST` to `Ibis Expression` would enhance multi-backend compatibility. However, this would require a significant refactoring of the Mosaic SQL package, and selecting an Ibis internal representation as the lingua franca for Mosaic queries would restrict usage to the Python server, as there are no Ibis bindings available for Rust or Node.js.

### Data Loading

The introduced data loading mechanism is quite basic. Based on the observed use cases, the Mosaic Coordinator calls `exec` to perform table/schema creation and deletion operations, while the `query` function retrieves data from existing tables. We take advantage of this by using a dedicated in-memory DuckDB connection to load data according to the specified expression, followed by `self.backend.create_table(table_name, table.to_pyarrow(), overwrite=True)` to create the same table in the target backend.

Building serialized Ibis expressions on the client side and deserializing them on the server side would be beneficial, as Ibis supports dedicated data loading methods (`read_csv`, `read_parquet`, `read_json`, `read_delta`) and can execute more complex expressions (including raw backend-native SQL) to create tables.

### Ibis as a Meta-Connector

The changes introduced in this pull request present a rough solution for how the default DuckDB backend could coexist with the Ibis meta-backend within the same server package. However, it could also be more aligned with Mosaic's philosophy of composable components to develop this solution as a standalone connector, making less radical adjustments to other packages on an as-needed basis.

### DIFF
--- a/packages/core/src/preagg/PreAggregator.js
+++ b/packages/core/src/preagg/PreAggregator.js
@@ -316,7 +316,10 @@ function preaggregateInfo(clientQuery, active, preaggCols, schema) {
   // generate creation query string and hash id
   const create = query.toString();
   const id = (fnv_hash(create) >>> 0).toString(16);
-  const table = `${schema}.preagg_${id}`;
+  
+  // Simplification of Ibis experiment by ensuring all tables land in the default schema
+  // const table = `${schema}.preagg_${id}`;
+  const table = `preagg_${id}`;
 
   // generate preaggregate select query
   const select = Query
--- a/packages/duckdb-server/pkg/__main__.py
+++ b/packages/duckdb-server/pkg/__main__.py
@@ -1,7 +1,8 @@
 import logging
 import sys
 
-import duckdb
+import ibis
+import sqlglot
 from diskcache import Cache
 
 from pkg.server import server
@@ -10,12 +11,54 @@ logger = logging.getLogger(__name__)
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
 
 
+def process_query(query: str) -> str:
+    logger.debug(f"Processing query: {query}")
+    target_query = sqlglot.transpile(query, read="duckdb", write="postgres")[0]
+    logger.debug(f"Transpiled query: {target_query}")
+    return target_query
+
+
+class MosaicConnection:
+    def __init__(self, backend: ibis.BaseBackend):
+        self.backend = backend
+        self.duckdb = ibis.duckdb.connect()
+
+    def query(self, query: str):
+        # We query duckdb natively when encountering a describe statement or when backend is set to duckdb
+        # SQLGlot parses DESCRIBE statements under the hood, but packages/core/src/util/field-info.js expects responses in duckdb's format
+        if query.lower().startswith("describe") or self.backend.name == "duckdb":
+            res = self.duckdb.con.query(query).arrow()
+            # We still return an ibis.Table object to keep the interface consistent
+            return ibis.memtable(res)
+
+        # In all other cases we go through the Ibis backend's SQL API
+        # DuckDB query -> Ibis internal query representation -> Backend-specific query
+        table = self.backend.sql(query, dialect="duckdb")
+        logger.debug(f'Ibis internal representation: {table}')
+        return table
+
+    def execute(self, query: str):
+        # For this experiment we create all tables in DuckDB first, then sync them to the backend
+        # (many backends do not support read_parquet, do not have access to local files, etc.)
+        execute_res = self.duckdb.con.execute(query)
+        self.sync_tables()
+        return execute_res
+
+    def sync_tables(self):
+        # We use the DuckDB backend instance as our catalog of tables and copy them to the target backend
+        for table_name in self.duckdb.list_tables():
+            table = self.duckdb.table(table_name)
+            self.backend.create_table(table_name, table.to_pyarrow(), overwrite=True)
+
+
 def serve():
-    db_path = sys.argv[1] if len(sys.argv) >= 2 else ":memory:"  # noqa: PLR2004
+    backend_resource = sys.argv[1] if len(sys.argv) >= 2 else "duckdb://:memory:"  # noqa: PLR2004
 
-    logger.info(f"Using DuckDB {db_path}")
+    logger.info(f"Using backend resource {backend_resource}")
+    backend = ibis.connect(backend_resource)
 
-    con = duckdb.connect(db_path)
+    con = MosaicConnection(backend)
+
     cache = Cache()
 
     logger.info(f"Caching in {cache.directory}")
--- a/packages/duckdb-server/pkg/__main__.py
+++ b/packages/duckdb-server/pkg/__main__.py
@@ -33,6 +33,7 @@ class MosaicConnection:
 
         # In all other cases we go through the Ibis backend's SQL API
         # DuckDB query -> Ibis internal query representation -> Backend-specific query
+        logger.debug(f'Executing query on backend: {self.backend.name}')
         table = self.backend.sql(query, dialect="duckdb")
         logger.debug(f'Ibis internal representation: {table}')
         return table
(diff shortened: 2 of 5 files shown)


jheer:
Thanks for preparing this!! It is a great proof-of-concept, and I certainly understand the current workarounds for data loading. I love how minimal the changes were to get this working.

Thinking further ahead, one obvious limitation of this approach is the requirement of using a local Python server as a proxy to other backends. This is viable in some contexts, but ultimately we would like to support multiple databases directly from JavaScript as well. It would be particularly great if there is a SQLGlot equivalent available in either JS or WASM... or, perhaps learning from SQLGlot to expand our existing mosaic-sql package with support for other SQL dialects.

peter-gy:
Thank you for the feedback.

> one obvious limitation of this approach is the requirement of using a local Python server as a proxy to other backends [...] ultimately we would like to support multiple databases directly from JavaScript as well

Do you see the dependence on a server written in Python instead of JavaScript as the limitation here or the sheer presence of any serverful proxy component between the browser and backends other than DuckDB? My understanding is that since databases such as Postgres, BigQuery, etc. rely on binary protocols and dedicated drivers, short of a per-backend WASM implementation it is not an option to establish direct connection to them from a browser tab via JavaScript.

> It would be particularly great if there is a SQLGlot equivalent available in either JS or WASM...

SQLGlot itself is available in WASM through Pyodide, as is Ibis: https://ibis-project.org/tutorials/browser/repl.
Similar JS-based packages I am aware of include https://github.com/knex/knex and https://github.com/taozhi8833998/node-sql-parser, but SQLGlot seems to be by far the most comprehensive solution for converting between SQL dialects. It strictly focuses on the AST,  and is not coupled with backend-specific execution logic.

The way I see it, to support multiple databases, roughly speaking we need to solve two main problems. (1) convert Mosaic AST into a backend-specific expression and (2) execute this expression against the chosen backend. For (1) we could do everything on the client-side, as only symbolic manipulation would be going on for which a browser environment with JS / WASM is suitable. Regarding (2), for backends unlike DuckDB, without native WASM support I see the presence of a proxy as necessary but I might be missing use cases here.

jheer:
> The way I see it, to support multiple databases, roughly speaking we need to solve two main problems. (1) convert Mosaic AST into a backend-specific expression and (2) execute this expression against the chosen backend. For (1) we could do everything on the client-side, as only symbolic manipulation would be going on for which a browser environment with JS / WASM is suitable. Regarding (2), for backends unlike DuckDB, without native WASM support I see the presence of a proxy as necessary but I might be missing use cases here.

At least for those backends that are (1) network accessible and (2) can return results in a Mosaic-friendly format (for now, that's primarily Apache Arrow), I would want to retain the ability to have the browser connect to the (presumably preexisting) backend directly and not have to setup an intermediate server to proxy between the browser and the backend. (This is not about JS vs. Python per se, it's about the need for an additional intermediate service layer.)

peter-gy:
Thanks for the clarification, I see what you mean. ClickHouse seems to be a database similar to the kind you described, as it provides an HTTP interface and can return results in Arrow format. I'll think about how a generic integration could be designed along those lines.

An additional data point to consider around using Mosaic with multiple databases is that the Apache Arrow Flight protocol is already supported by backends such as Dremio, Apache Doris, Postgres, InfluxDB and hopefully many more in the future. This might be interesting as https://github.com/Query-farm/duckdb-airport-extension enables the use of Arrow Flight in DuckDB, allowing direct connections to Arrow Flight-compatible backends through DuckDB, even if no DuckDB `ATTACH` support is implemented in the core DuckDB codebase. However, similar to the `ATTACH`-based approach, this approach using the `duckdb-airport-extension` would also not be available via DuckDB WASM because the extension is not (yet? Query-farm/duckdb-airport-extension#12) distributed for WebAssembly architectures.

PR #682, opened 2025-02-09
feat: Add publish package

rpechuk:
This PR contains and MVP for the new `Publish` CLI tool that will be used to transform YAML/JSON specs into shippable visualizations.

## Current Changes:
- Introduced `Input` and `Interactor` base classes, refactored `Menu`, `Search`, `Slider`, `Table`, and all interactors to extend from these for consistent activation behavior (added common interface for `activate` function).
- Added `mosaic-publish` CLI tool for compiling and optimizing specs, handling data export, interactivity, and optimizations.

## TODOs:
- [ ] Add more user control over optimization params/streamline current user control (selective dataset inclusion, DuckDB export settings, granular optimization flags, etc.).
- [x] Refactor duplicated logic between `MosaicClient` and new base classes. Namely combine both base classes into one base class that discribes any activatable client.
    - Open Question: Where should this be created since these span multiple `packages`
- [ ] Expand README and add tests for CLI and utilities
- Add further optimization and improve upon current optimizations:
    - Namely work on better pre-rendering + hydration
    - Automatic Tiling implementation
    - Cache hydration
    - Smart spec re-writing
    - etc.

## Open Questions:
- Should `Input` and `Interactor` remain separate or be unified?
- How should we handle interactive pre-rendered visualizations before data is loaded?
- Are there any better/other output formats we shoudl support?
- Is the current way of handling dependencies correct (w/o relative reference to package and instead using npm version)?


rpechuk:
Following up on #705 does it make sense to move away from CDN imports in published visualization code to avoid having to cut a release of other packages for a `publish` package release?

IMO, this has both benifts and downsides so maybe it makes sense to allow end user to choose and parametrize available optimizations based on user choice and available functionality.

PR #683, opened 2025-02-09
feat: consolidated activatable logic to core

rpechuk:
## Changes:

- Add `Activatable` interface to `mosaic-core` types, requires an `activate()` method for selection activation.
- Add `isActivatable` type check method to `mosaic-core`.
- Update inputs (`Menu`, `Search`, `Slider`, `Table`) to implement the `Activatable` interface.
- Update vgplot interactors to implement the `Activatable` interface.
- Refactor type exports for `mosaic-core`.


packages/core/src/ActivatableClient.ts
```diff
@@ -0,0 +1,23 @@
+import { MosaicClient } from "./MosaicClient.js";
+
+/**
+ * Interface for activatable objects.
+ * @see {@link ActivatableClient}
+ * @see {@link ActivatableClient.activate}
+ */
+export abstract class Activatable {
```
jheer: Why is this an `abstract class`? I'd expect it to be an `interface` instead.

packages/core/src/ActivatableClient.ts
```diff
@@ -0,0 +1,23 @@
...
+    abstract activate(): void;
+}
+
+/**
+ * Base class for activatable Mosaic clients.
+ */
+export class ActivatableClient extends MosaicClient implements Activatable {
+    /**
```
jheer: It looks like you are using 4 spaces for "tabs", but the Mosaic packages use 2 spaces. Please use styles consistent with the rest of the project.

packages/core/src/index.js
```diff
@@ -1,4 +1,5 @@
 export { MosaicClient } from './MosaicClient.js';
+export { Activatable, ActivatableClient } from './ActivatableClient.js';
```
jheer: This is interleaving TypeScript and JavaScript in ways that violate the current setup. Types and interfaces should be defined in TypeScript files, then imported for use (typically via JSDoc comments) for typing. Any actual classes and methods should be defined in JavaScript files.

jheer: The idea is that the code should be runnable as-is, without requiring `tsc` to have been run first.

packages/inputs/src/Menu.js
```diff
@@ -8,7 +8,7 @@ const isObject = v => {
 
 export const menu = options => input(Menu, options);
 
-export class Menu extends MosaicClient {
+export class Menu extends ActivatableClient {
```
jheer: I don't think ActivatableClient is needed here, as this class implements the full `Activatable` interface on its own. Typically I'd expect the class to have a top-level JSDoc comment that includes an annotation indicating that the `Menu` class implements the `Activatable` interface.

That said, I'm OK using a base class if it's helpful elsewhere for runtime type-checking (as opposed to duck typing checks for the `activate` method).

packages/inputs/src/Table.js
```diff
@@ -208,6 +208,23 @@ export class Table extends MosaicClient {
     return this;
   }
 
+  activate() {
```
jheer, line 218: What's the logic here? I see this calling selection update repeatedly for EVERY row and that does not look right... I would expect `selection.activate` (not `update`) to be called just once.

packages/inputs/src/input.js
```diff
@@ -4,4 +4,4 @@ export function input(InputClass, options) {
   const input = new InputClass(options);
   coordinator().connect(input);
   return input.element;
-}
+}
```
jheer: This looks like an unnecessary change?

jheer (changes requested):
Looks good overall, but see comments for requested changes regarding use of types (TS vs. JS) and also what the `table` input is doing.

packages/inputs/src/Menu.js
jheer: Another option is to provide an `isActivatable` utility method, perhaps something like this:

```js
/**
 * Test if a value implements the Activatable interface.
 * @param {any} The value to test
 * @returns {value is Activatable} true if activatable, false otherwise
 */
export function isActivatable(value) {
  return typeof value?.activate === 'function';
}
```

packages/core/src/ActivatableClient.ts
rpechuk: Great questions, originally I was under the impression that an abstract class would throw an error if the subclass didn't implement the `activate` function even in JS (due to how it would compile into JS), however after more investigation this doesn't actually seem to be the case so I will switch to a `interface` 👍

packages/core/src/index.js
rpechuk: Is the new way i'm doing it with `@typedef {import('./Activatable.js').Activatable} Activatable` in the `index.js` file and then the following above the classes that implement it the proper way to do it?

```
/**
 * @typedef {import('@uwdata/mosaic-core').Activatable} Activatable
 */

/**
 * @implements {Activatable}
 */
 ```
Note: Unless I explicitly include the `@typedef` this doesn't work and TS complains about import not being recognized.

Comment (jheer) [reply packages/core/src/index.js]: You can also do this instead:

```js
/**
 * @implements {import('@uwdata/mosaic-core').Activatable}
 */
```

rpechuk: Originally I tried this but got errors in tests and after many different attempts locally this seemed to be the only way I could think of to fix it. 
Errors in tests: https://github.com/uwdata/mosaic/actions/runs/13249311572/job/36983225553#step:7:186
Example in `Highlight.js`: https://github.com/uwdata/mosaic/pull/683/commits/a50afbaa17d67fd6a567a6d76a70ecc19a39d3b9#diff-ded8a826fd5daaa70cec86ccb7d6e8e2d652ee9108a1297dcfa801389813e812R38

packages/inputs/src/Menu.js
```diff
@@ -2,12 +2,21 @@ import { MosaicClient, Param, isParam, isSelection, clausePoint } from '@uwdata/
...
 const isObject = v => {
   return v && typeof v === 'object' && !Array.isArray(v);
 };
 
 export const menu = options => input(Menu, options);
 
+/**
+ * A HTML select based dropdown menu input.
+ * 
+ * @implements {Activatable}
```
domoritz, line 15: Just adding this doesn't work?

```suggestion
 * @implements {import('@uwdata/mosaic-core').Activatable}
```

rpechuk: Unfortunately no, it causes errors in the tests (specifically during running `tsc` in the spec package). See here for logs: https://github.com/uwdata/mosaic/actions/runs/13249311572/job/36983225553#step:7:186

packages/core/src/index.js
rpechuk: To resolve this unfortunately this doesn't seem to be supported :/

https://github.com/microsoft/TypeScript/issues/49905
https://github.com/microsoft/TypeScript/issues/58542

jheer: That is frustrating. It seems a bit annoying to have to add a typedef in every place we want to implement the Activatable interface. For the inputs package, maybe we should add an `Input` base class that extends `MosaicClient` and implements `Activatable`? That way you only have to set it up once...

rpechuk: I moved away from typedef since it was extremely verbose to a slightly less verbose method:
```js
 * @import {Activatable} from '@uwdata/mosaic-core'
 * @implements {Activatable}
 */
 ```
 
With that said I still think that there is a case to have `Input` and `Interactor` base classes that do this once and then the rest of the objects extend these to avoid duplicating these. Will defer to you on this.

Comment (jheer) [reply packages/core/src/index.js]: I think that works well. We can always create base classes in a future PR.

File: packages/core/src/Activatable.ts [hunk 8]
```diff
@@ -0,0 +1,10 @@
+import { isActivatable } from "./util/is-activatable.js";
+
+/**
+ * Interface for activatable objects.
+ * 
+ * @see {@link isActivatable}
```
jheer: I don't think this (or the import) are really needed here. So we could simplify that.

packages/core/src/types.ts
```diff
@@ -1,5 +1,8 @@
 import type { DescribeQuery, ExprNode, Query } from '@uwdata/mosaic-sql';
 
+export * from './util/selection-types.js';
```
jheer: We should move these exports to index-types.ts instead. I don't think we need them here.

We could also move the Activatable interface definition into this file.

packages/plot/src/interactors/Highlight.js
```diff
@@ -80,6 +84,10 @@ export class Highlight {
       }
     }
   }
+
+  activate() {
+    return this.update();
```
jheer: Why do we call update here?

packages/core/src/index.js
rpechuk: Sounds good! In that case I'll resolve this and if everything else is good we should be good to merge!

jheer (changes requested):
Looking good! Just a few remaining minor suggestions and a question to address.

packages/plot/src/interactors/Highlight.js
rpechuk: Good catch, that was a remnant I meant to remove. Originally I just had the activate function be a NOOP however thinking about it now since we have the `isActivatble` check we can also just remove the function all together since it doesn't actually make sense on the Highlight interactor. What do you think would be best?

jheer: Just remove it then!

PR #684, opened 2025-02-10
Add Aggregate Function `geomean` (mosaic-sql)

spren9er:
This pull request introduces a new aggregate function — `geomean` — to the `mosaic-sql` package for computing the geometric mean.

### What does this pull request cover?

- Add `geomean` aggregate function
- Add preaggregation handling (in `sufficient-statistics.js`)
- Add tests (in `preaggregator.test.js` and `aggregate.test.js`)
- Update documentation

### What is missing?

The JSON schema file has not yet been updated.
It is currently unclear whether this file will be auto-generated or if a manual modification is required.

### DIFF
--- a/docs/api/sql/aggregate-functions.md
+++ b/docs/api/sql/aggregate-functions.md
@@ -99,6 +99,12 @@ Create an aggregate function that calculates the sum of the input _expression_.
 
 Create an aggregate function that calculates the product of the input _expression_.
 
+## geomean
+
+`geomean(expression)`
+
+Create an aggregate function that calculates the geometric mean of the input _expression_.
+
 ## median
 
 `median(expression)`
--- a/packages/core/src/preagg/sufficient-statistics.js
+++ b/packages/core/src/preagg/sufficient-statistics.js
@@ -1,4 +1,4 @@
-import { AggregateNode, and, argmax, argmin, count, div, ExprNode, isNotNull, max, min, mul, pow, regrAvgX, regrAvgY, regrCount, sql, sqrt, sub, sum } from '@uwdata/mosaic-sql';
+import { AggregateNode, and, argmax, argmin, count, div, ExprNode, isNotNull, max, min, mul, pow, product, regrAvgX, regrAvgY, regrCount, sql, sqrt, sub, sum } from '@uwdata/mosaic-sql';
 import { fnv_hash } from '../util/hash.js';
 
 /**
@@ -18,6 +18,8 @@ export function sufficientStatistics(node, preagg, avg) {
       return sumExpr(preagg, node);
     case 'avg':
       return avgExpr(preagg, node);
+    case 'geomean':
+      return geomeanExpr(preagg, node);
     case 'arg_max':
       return argmaxExpr(preagg, node);
     case 'arg_min':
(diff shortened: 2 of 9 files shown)


docs/api/sql/aggregate-functions.md
```diff
@@ -99,6 +99,12 @@ Create an aggregate function that calculates the sum of the input _expression_.
 
 Create an aggregate function that calculates the product of the input _expression_.
 
+## geomean
+
+`geomean(expression)`
```
domoritz, line 104: should this be `geomean` or `geoMean` to be consistent with others?

domoritz, line 104: should this be `geomean` or `geoMean` to be consistent with others?

domoritz (approved):

docs/api/sql/aggregate-functions.md
jheer: I think it should be `geomean`, as this aligns with SQL use. For example, elsewhere we use `argmax`, not `argMax`.

jheer: I vote for `geomean` (keep it as-is)

jheer (approved):
This looks great. Thanks!

packages/core/src/preagg/sufficient-statistics.js
```diff
@@ -155,6 +157,21 @@ function avgExpr(preagg, node) {
...
+ * Generate an expression for calculating geometric means over data dimensions.
+ * As a side effect, this method adds a column to the input *preagg* object
+ * to track the count of non-null values per-partition.
+ * @param {Record<string, ExprNode>} preagg A map of columns (such as
+ *  sufficient statistics) to pre-aggregate.
+ * @param {AggregateNode} [node] The originating aggregate function call.
+ * @returns {ExprNode} An aggregate expression over pre-aggregated dimensions.
+ */
+function geomeanExpr(preagg, node) {
+  const as = addStat(preagg, node);
```
jheer: Should `node` also be passed as a third argument to `addStat`? Think about cases like `GEOMEAN(x) FILTER (x > 2)`.

packages/core/src/preagg/sufficient-statistics.js
```diff
@@ -155,6 +157,21 @@ function avgExpr(preagg, node) {
   return div(sum(mul(as, name)), expr);
 }
 
+/**
+ * Generate an expression for calculating geometric means over data dimensions.
+ * As a side effect, this method adds a column to the input *preagg* object
+ * to track the count of non-null values per-partition.
+ * @param {Record<string, ExprNode>} preagg A map of columns (such as
+ *  sufficient statistics) to pre-aggregate.
+ * @param {AggregateNode} [node] The originating aggregate function call.
```
jheer: I think `node` is a required argument, but here it is listed as an optional argument.

packages/core/src/preagg/sufficient-statistics.js
```diff
@@ -155,6 +157,21 @@ function avgExpr(preagg, node) {
...
+ * @returns {ExprNode} An aggregate expression over pre-aggregated dimensions.
+ */
+function geomeanExpr(preagg, node) {
+  const as = addStat(preagg, node);
+  const { expr, name } = countExpr(preagg, node);
+  return pow(product(pow(as, name)), div(1, expr));
```
jheer: This looks "theoretically" correct to me: given a set per-bin geometric means, exponentiate by the (bin-level) count to "undo" the bin-level nth-root, multiply the results, then take the (global-level) nth-root.

However, might this suffer from overflow? That intermediate product could get very large. (The same result as if we just used product as the sufficient statistic rather than geomean, which would also simplify the overall scheme to just a single `pow` call for the nth-root.)

This might be worth testing more with large numbers. If we see issues, a more robust alternative would be to instead use the sum of log values as the sufficient statistic. Then the output aggregate expression would be `exp(div(sum(as), expr))`.

jheer (changes requested):
Please see comments, particularly about adding tests for very large products and a possible solution if they reveal a problem with the current implementation.

packages/core/src/preagg/sufficient-statistics.js
spren9er: Yes, I agree.

The code is already updated as it is quite clear, that there will be numerical issues.

duckdb_geomean

Note: DuckDB uses log-based computation for `geomean` as well (see [here](https://github.com/duckdb/duckdb/blob/f317a112e0c789a11987d9de6d621d8469c00fb8/src/catalog/default/default_functions.cpp#L110)).

packages/core/src/preagg/sufficient-statistics.js
spren9er: Fixed

packages/core/src/preagg/sufficient-statistics.js
spren9er: Added to new sufficient statistic (sum of log values)

jheer (approved):
Looks great. Thanks @spren9er!

PR #685, opened 2025-02-13
Move example projects to dedicated folder.

jheer:
- Move Mosaic example projects to a dedicated `examples` folder.
- Update build/test configuration to target main packages only.


examples/vega-example/dist/assets/index-CvRx8YOT.js
domoritz, line 1: the dist folder should not be in git imo

domoritz (approved):

PR #686, opened 2025-02-14
Refactor inputs, add better types.

jheer:
- Add `Input` class, update inputs to inherit from it.
- Add types for input generation methods.
- Add types and docs for table options.


domoritz (approved):
Looks good but I think https://github.com/uwdata/mosaic/pull/687 is the cleaner way for vitest to get it working with workspaces and in individual packages.

rpechuk (approved):
LGTM! Should we do something similar with Interactors?

domoritz:
I removed the vitest changes from this pull request.

PR #688, opened 2025-02-14
feat: pure esm packages

domoritz:
Following the advice in https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c.

Previously, we declared that what we released bundles that could be imported in a browser via a good old script import but that's not true since you do need to import as an esm module. This modernizes our exports and should also make it easier to update how we bundle things.


domoritz:
I don't think there is much benefit in bundling given that jsdelivr does it for us if we use ESM. It also puts in the right versions for dependencies. Here is an example for vgplot: https://cdn.jsdelivr.net/npm/@uwdata/vgplot@0.12.2/+esm.

jheer (changes requested):
This looks pretty good, but looks like there is some lingering clean up to be done. In places where the `build` script is removed, there may still be a now unneeded `prebuild` script as well as a call to the `build` script as part of publishing. We could remove those for now, though ultimately we will want to "build" type definitions in those packages.

domoritz:
I cleaned up the calls and yes, we want to ultimately have the type build. I am also still thinking about how we can support typescript while not needing to build in the monorepo during development. There may be a way to do that with path rewriting in tsconfigs.

jheer (approved):

PR #692, opened 2025-02-16
feat: improve `ast-to-dom` typing

rpechuk:
Improved the typing of the `options` parameter of `ast-to-dom` to be based on `InstantiateContext` as this is what it is being used for.

### DIFF
--- a/packages/spec/src/ast-to-dom.js
+++ b/packages/spec/src/ast-to-dom.js
@@ -7,10 +7,7 @@ import { error } from './util.js';
 /**
  * Generate a running web application (DOM content) for a Mosaic spec AST.
  * @param {SpecNode} ast Mosaic AST root node.
- * @param {object} [options] Instantiation options.
- * @param {string} [options.baseURL] The base URL for loading data files.
- * @param {any[]} [options.plotDefaults] Array of default plot attributes.
- * @param {Map<string, Param>} [options.params] A map of predefined Params/Selections.
+ * @param {ConstructorParameters<typeof InstantiateContext>[0]} [options] Instantiation options.
  * @returns {Promise<{
  *   element: HTMLElement | SVGSVGElement;
 *    params: Map<string, Param | Selection>;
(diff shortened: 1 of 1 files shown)


jheer (approved):
Thanks, this looks good! However, if not too much hassle it would be great to also add proper types and comments for the `InstantiateContext` parameters below.

packages/publish/src/MosaicPublisher.ts
```diff
@@ -101,8 +101,7 @@ export class MosaicPublisher {
     mockCanvas(globalThis.window);
 
     // Load the visualization in the DOM and gather interactors/inputs
-    // TODO: fix type issue with astToDOM to remove the any cast
-    const { element } = await astToDOM(this.ast, { api: this.ctx.api } as any);
+    const { element } = await astToDOM(this.ast, { api: this.ctx.api });
```
jheer: I don't think this file should be in this PR, as it pertains to the WIP publish package.

jheer (changes requested):
Looks good except for the `publish` package file that snuck in.

packages/publish/src/MosaicPublisher.ts
rpechuk: Would it be best for me to remove these changes and make them in the publish package PR (changing the base of that PR to this branch and the base of this PR to main)? As of right now these changes would be merged into the publish package PR which is why I included this.

jheer: As I see. I was thinking of merging into `main`, hence the comment. I still think that would be ideal.

rpechuk: Gotcha, will address accordingly 👍

rpechuk: Should be fixed now! See #682 for above changes and they shouldn't be in this PR anymore

jheer (approved):

PR #695, opened 2025-02-17
feat: Add cte method for fine-grained WITH clause control.

### LINKED ISSUE #693: Enforcing the Use of Materialized CTEs in M4 Query
In PR #691, the M4 query was rewritten to use Common Table Expressions (CTEs) instead of subqueries. When CTEs are materialized, query performance can improve significantly because intermediate results are cached and reused.

DuckDB automatically materializes CTEs if

1. The CTE performs a grouped aggregation, and
2. The CTE is queried more than once.

See also [here](https://duckdb.org/docs/sql/query_syntax/with.html#cte-materialization).

The M4 query benefits greatly from materialization since the CTE is queried four times. However, materialization currently only applies when a CTE performs a grouped aggregation, even though it can also speed up non-aggregated queries.

For example, in the _Overview-Detail Chart_ example from Mosaic, where data is used as-is (without aggregation), the performance of the brushing action improves significantly when using a materialized M4 query — cutting execution time in half. Although the M4 query is already fast in this case (due to the small dataset), tests with 1 million data points show a noticeable improvement, transforming the animation from laggy to much smoother.

Therefore, it would be great to be able to enforce materialization for the M4 query. To achieve this, `mosaic-sql` should be updated to support `MATERIALIZED` (and `NOT MATERIALIZED` for completeness) for CTEs (`WithClauseNode`) to provide explicit control over materialization.

jheer:
- Add `cte(...)` method for direct construction of WITH clause nodes.
- Add `materialized` property to WITH clause nodes to control explicit CTE materialization.
- Update M4 transform to enforce CTE materialization.

Close #693.

### DIFF
--- a/docs/api/sql/expressions.md
+++ b/docs/api/sql/expressions.md
@@ -9,6 +9,15 @@ SQL expression builders. All SQL expressions are represented in the form of an a
 Create an expression AST node that references a column by _name_.
 Upon string coercion, the column name will be properly quoted.
 
+## cte
+
+`cte(name, query, materialized)`
+
+Create an AST node for a Common Table Expression (CTE) to be used within a SQL `WITH` clause.
+The resulting node can be passed as an argument to the `Query.with()` method.
+The string-valued _name_ and Query-valued _query_ arguments are required.
+The optional boolean-valued _materialized_ argument indicates if the CTE should be materialized during query execution; if unspecified, it is left to the backing database to decide.
+
 ## literal
 
 `literal(value)`
--- a/docs/api/sql/queries.md
+++ b/docs/api/sql/queries.md
@@ -101,8 +101,8 @@ The _tables_ may be table name strings, queries or subquery expressions, and map
 
 `Query.with(...expressions)`
 
-Provide a set of named subqueries in the form of [common table expressions](https://duckdb.org/docs/sql/query_syntax/with.html) and return this query instance.
-The input _expressions_ should consist of one or more maps (as JavaScript `object` values) from subquery names to query expressions.
+Provide a set of named subqueries in the form of [common table expressions (CTEs)](https://duckdb.org/docs/sql/query_syntax/with.html) and return this query instance.
+The input _expressions_ should consist of one or more maps (as JavaScript `object` values) from subquery names to query expressions and/or CTE instances produced by the `cte` method.
 
 ## distinct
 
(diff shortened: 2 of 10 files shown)


domoritz (approved):
Implementation looks good. 

Just checking my understanding. Materialization will be good when the CTE is needed again. It's probably not a good idea when the queries using the CTE have a restrictive filter as that would mean that we materialize a large table but don't really need to. I can imagine this happening when we have an M4 query to simplify a line chart but we are only looking at a small section of it. We should make sure that the selection (with clause) is pushed into the lateralized CTE. My understanding is that a database would not push the selection down if there is a materialized CTE but would if there is no materialized CTE (in which case the CTE is fused with the query on top of it).

jheer:
> Just checking my understanding. Materialization will be good when the CTE is needed again.

My understanding is that materialization is only occurring for that single query, then discarded (ie it is not persistent). This is still useful to prevent duplicate work when the same CTE is invoked multiple times in a single query. M4 invokes the same CTE four times. In the case where the input query to M4 (which becomes the CTE) has a selective filter we should get a nice performance boost as we then only have to filter once. I checked and I’m seeing a ~2x improvement on the overview plus detail example, which includes quite selective filtering.

So for M4 enforcing materialization is a win. I can’t speak to other scenarios where filtering happens *outside* the CTE, but we’re not enforcing anything in such other cases.

domoritz:
The case I was thinking was something like this

```sql
WITH cte AS (
    SELECT value * 2 AS value FROM table
)
SELECT *
FROM cte e
WHERE c.value > 10 AND c.value < 10.1;
```

Here the query on the CTE is very selective because the CTE is not materialized, the query can become.

```
SELECT value * 2 AS value
FROM table e
WHERE e.value * 2 > 10 AND e.value * 2 < 10.1;
```

but with a materialized CTE

```sql
WITH MATERIALIZED cte AS (
    SELECT value * 2 AS value FROM table
)
SELECT *
FROM cte e
WHERE c.value > 10 AND c.value < 10.1;
```

the query may run slower than without materialization. My moving the selection, it would be fast again

```sql
WITH MATERIALIZED cte AS (
    SELECT value * 2 AS value FROM table,
    WHERE c.value > 10 AND c.value < 10.1;
)
SELECT *
FROM cte e
```

The query here is artificial but I was wondering whether we run into a similar situation in a line chart where the visible domain is small.

PR #705, opened 2025-03-03
feat: Add cache exporting and hydration capabilities

rpechuk:
Added better typing to caches and also the ability to import/export caches to Arrow IPC stream format. Then integrated this new functionality into the `publish` package alongisde better type propogation to ensure everything is working properly.

> **IMPORTANT NOTE:** When this is merged a new release will need to be cut due to the use of the new `.import()` function in the generated code as part of publish. This could be fixed by bundling current version of external packages into published visualization directory and refrencing this instead of CDN but let's discuss this in #682.

### DIFF
--- a/packages/core/src/util/cache.js
+++ b/packages/core/src/util/cache.js
@@ -1,4 +1,4 @@
-import { list, tableFromArrays, tableFromIPC, tableToIPC,  uint8, utf8 } from "@uwdata/flechette";
+import { tableFromIPC, tableToIPC } from "@uwdata/flechette";
 
 const requestIdle = typeof requestIdleCallback !== 'undefined'
   ? requestIdleCallback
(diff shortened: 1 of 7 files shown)


packages/core/src/util/cache.js
```diff
@@ -1,16 +1,78 @@
+import { list, tableFromArrays, tableFromIPC, tableToIPC,  uint8, utf8 } from "@uwdata/flechette";
+
 const requestIdle = typeof requestIdleCallback !== 'undefined'
   ? requestIdleCallback
   : setTimeout;
 
+/**
+ * @typedef {import('@uwdata/flechette').Table | Promise<import('@uwdata/flechette').Table>} CacheEntry
+ * @typedef {Object} Cache
```
jheer: I believe this should be `object`. IIRC `Object` annoyingly has different semantics.

packages/core/src/util/cache.js
```diff
@@ -1,16 +1,78 @@
...
+      acc.key.push(key);
+      acc.value.push(bytes);
+      return acc;
+    },
+    { key: [], value: [] }
+  );
+  const table = tableFromArrays(cache_object, {
+    types: {
+      key: utf8(),
+      value: list(uint8())
```
jheer: Shouldn't this be a [`binary`](https://idl.uw.edu/flechette/api/data-types#binary) type instead?

packages/core/src/util/cache.js
```diff
@@ -1,16 +1,78 @@
...
   ? requestIdleCallback
   : setTimeout;
 
+/**
+ * @typedef {import('@uwdata/flechette').Table | Promise<import('@uwdata/flechette').Table>} CacheEntry
+ * @typedef {Object} Cache
+ * @property {(key: string) => CacheEntry | undefined} get Retrieves a value from the cache.
+ * @property {(key: string, value: CacheEntry) => any} set Stores a value in the cache and returns it.
+ * @property {() => void} clear Clears all entries in the cache.
+ * @property {() => Uint8Array | null} export Exports the cache as an array of bytes in Arrow IPC binary format or null if the cache is empty.
```
jheer: It looks like these methods export/import the entire cache in one go. I wonder if we instead want something more modular, that imports key/value pairs (string keys, Arrow IPC values) rather than a single Arrow IPC table containing all entries. Curious if you had an overriding reason for the current approach?

packages/core/src/util/cache.js
```diff
@@ -1,16 +1,78 @@
...
   set: (key, value) => value,
-  clear: () => {}
+  clear: () => {},
+  export: () => null,
+  import: () => {}
 });
 
+/**
+ * Creates a Least Recently Used (LRU) cache with a fixed size and time-to-live (TTL).
+ * @param {Object} [options] Configuration options for the cache.
```
jheer: `object`

packages/core/src/util/cache.js
jheer: One idea is to have the cache import/export explicitly work with key value pairs (e.g., `Map<string,Uint8Array>` or `[string,Uint8Array][]`). The publish module can then make its own decisions about how to bundle (or not bundle) these up for reading/writing cache content to external files. (p.s. note also that Arrow includes a [`map` type](https://idl.uw.edu/flechette/api/data-types#map) you could use if it helpfully simplifies later serialization.)

jheer (changes requested):

packages/core/src/util/cache.js
```diff
@@ -40,19 +67,37 @@ export function lruCache({
...
+      );
     },
-    set(key, value) {
-      cache.set(key, { last: performance.now(), value });
-      if (cache.size > max) requestIdle(evict);
-      return value;
-    },
-    clear() { cache = new Map; }
+    import(data) {
+      Array.from(data).forEach(
```
jheer: Here is an alternative. Note that we need to use `decodeIPC` (from the util folder) to ensure consistent Arrow parsing relative to the database connectors.

```js
for (const [key, entry] of data) {
  set(key, decodeIPC(entry.value));
}
```

packages/core/src/util/cache.js
```diff
@@ -40,19 +67,37 @@ export function lruCache({
...
+    set,
+    clear() { cache = new Map; },
+    export() {
+      return Array.from(cache).reduce(
+        (acc, [key, entry]) => {
+          acc.set(key, tableToIPC(entry.value, { format: 'stream' }));
+          return acc;
+        },
+        new Map()
+      );
```
jheer: Here is an alternative, which returns a [key, value] array instead of a `Map`. Also, we don't need to indicate the `stream` IPC format as that is the default.

```js
return Array.from(cache).map([key, entry] => [key, tableToIPC(entry.value)]);
```

packages/publish/src/MosaicPublisher.ts
```diff
@@ -240,10 +241,21 @@ export class MosaicPublisher {
...
    */
   private writeFiles(isInteractive: boolean, element?: HTMLElement | SVGElement) {
+    const cache = this.ctx.coordinator.manager.cache().export();
+    const cacheFile = '.cache.arrow';
+    if (cache) {
+      const cacheBytes = tableToIPC(tableFromArrays({ cache: [cache] }, {
+        types: {
+          cache: map(utf8(), binary())
+        }
+      }), { format: 'stream' })!;
```
jheer: Again, don't need the `'stream'` argument as that is the default.

jheer (changes requested):
Looking good, modulo some small suggested changes.

PR #716, opened 2025-03-11
Push preagg group by columns to aggregate subqueries.

jheer:
- Fix preaggregator by ensuring group by columns are correctly pushed down to subqueries that perform aggregation.

### DIFF
--- a/packages/core/src/preagg/PreAggregator.js
+++ b/packages/core/src/preagg/PreAggregator.js
@@ -1,4 +1,4 @@
-import { Query, and, asNode, ceil, collectColumns, createTable, float64, floor, isBetween, int32, mul, round, scaleTransform, sub, isSelectQuery, ExprNode, SelectQuery } from '@uwdata/mosaic-sql';
+import { Query, and, asNode, ceil, collectColumns, createTable, float64, floor, isBetween, int32, mul, round, scaleTransform, sub, isSelectQuery, ExprNode, SelectQuery, isAggregateExpression, ColumnNameRefNode } from '@uwdata/mosaic-sql';
 import { preaggColumns } from './preagg-columns.js';
 import { fnv_hash } from '../util/hash.js';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/preagg/PreAggregator.js
```diff
@@ -331,20 +331,45 @@ function preaggregateInfo(clientQuery, active, preaggCols, schema) {
...
+    // so we memoize and exit early if already seen
     if (memo.has(q)) return;
     memo.add(q);
+
     if (isSelectQuery(q) && q._from.length) {
+      // select the pushed down columns
+      // note that the select method will deduplicate for us
       q.select(cols);
+      if (isAggregateQuery(q)) {
+        // if an aggregation query, we need to push to groupby as well
```
domoritz, line 350: DuckDB has a convenient `GROUP BY ALL` but it's probably good not to rely on it so we can more easily support other databases.

packages/core/src/preagg/PreAggregator.js
```diff
@@ -331,20 +331,45 @@ function preaggregateInfo(clientQuery, active, preaggCols, schema) {
...
+        // we also deduplicate as the column may already be present
+        const set = new Set(
+          q._groupby.flatMap(x => x instanceof ColumnNameRefNode ? x.name : []
+        ));
+        q.groupby(cols.filter(c => !set.has(c)));
```
domoritz, line 355: Wouldn't it be easier to deduplicate in the method for adding clauses/column refs to the groupby?

domoritz (approved):

packages/core/src/preagg/PreAggregator.js
jheer: Agreed.

packages/core/src/preagg/PreAggregator.js
jheer: I looked into that first, but its a bit trickier to handle the general case where all kinds of different expressions get passed in. For example, one could first render everything out to strings and then dedupe based on those. Here it seemed simpler: we only need to check against raw column references as we know the inputs are limited to string column names.

PR #717, opened 2025-03-11
Safeguard calls on unregistered clients.

jheer:
- Fix request update calls to be no-ops when clients are not registered with a coordinator.


domoritz (approved):
Makes sense.

Issue #461, opened 2024-07-15
Do not update charts that are out of view


domoritz:
In large dashboards, some charts may be outside the current view and therefore don't need to be updated. It would be interesting to explore ways to not update charts that are invisible at the moment but update them when they come into view.

declann:
+1, I'm experiencing duckdb is like a rocket shipping results at any granularity but when it results in many marks and their interactor stores, the front-end gets janky relatively quickly.

I'm putting busy vgplots under Details elements so that I can programaticaly introduce them (and the corresponding front-end load), but there is no mechanism to disconnect them afterwards.

jheer:
Mosaic has support for this since https://github.com/uwdata/mosaic/pull/721.

You can use an [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to toggle the client `enabled` property in response to the visibility of a corresponding DOM element.

Here is one way to set that up in TypeScript, using the `makeClient` utility:

```ts
import { makeClient, type MosaicClient } from "@uwdata/mosaic-core";

const CLIENT = "__client__";

type ClientElement = Element & { [CLIENT]: MosaicClient };

let observer: IntersectionObserver;

if (typeof IntersectionObserver !== "undefined") {
  observer = new IntersectionObserver(entries => {
    for (const entry of entries) {
      const target = entry.target as ClientElement;
      target[CLIENT].enabled = entry.isIntersecting;
    }
  });
}

/**
 * Create a new client instance associated with an observed DOM element.
 * When the DOM element is off-screen, the client is deactivated.
 */
export function observeClient(
  options: Parameters<typeof makeClient>[0] & { element?: Element }
): ReturnType<typeof makeClient> {
  const { element, ...clientOptions } = options;

  // create the client instance
  const client = makeClient({
    enabled: !element,
    ...clientOptions
  });

  if (element) {
    // annotate DOM element with client
    Object.assign(element, { [CLIENT]: client });
    // add the DOM element to the IntersectionObserver
    observer.observe(element);
  }

  return client;
}
```

declann:
Isn't this still an issue in relation to `vgplot` ?

PR #721, opened 2025-03-16
Add client.enabled property.

jheer:
- Add client `enabled` property. When `enabled` is set to false, the client will not initialize nor respond to query requests and selection updates. This setting can improve performance, for example by suppressing updates to off-screen interface components. When `enabled` is set to true, any pending initialization or query requests will then proceed.
- Add `enabled` option to `makeClient` helper utility.
- Refactor initialization logic between coordinator and client.
- Add tests for client `enabled` property.

### DIFF
--- a/packages/core/src/Coordinator.js
+++ b/packages/core/src/Coordinator.js
@@ -1,6 +1,5 @@
 import { socketConnector } from './connectors/socket.js';
 import { PreAggregator } from './preagg/PreAggregator.js';
-import { queryFieldInfo } from './util/field-info.js';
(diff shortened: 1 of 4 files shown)


domoritz (approved):
Awesome. I suppose this will help with #461.

I am a bit concerned about the name. An active selection/clause means it's the last interacted with but an active client means it's a client that is enabled. Maybe that's okay but as an alternative, we could call it `enabled`. I don't feel strongly about it, though.

jheer:
I prefer `enabled` as well. I'll make that change, then merge!

PR #722, opened 2025-03-17
Drop client fields and fieldInfo methods, use prepare instead.

jheer:
This PR simplifies and streamlines the client lifecycle to use a single `prepare` call as part of client initialization. It includes the breaking change of removing the Mosaic client `field` and `fieldInfo` lifecycle methods. Instead of these methods, clients can directly call mosaic-core's `queryFieldInfo` within a client's `prepare` lifecycle method.

- **Breaking**: Drop client `fields` and `fieldInfo` lifecycle methods, rely solely on `prepare` instead. The `queryFieldInfo` can be used instead to lookup field info directly; it uses the same types as the removed methods.
- Update inputs and plot package to call `queryFieldInfo` within the client `prepare` method.


domoritz (approved):
I like this simplification. Makes sense!

PR #723, opened 2025-03-17
Add filterStable option to makeClient.

jheer:
- Add `filterStable` option to makeClient.


jheer:
> This is exposing an internal so that one can disable preaggregation. The name isn't the most meaningful if someone wants to disable preaggregation for performance reasons rather than because the filter isn't stable. We can think of a more generic external API later. For now, this makes sense.

That's right. This is for internal use by developers of custom client implementations. Toggling preaggregation more generally is a separate concern and can already be achieved via `coordinator.preaggregator.enabled = false`.

domoritz:
I think developers might want to toggle preaggregation per client as well if they anticipate only few queries for example. This api allows that but the name is potentially not what developers would look for.

PR #737, opened 2025-04-02
fix: MosaicClient types

kwonoh:


kwonoh:
The test randomly fails
- https://github.com/uwdata/mosaic/actions/runs/14228988816/job/39875235423#step:7:77
- https://github.com/uwdata/mosaic/actions/runs/14230224439/job/39879116551#step:7:76

domoritz:
They seems to be something with the order of tests. Somewhere we need to first install the extension. 

I've seen this before so it's probably safe to ignore for now.

domoritz (approved):
Looks good. Thanks for the pull request.

PR #744, opened 2025-04-11
Add classes/types for database connectors.

jheer:
- Add database connector types.
- Add class-based implementations of socket and wasm connectors.
- Add `destroy` method to client, transferred from `make-client` helper.

### DIFF
--- a/packages/core/src/connectors/Connector.ts
+++ b/packages/core/src/connectors/Connector.ts
@@ -25,6 +25,6 @@ export interface JSONQueryRequest extends QueryRequest {
 export interface Connector {
   /** Issue a query and return the result. */
   query(query: ArrowQueryRequest): Promise<Table>;
-  query(query: ExecQueryRequest): Promise<undefined>;
+  query(query: ExecQueryRequest): Promise<void>;
   query(query: JSONQueryRequest): Promise<Record<string, any>[]>;
 }
--- a/packages/core/src/connectors/Connector.ts
+++ b/packages/core/src/connectors/Connector.ts
@@ -1,0 +1,30 @@
+import type { Table } from '@uwdata/flechette';
+
+export interface QueryRequest {
+  /** The query type. */
+  type?: string;
+  /** A SQL query string. */
+  sql: string;
+}
+
+export interface ArrowQueryRequest extends QueryRequest {
+  /** The query type. */
+  type?: 'arrow';
+}
+
+export interface ExecQueryRequest extends QueryRequest {
+  /** The query type. */
+  type: 'exec';
+}
+
+export interface JSONQueryRequest extends QueryRequest {
+  /** The query type. */
+  type: 'json';
+}
+
+export interface Connector {
+  /** Issue a query and return the result. */
+  query(query: ArrowQueryRequest): Promise<Table>;
+  query(query: ExecQueryRequest): Promise<undefined>;
+  query(query: JSONQueryRequest): Promise<Record<string, any>[]>;
+}
--- a/packages/core/src/connectors/socket.js
+++ b/packages/core/src/connectors/socket.js
@@ -1,100 +1,108 @@
+/** @import { Connector } from './Connector.js' */
 import { decodeIPC } from '../util/decode-ipc.js';
 
+/**
+ * Connect to a DuckDB server over a WebSocket interface.
+ * @param {string} uri The URI for the DuckDB socket server.
+ * @returns {SocketConnector} A connector instance.
+ */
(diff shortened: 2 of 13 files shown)


packages/core/src/connectors/socket.js
```diff
@@ -1,100 +1,108 @@
...
-     * @param {string} [query.name] The name of a bundle to create or load.
-     * @returns the query result
-     */
-    query(query) {
-      return new Promise(
-        (resolve, reject) => enqueue(query, resolve, reject)
-      );
-    }
-  };
+  query(query) {
```
rpechuk, line 120: This won't propogate the typing from the interface and will show up as `query: any => Promise<any>`. Should we keep the JSDoc comment for typing here? Alternatively and probably better would be to include the type comment `/** @type {import('./connectors/Connector.js').Connector} */` above `this.db` in `QueryManager.js` (and any other relavant spots) to ensure the typing is correct there as this is where the connectors are actually being used.

packages/core/src/connectors/Connector.ts
```diff
@@ -0,0 +1,30 @@
...
+  /** The query type. */
+  type: 'exec';
+}
+
+export interface JSONQueryRequest extends QueryRequest {
+  /** The query type. */
+  type: 'json';
+}
+
+export interface Connector {
```
rpechuk, line 25: Should we use `QueryResult` here and/or move that into this file for better typing/organization.

rpechuk (approved):
Overall looks good to me! Just left two comments with things to consider in terms of further improving the typing associated with connectors.

packages/core/src/connectors/Connector.ts
```diff
@@ -0,0 +1,30 @@
...
+  /** Issue a query and return the result. */
+  query(query: ArrowQueryRequest): Promise<Table>;
+  query(query: ExecQueryRequest): Promise<undefined>;
```
domoritz: Should this be undefined or void?

packages/core/src/connectors/socket.js
domoritz: This class implements the connector but the types are not showing up here. It works if we define the return type in line 7 as `Connector`.

domoritz: Maybe we can figure out how to move to actual ts soon to make it easier to deal with types and inheritance.

domoritz (approved):

packages/core/src/connectors/socket.js
rpechuk: > Maybe we can figure out how to move to actual ts soon to make it easier to deal with types and inheritance.

Agreed! A good first step could be to convert all the `.js` files to `.ts` and enable strict type checking. That should surface a bunch of type errors we can work through incrementally, and hopefully get us to full TypeScript coverage (In the meantime all new files we create we should do in TS to avoid any weird conflicts).

domoritz: First we need to find out how we can get development builds against the ts files when we compile them for deployment builds.

packages/core/src/util/cache.js
```diff
@@ -1,13 +1,23 @@
...
+ */
 export const voidCache = () => ({
   get: () => undefined,
   set: (key, value) => value,
   clear: () => {}
 });
 
+/**
+ * Create a new cache that uses an LRU eviction policy.
+ * @type {(options?: {max?: number, ttl?: number}) => Cache}
```
domoritz: I think I prefer types for the arguments and return type instead.

packages/core/src/util/cache.js
```diff
@@ -1,13 +1,23 @@
...
+
 const requestIdle = typeof requestIdleCallback !== 'undefined'
   ? requestIdleCallback
   : setTimeout;
 
+/**
+ * Create a new cache that ignores all values.
+ * @type {() => Cache}
+ */
 export const voidCache = () => ({
```
domoritz: Nit: could be a function instead

packages/core/src/types.ts
```diff
@@ -62,3 +62,23 @@ export interface Activatable {
    */
   activate(): void;
 }
+
+/**
+ * Interface for cache implmenetions.
```
domoritz:
```suggestion
 * Interface for cache implementations.
```

domoritz (approved):

packages/core/src/connectors/Connector.ts
jheer: I don't think so. `QueryResult` is an abstraction provided by the query manager. The database connectors are lower-level.

PR #747, opened 2025-04-19
Mosaic Widget: avoid `params` mutation - fixes #746

### LINKED ISSUE #746: Mosaic Widget `params` reactivity failing due to traitlet mutation
Hi,

In a marimo notebook I tried to make something react off `params` from a Mosaic widget - it didn't work, and seems to be caused by how params is set before being updated into the model.

Here in the listener: https://github.com/uwdata/mosaic/blob/1c29646087aa630a84d66df9d29b1d21c09d2160/packages/widget/src/index.js#L78

I replaced Js in my browser to:

~~~js
  const newParams = {
    ...params,
    [name]: {
      value,
      ...(isSelection(param) ? { predicate: String(param.predicate()) } : {}),
    },
  };
  view.model.set('params', newParams);
  view.model.save_changes();
~~~

This fixes my issue

declann:
Permits Mosaic Widget `params` reactivity in marimo (don't know if issue affected Jupyter)

### DIFF
--- a/packages/widget/src/index.js
+++ b/packages/widget/src/index.js
@@ -76,7 +76,7 @@ export default {
 
         param.addEventListener('value', (value) => {
           view.model.set('params', {
-            ...params,
+            ...view.model.get('params'),
             [name]: {
               value,
               ...(isSelection(param) ? { predicate: String(param.predicate()) } : {})
--- a/packages/widget/src/index.js
+++ b/packages/widget/src/index.js
@@ -75,11 +75,13 @@ export default {
         };
 
         param.addEventListener('value', (value) => {
-          params[name] = {
-            value,
-            ...(isSelection(param) ? { predicate: String(param.predicate()) } : {}),
-          };
-          view.model.set('params', params);
+          view.model.set('params', {
+            ...params,
+            [name]: {
+              value,
+              ...(isSelection(param) ? { predicate: String(param.predicate()) } : {})
+            },
+          });
           view.model.save_changes();
         });
       }
--- a/packages/widget/src/index.js
+++ b/packages/widget/src/index.js
@@ -66,10 +66,10 @@ export default {
       view.el.replaceChildren(dom.element);
 
       /** @type Params */
-      const params = {};
+      const initialParams = {};
 
       for (const [name, param] of dom.params) {
-        params[name] = {
+        initialParams[name] = {
           value: param.value,
           ...(isSelection(param) ? { predicate: String(param.predicate()) } : {}),
         };
@@ -86,7 +86,7 @@ export default {
         });
       }
 
-      view.model.set('params', params);
+      view.model.set('params', initialParams);
       view.model.save_changes();
     }
 


packages/widget/src/index.js
```diff
@@ -75,11 +75,13 @@ export default {
         };
 
         param.addEventListener('value', (value) => {
-          params[name] = {
-            value,
-            ...(isSelection(param) ? { predicate: String(param.predicate()) } : {}),
-          };
-          view.model.set('params', params);
+          view.model.set('params', {
+            ...params,
```
domoritz: Hmm, how is this correct? This doesn't update the `params` variable so doesn't that mean we would lose any changes to params that are not `name`?

domoritz: Maybe this would be correct if we instead got the params from `view.model.get('params')` before setting them again.

declann: Sorry I far too shallow tested this.
When I get from `view.model.get('params')` it works but there are still issues: I only see `range` and `click` entries and `click` has strange defaults
[Screencast from 2025-04-19 19-51-06.webm](https://github.com/user-attachments/assets/ca25bc7b-256d-46e9-860a-750c29106fbd)

declann: Whereas without my changes there are more entries for `domain` and `colors` - I presume non-selection params that should be maintained:
image

declann: Above I made a different error in the initial `model.set`.
Starting from `view.model.get('params')` as suggested does work fine in this example.
[Screencast from 2025-04-19 20-04-17.webm](https://github.com/user-attachments/assets/20df5593-c26e-4e5f-a301-abde2af31f81)

domoritz:
It feels a bit odd to create the params variable and then not keeping it up to date. We should probably rename it to initialParams.

I also don't full get yet why the original code broke. Can you explain it again?

declann:
> It feels a bit odd to create the params variable and then not keeping it up to date. We should probably rename it to initialParams.

:+1: Done this is clearer

> I also don't full get yet why the original code broke. Can you explain it again?

I can't off hand confidently explain all the details down to the wire (mutation and pass by reference for Arrays and Objects in JS, and comparison in anywidget signals), but it corresponds to an issue I encountered in other reactive systems like Observable.

My understanding in this case is that the code was mutating elements in an array, but it's the same array from the pov of the reactive system.

For the reactive system to propagate the changes, you should set to a new array.

Tom MacWright wrote a good piece on mutation relating to Observable JS here: https://observablehq.com/@tmcw/observable-anti-patterns-and-code-smells#mutation

> these cells interact with each other in a way that the notebook can’t track: running the second one doesn’t update the results in the third, because we don’t statically track mutation.
> 
> Luckily, there are simple ways to avoid this: following, again, the rule that if you want to get a modified object, it should be a new object, not a modification of the one you have. [Spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), noted as ... in code, makes this great

(Relating to Objects)

declann:
Not sure how effective it is but here's a notebook without/with a copy, and simple JS comparison results: https://observablehq.com/d/5b2e30baf59e289c

I believe that redux, react, Observable, presumably anywidget all boil down to these simple comparisons because alternatives are quite expensive. Therefore modifying things in place that looks fine from a pure JS pov, becomes a "code smell", and there are libraries to guarantee immutability like [this one](https://immutable-js.com/) - note I'm not an expert at all here.

domoritz:
Ahh, that makes sense. It's think https://github.com/uwdata/mosaic/pull/748 might be a cleaner fix.

domoritz:
I added freeze in the other pull request. Let's use that.

declann:
Thanks Dominik

domoritz:
Tanks for digging into this. Does the other pull request look good to you?

declann:
Yes it's good and fixes my testcase in marimo - thanks!

Also `specs.ipynb` notebook which sets the `spec` traitlet to change specs still works.

I suppose that maintaining the `params` object in this JS is simpler than getting from the model in the listener and I don't see any problem with it.

domoritz:
I don't know what `get` does. Maintaining the params object is definitely fast and easier to grasp.

PR #748, opened 2025-04-20
fix: avoid param mutation

### LINKED ISSUE #746: Mosaic Widget `params` reactivity failing due to traitlet mutation

domoritz:
fixes #746

New version of https://github.com/uwdata/mosaic/pull/747. Thanks @declann. 

Uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze to avoid mutation of params object.

### DIFF
--- a/packages/widget/pyproject.toml
+++ b/packages/widget/pyproject.toml
@@ -28,7 +28,8 @@ dev = [
   "jupyterlab",
   "pandas",
   "pyyaml",
-  "ruff"
+  "ruff",
+  "watchfiles"
 ]
 
 [tool.hatch.version]
--- a/packages/widget/src/index.js
+++ b/packages/widget/src/index.js
@@ -75,18 +75,19 @@ export default {
         };
 
         param.addEventListener('value', (value) => {
-          params = {
+          params = Object.freeze({
             ...params,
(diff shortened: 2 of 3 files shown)


packages/widget/pyproject.toml
```diff
@@ -28,7 +28,8 @@ dev = [
   "jupyterlab",
   "pandas",
   "pyyaml",
-  "ruff"
+  "ruff",
+  "watchfiles"
```
domoritz, line 32: Added to allow for HMR in widget

PR #749, opened 2025-04-21
feat: add ENUM to jsType

kwonoh:


packages/core/src/util/js-type.js
```diff
@@ -41,7 +41,9 @@ export function jsType(type) {
     case 'GEOMETRY':
       return 'object';
     default:
-      if (type.startsWith('DECIMAL')) {
+      if (type.startsWith("ENUM")) {
```
domoritz: Why starts with instead of just matching? Is that because the values are part of the type? https://duckdb.org/docs/stable/sql/data_types/enum

kwonoh: Yes, the value of `type` variable in this function for `ENUM` types looks like `ENUM('value1', 'value2', 'value3')`

packages/core/src/util/js-type.js
```diff
@@ -41,7 +41,9 @@ export function jsType(type) {
...
+        return "string";
```
jheer: Please use consistent formatting (e.g., single quoted strings). Otherwise, this looks great.

jheer (changes requested):

packages/core/src/util/js-type.js
kwonoh: Updated. I think the quote style rule should be added in the eslint config file or something similar.

domoritz: Yes. Let's add it. Not sure why it's not in the recommended rules.

jheer:
Thanks @kwonoh!

PR #750, opened 2025-04-22
feat: changed describe query to support clickhouse

jerryc214:
Previously describe query used `DESCRIBE`, which was only supported by DuckDB. Changed query to `DESC()` which is supported by both DuckDB and Clickhouse.

### DIFF
--- a/packages/sql/src/ast/query.js
+++ b/packages/sql/src/ast/query.js
@@ -515,7 +515,7 @@ export class DescribeQuery extends SQLNode {
    * @returns {string}
    */
   toString() {
-    return `DESCRIBE ${this.query}`;
+    return `DESC(${this.query})`;
   }
 }
 
--- a/packages/sql/test/query.test.js
+++ b/packages/sql/test/query.test.js
@@ -500,12 +500,12 @@ describe('Query', () => {
 
   it('supports describe queries', () => {
     const q = Query.select('foo', 'bar').from('data');
-    expect(Query.describe(q).toString()).toBe(`DESCRIBE ${q}`);
+    expect(Query.describe(q).toString()).toBe(`DESC(${q})`);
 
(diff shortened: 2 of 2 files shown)


domoritz (commented):
The tests need to be updated.

It looks like click house supports describe, though: https://clickhouse.com/docs/sql-reference/statements/describe-table.

jheer (changes requested):
The change looks good, providing a mechanism that will work with both DuckDB and ClickHouse.

However, the test cases need to be updated to match the changes to the code. Please update `packages/sql/test/query.test.js` (e.g., around line 501) accordingly.

domoritz (commented):
What about my question?

jerryc214:
Sorry about not replying earlier. ClickHouse supports `DESCRIBE` so far as describing entire tables. However, the `DESCRIBE` referenced here is the one used for describing _queries_ to collect the schema of a query result - something that ClickHouse uses a different syntax for. Instead, we found that `DESC(query)` is the syntax for describing query results in ClickHouse and still works the same in DuckDB.

domoritz (approved):
That is weird. Is this documented somewhere? Either way, then this change makes sense.

domoritz:
Actually, looks like there are a few more instances that need to be changed: https://github.com/search?q=repo%3Auwdata%2Fmosaic%20DESCRIBE&type=code

derekperkins:
https://github.com/uwdata/mosaic/blob/2e0d35f15c9ee5a5d562bed4aa340487d54f610a/packages/mosaic/sql/src/visit/codegen/duckdb.ts#L109-L112

We should be able to close this, I saw that `DESC` was changed in #886. For additional incompatibilities, there's now a route to add native ClickHouse support by adding a new visitor

PR #772, opened 2025-05-15
feat: Create `ClauseSource` interface and add book keeping to `Coordinator` for clause sources

rpechuk:


packages/core/src/Coordinator.js
```diff
@@ -244,6 +246,17 @@ export class Coordinator {
 
     // connect filter selection
     connectSelection(this, client.filterBy, client);
+
+    // Bookkeep all clause sources
+    if (isClauseSource(client)) {
```
jheer, line 268: We need a more general solution. The technique of inspecting the client based on known types (such as Plot) is breaking our intended abstraction barriers. Clients and clause sources are often paired, but in general can be distinct. I think we need to find a way to have interactors register/deregister themselves; this will require that they have access to the corresponding coordinator instance, which may involve refactoring parts of the other packages.

packages/inputs/src/Table.js
```diff
@@ -275,6 +275,13 @@ export class Table extends Input {
     }
   }
 
+  reset() {
+    this.currentRow = -1;
+    if (isSelection(this.selection)) {
+      this.selection.reset();
```
jheer: The semantics of `reset` should be clarified. I think of the resetting the clause source as separate from a reset of a selection. The reset of the clause source should reset any interface state (like the brush extents, or text query string) but should not affect the selection state. The selection resolver instead is responsible for the selection state, and calls reset on the clause sources.

packages/plot/src/interactors/Nearest.js
```diff
@@ -83,6 +83,11 @@ export class Nearest {
     const v = this.channels.map(() => 0);
     this.selection.activate(this.clause(v));
   }
+
+  reset() {
+    this.valueIndex = -1;
+    this.selection.reset();
```
jheer: See earlier comment about selection vs. clause source reset.

packages/plot/src/interactors/PanZoom.js
```diff
@@ -105,6 +105,11 @@ export class PanZoom {
       this.ysel.activate(this.clause(yscale.domain, yfield, yscale));
     }
   }
+
+  reset() {
+    if (this.panx) this.xsel.reset();
+    if (this.pany) this.ysel.reset();
```
jheer: See previous comments.

packages/publish/README.md
```diff
@@ -32,10 +32,11 @@ The functionality of mosaic-publish can also be accessed programmatically via it
 
 ```js
 import { MosaicPublisher } from '@uwdata/publish';
+import fs from 'fs';
 
 const publisher = new MosaicPublisher({
-  spec: './path/to/spec.json',
-  outputPath: './docs/public/schema'
+  spec: fs.readFileSync('./path/to/spec.yaml', 'utf8'),
```
jheer: This seems unrelated to the current PR?

jheer (changes requested):
I think we still need to revisit both how clause sources are identified, and what the semantics of clause source reset are. See more specific comments for more.

packages/publish/README.md
rpechuk: This was meant to be on the base `publish` pr. Should be moved now.

packages/inputs/src/Table.js
rpechuk: Should be fixed now! I'm a bit unsure if what i'm doing for the `PanZoom.js` reset function is correct but from what I understand everything should be good in its current state!

packages/core/src/Coordinator.js
rpechuk: Should be resolved now, let me know what you think of the new way I'm doing it.

packages/core/src/Coordinator.js
```diff
@@ -244,6 +265,9 @@ export class Coordinator {
 
     // connect filter selection
     connectSelection(this, client.filterBy, client);
+
+    // Connect this client as a clause source if necessary
+    this.connectClauseSource(client);
```
jheer: Use a typeguard if statement here instead of within the `connectClauseSource` method?

packages/inputs/src/Table.js
```diff
@@ -275,6 +275,17 @@ export class Table extends Input {
     }
   }
 
+  reset() {
+    this.currentRow = -1;
+    
+    if (this.sortHeader) {
```
jheer: Reset should only reset interface state related to any (now removed) selection clauses.

packages/plot/src/interactors/PanZoom.js
```diff
@@ -105,6 +110,31 @@ export class PanZoom {
       this.ysel.activate(this.clause(yscale.domain, yfield, yscale));
     }
   }
+
+  reset() {
```
jheer: Make this a no-op

jheer (changes requested):
Almost there. Just some smaller changes left to make!

jheer (changes requested):
Looking better! I made a few minor changes to simplify code. However, this also led me to one question and one remaining larger issue.

Question: the interactors are typically registering with the coordinator under the `init` method. Why here rather than the constructor? The `init` method currently gets invoked every time the chart gets a full re-render. It's safe to do that (as the Coordinator uses a Set to track clause sources), but maybe its cleaner to do it once?

Issue: a number of interactors don't (yet) follow the newly established rules for a clause source, namely that the `clause()` method must return a meaningful result when called without arguments. We still need to update some internals to create the right default values when arguments are not provided (IIRC, pan/zoom, toggle, region... maybe others).

rpechuk:
> Question: the interactors are typically registering with the coordinator under the `init` method. Why here rather than the constructor? The `init` method currently gets invoked every time the chart gets a full re-render. It's safe to do that (as the Coordinator uses a Set to track clause sources), but maybe its cleaner to do it once?

Fixed! Originally was under the impression `mark` wasn't set until `init` however this isn't the case.

> Issue: a number of interactors don't (yet) follow the newly established rules for a clause source, namely that the `clause()` method must return a meaningful result when called without arguments. We still need to update some internals to create the right default values when arguments are not provided (IIRC, pan/zoom, toggle, region... maybe others).

This should now be fixed. Per our in person conversation the way I handled this was by looking at "default values" passed during `activate` and ensured that the no parameter version of `clause` would work the same way (changing `activate accordingly`).

packages/plot/src/interactors/PanZoom.js
```diff
@@ -52,10 +55,15 @@ export class PanZoom {
   }
 
   clause(value, field, scale) {
-    return clauseInterval(field, value, {
+    const { xscale, xfield, yscale, yfield } = this;
```
jheer: This seems a little dicey, but I don't have a better alternative at the moment. The core issue is that panzoom manages two selections, so it's really two clause sources at once. Yet here we are treating it as a single source.

In this one case, for now perhaps its simpler to simply throw an error if called with no arguments. This "violates" the contract but in a principled way that others could handle (e.g., try/catch around use of `clause`).

Another more complicated (but more correct) route would be to actually register up to two different clause sources with the coordinator to cover both x and y selections.

rpechuk: Definitely agree on the last point, I think the best way would be to register the x/y selections separately using two clause sources. With that said I think right now throwing an error is the best option since the way that it is being done right now is definetly suspicious and could lead to weird behavior. Will change this now and merge into the publisher branch!

PR #792, opened 2025-06-14
feat: use typescript for the core package

### LINKED ISSUE #312: Add TypeScript Support

domoritz:
Fixes #312 

Update the vite config to use the source files so we don't need to run tsc manually. I migrated the code in core and mostly kept things as is.


packages/core/src/Selection.ts
```diff
@@ -28,7 +28,7 @@ export function isSelection(x: any): x is Selection {
   return x instanceof Selection;
 }
 
-function create(options: SelectionResolverOptions, include?: Selection | Selection[]): Selection {
+function create(options: SelectionResolverOptions, include?: Selection[]): Selection {
```
jheer: Why the change in types here? The code was designed to accept either a single selection or an array. And if we make the change, the code below could be adjusted to avoid the array wrap and `flat` call.

domoritz: Ah, I misinterpreted the constructor and logic here. I changed this to `include ? [include].flat() : undefined` so that the code typechecks correctly.

packages/core/src/index.ts
```diff
@@ -20,9 +20,16 @@ export {
...
 export { queryFieldInfo } from './util/field-info.js';
 export { jsType } from './util/js-type.js';
 export { isActivatable } from './util/is-activatable.js';
+export type { QueryResult } from './util/query-result.js';
+
+export * from './types.js';
+export * from './connectors/Connector.js';
+
+export type * from './Selection.js';
+export type {SelectionClause} from './SelectionClause.js';
```
kwonoh: Other types need to be exported as well (e.g., ClauseMetadata, ScaleType, Extent)
`export type * from './SelectionClause.js'`

domoritz:
@kwonoh I know you have experience with TypeScript so I'd appreciate your review.

tsconfig.base.json
```diff
@@ -1,13 +1,19 @@
...
     "sourceMap": true,
     "strict": true,
     "moduleResolution": "node16",
     "allowJs": true,
+    "checkJs": true,
     "forceConsistentCasingInFileNames": true,
 
+    // disable some checks for now until we have stricter types
+    "noImplicitAny": false,
+    "strictNullChecks": false,
```
kwonoh: I understand we're still in the early phase of the transition, but just a quick note that `strictNullChecks` is a very helpful feature. We should enable it once the typings are more robust.

packages/core/src/preagg/PreAggregator.ts
```diff
@@ -194,9 +209,10 @@ export class PreAggregator {
       info = preaggregateInfo(client.query(filter), active, preaggCols, schema);
       info.result = mc.exec([
         `CREATE SCHEMA IF NOT EXISTS ${schema}`,
+        // @ts-expect-error make type stricter
```
kwonoh: We don't get a type error here if `"strictNullChecks": true`

packages/core/src/QueryConsolidator.ts
```diff
@@ -240,29 +268,27 @@ async function processResults(group, cache) {
...
- * @param {[string, string][]} map Column name map as [source, target] pairs
+ * @param data Consolidated query result, as an Arrow Table
+ * @param map Column name map as [source, target] pairs
  * @returns the filtered table data
  */
-function filterResult(data, map) {
+function filterResult(data: Table, map: Array<[string, string]>): any[] {
   const lookup = new Map(map);
-  const result = [];
+  const result: any[] = [];
```
kwonoh, line 287: I'm encountering the below error when building one of my projects that has `"strictNullChecks": true` using this branch:
```
../../../../repo/mosaic/packages/core/src/QueryConsolidator.ts:288:19 - error TS2763: Cannot iterate value because the 'next' method of its iterator expects type 'null', but for-of will always send 'undefined'.

288   for (const d of data) {
                      ~~~~
```

It seems to originate from this line in the flechette repo:
https://github.com/uwdata/flechette/blob/158477706cb8e0268a081b190989e706f9f6437d/src/table.js#L158

I will open a PR to the flechette repo after I have better understanding of the `Generator` type

cc @jheer

packages/core/src/index.ts
domoritz: Done

packages/core/src/preagg/PreAggregator.ts
domoritz: Fixed

tsconfig.base.json
domoritz: I agree. I want to do this incrementally. I enabled stictNullCheck for the core package. I thought `strict:true` would enable it but it seems like the ts options are not as smart and only allow overriding specific properties at a time.

I enabled strict type checks for core and disabled it at the top level for now. Once this is merged, it would be awesome if you could help move package by package over and enable `strict: true` over time.

domoritz:
@kwonoh I addressed your comments and also made the types for core strict. We can move the rest incrementally but this pull request should set things up.

PR #798, opened 2025-06-19
Port sql package to typescript.

jheer:
- Port `mosaic-sql` package to TypeScript.


packages/core/package.json
```diff
@@ -26,10 +26,11 @@
     "url": "https://github.com/uwdata/mosaic.git"
   },
   "scripts": {
-    "prebuild": "rimraf dist && mkdir dist",
+    "clean": "rimraf dist && mkdir dist",
     "lint": "eslint src test",
-    "test": "vitest run && tsc",
-    "prepublishOnly": "npm run test && npm run lint && tsc"
+    "test": "vitest run && npm run tsc",
+    "tsc": "tsc",
```
domoritz, line 32: What's this for? Running `tsc --build` at the top level should be preferred usually. You can always run `npx tsc` if you need to in the local folder. 

`lerna run tsc` is super slow at the top level.

packages/sql/src/ast/select.ts
```diff
@@ -1,34 +1,28 @@
-/** @import { ExprNode } from './node.js' */
+import type { ExprNode } from './node.js';
 import { SELECT_CLAUSE } from '../constants.js';
 import { quoteIdentifier } from '../util/string.js';
 import { ColumnRefNode } from './column-ref.js';
 import { SQLNode } from './node.js';
 
 export class SelectClauseNode extends SQLNode {
+  /** The select expression. */
+  readonly expr?: ExprNode | null;
```
domoritz: What's the difference between null and undefined here?

packages/sql/src/transforms/scales.ts
```diff
@@ -93,16 +133,16 @@ const scales = {
   utc: scaleTime
 };
 
-export function scaleTransform(options) {
+export function scaleTransform<T>(options: ScaleOptions): ScaleDescriptor<T> | null {
```
domoritz: Can this method really return null? At least from the types it seems like we always have a scale assuming that options.type is correctly typed. I found typescript works best when types are kept tight.

domoritz (approved):
Fixed a typo and a few types and left some questions. Otherwise, awesome!

packages/core/package.json
jheer: The `tsc` call under `test` was added earlier. It does not appear to be needed (the local `test` script is not called when running tests at the full monorepo level), so I will remove these.

packages/sql/src/ast/select.ts
jheer: The `Query` class supports inclusion of null or undefined clauses as a way of removing a previously entered clauses with the same alias. However, your comment made me realize that this should be localized to the Query implementation and should not propagate to this AST node class. I will tighten up the types to disallow both null and undefined here.

packages/sql/src/transforms/scales.ts
jheer: Updated code to throw an error if an invalid scale type is provided. Also refactored the core package to use the scale types exported by the sql package.

PR #801, opened 2025-06-21
chore: improve build and test scripts

domoritz:
Tests and the docs now use aliases to run Vite so they will pick up the source files---no need to compile/run tsc after making changes. This should also make it easy to use watch tasks without having to worry about picking up the latest changes.

I tested running `npx run clean` at the top level and then compiling the individual packages/running tests etc. 

As a note, `tsc` just runs typescript in the current package, `tsc --build` also compiles dependencies (if they need to be updated).

You can run `npx tsc --build` or `npx tsc` (if other dependencies are already compiled) to type check any package.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -7,11 +7,7 @@ export default defineConfig({
   description: 'Scalable, interactive data visualization',
   base: '/mosaic/',
   vite: {
-    resolve: {
-      alias: Object.fromEntries(
-        Object.entries(viteConfig.resolve.alias).map(([key, value]) => [key, `.${value}`])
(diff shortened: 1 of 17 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,16 @@
+{
```
jheer: Does this need to be part of the PR? Just checking if this is intentional or if "leaked" in.

jheer (approved):

.claude/settings.local.json
domoritz: Nope. Will remove.

PR #809, opened 2025-06-25
Fix typescript lint errors in core package, adjust other packages.

jheer:
- Fix typescript lint errors in core package.
- Loosen sql `FilterExpr` type to allow boolean values.
- Adjust vgplot packages to use `ClauseSource` type.

### DIFF
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -8,6 +8,13 @@ import { type Logger, type QueryType } from './types.js';
 import { type QueryResult } from './util/query-result.js';
 import { type MosaicClient } from './MosaicClient.js';
 import { type SelectionClause } from './SelectionClause.js';
+import { MaybeArray } from '@uwdata/mosaic-sql';
+
+interface FilterGroupEntry {
+  selection: Selection;
+  clients: Set<MosaicClient>;
+  disconnect(): void;
+}
 
 /**
  * The singleton Coordinator instance.
@@ -38,9 +45,9 @@ export function coordinator(
 export class Coordinator {
   public manager: QueryManager;
   public preaggregator: PreAggregator;
-  public clients?: Set<MosaicClient>;
-  public filterGroups?: Map<Selection, any>;
-  protected _logger?: Logger;
+  public clients: Set<MosaicClient> = new Set;
+  public filterGroups: Map<Selection, FilterGroupEntry> = new Map;
+  protected _logger: Logger = voidLogger();
 
   /**
    * @param db Database connector. Defaults to a web socket connection.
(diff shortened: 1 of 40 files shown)


packages/mosaic/core/src/Coordinator.ts
```diff
@@ -38,9 +45,9 @@ export function coordinator(
 export class Coordinator {
   public manager: QueryManager;
   public preaggregator: PreAggregator;
-  public clients?: Set<MosaicClient>;
-  public filterGroups?: Map<Selection, any>;
-  protected _logger?: Logger;
+  public clients: Set<MosaicClient> = new Set;
```
domoritz: I often write it like this which I find easier.

```suggestion
  public clients = new Set<MosaicClient>;
```

packages/mosaic/core/src/MosaicClient.ts
```diff
@@ -23,7 +29,7 @@ export class MosaicClient {
   _filterBy: Selection | undefined;
   _requestUpdate: (event?: unknown) => void;
   _coordinator: Coordinator | null;
-  _pending: Promise<any>;
+  _pending: Promise<unknown>;
```
domoritz, line 32: unknown means that a user won't be able to really know anything about the type. Maybe we want to make this a generic?

packages/mosaic/core/src/connectors/rest.ts
```diff
@@ -1,41 +1,56 @@
...
-      }
-
-      return query.type === 'exec' ? req
-        : query.type === 'arrow' ? decodeIPC(await res.arrayBuffer(), ipc)
-        : res.json();
+export function restConnector(options: RestOptions = {}) {
+  return new RestConnector(options.uri, options.ipc);
+}
+
+class RestConnector implements Connector {
```
domoritz: Nice.

packages/mosaic/core/src/connectors/rest.ts
```diff
@@ -1,41 +1,56 @@
...
-      const res = await req;
-
-      if (!res.ok) {
-        throw new Error(`Query failed with HTTP status ${res.status}: ${await res.text()}`);
-      }
-
-      return query.type === 'exec' ? req
-        : query.type === 'arrow' ? decodeIPC(await res.arrayBuffer(), ipc)
-        : res.json();
+export function restConnector(options: RestOptions = {}) {
```
domoritz: should we remove the method initializers?

domoritz (approved):

packages/mosaic/core/src/MosaicClient.ts
jheer: Actually, this should always be `Promise<void>`. Fixed!

packages/mosaic/core/src/connectors/rest.ts
jheer: Removing them will break a bunch of downstream code. So let's deprecate and remove in the future.

packages/mosaic/core/src/MosaicClient.ts
jheer: No scratch that. It needs to be able to pass through additional promises. I think `unknown` is a good type here, as  the values may be of varied types and we don't care what the actual value is.

PR #814, opened 2025-06-28
feat: add duckdb-server-go implementation

### LINKED ISSUE #787: Interest in duckdb-server-go
Would you be willing to accept a PR with a Go implementation of the server, as an example similar to duckdb-server-rust? The rust and python servers appear to serve identical apis, so I would build this to match the same spec.

derekperkins:
Fully functional and tested Go implementation, should be at feature parity with Python and Rust. I have an open issue with the [go-duckdb](https://github.com/marcboeker/go-duckdb/issues/479) about a crash due to race conditions that can be triggered, especially dragging the selection in the athletes example in the default UI, but I'm fairly confident that is a library problem.

fixes #787


packages/server/duckdb-server-go/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Command-Line Options
+
+You can customize the server behavior with the following command-line flags:
+
+-   `--database <path>`: Path to a DuckDB database file (e.g., "database.db"). Defaults to an in-memory database.
+-   `--address <address>`: The HTTP address to listen on. Defaults to "localhost".
+-   `--port <port>`: The HTTP port to listen on. Defaults to "3000".
+-   `--connection-pool-size <size>`: The maximum size of the connection pool. Defaults to 10.
```
domoritz, line 28: How much do pools matter for DuckDB since queries run serially anyway, no? Or do they help with parallelizing some parts of requests?

packages/server/duckdb-server-go/go.mod
```diff
@@ -0,0 +1,40 @@
...
+
+require (
+	github.com/apache/arrow-go/v18 v18.3.1
+	github.com/coder/websocket v1.8.13
+	github.com/marcboeker/go-duckdb/v2 v2.3.2
+	github.com/maypok86/otter/v2 v2.0.0
+	github.com/stretchr/testify v1.10.0
+)
+
+require (
```
domoritz, line 13: Not super familiar with go development but do we need to specify the indirect dependencies?

domoritz (approved):
Looks great. Just two questions.

packages/server/duckdb-server-go/go.mod
derekperkins: Yes, this is a very standard, cleaned (`go mod tidy`) file. I tried to keep the imports as lean as possible.

packages/server/duckdb-server-go/README.md
derekperkins: I don't think there's any reason that queries need to run serially, though something about it is triggering the crash I mentioned in the PR description (https://github.com/marcboeker/go-duckdb/issues/479)
https://duckdb.org/docs/stable/connect/concurrency.html#concurrency-within-a-single-process

As for a default of 10, that was just a round number. I think the right number will be dependent on number of CPUs and average query latency

domoritz: Sounds good. I'm using the same pool size in the rust server at https://github.com/uwdata/mosaic/blob/477a97993c4186a35b88242ae37ff4d9b6754c5b/packages/server/duckdb-server-rust/src/app.rs#L57 so this seems reasonable. I used to have issues when we had temp tables (not shared). Prepared queries are also not shared but we don't use them either.

domoritz:
Thank you!

derekperkins:
Done https://github.com/uwdata/mosaic/pull/818

PR #831, opened 2025-07-01
chore: update node deps

domoritz:


domoritz:
After moving to https://github.com/paleolimbot/duckdb-nanoarrow, it looks like tests are passing reliably (ran a few times locally). And I tested the dev server and it worked. 

For moving to Neo arrow, we could use the approach in https://github.com/duckdb/duckdb-node-neo/issues/45#issuecomment-2596841594 but that means we lose streaming.

domoritz:
Dropped the duckdb updates and moved them to https://github.com/uwdata/mosaic/pull/832

jheer (approved):

PR #837, opened 2025-07-04
feat: add list_contains, list_has_all, list_has_any, and unnest

### LINKED ISSUE #830: feat: handle nested arrays for use cases like tags
A common use case for cross-filtering is to have user supplied tags/groups. Those are typically modeled as either an array of strings, or as key/value pairs. For example, if I had a list of blog posts, the table might look something like this.

```sql
CREATE TABLE posts_list (id INTEGER, title VARCHAR, tags VARCHAR[]);

CREATE TABLE posts_map (id INTEGER, title VARCHAR, tags MAP(VARCHAR, VARCHAR));

CREATE TABLE posts_struct_list (id INTEGER, title VARCHAR, tags STRUCT(key VARCHAR, value VARCHAR)[]);
```

The biggest data mistake that I see is users joining against those nested values, creating multiple rows, altering aggregations based on the number of tags in the row, or even eliminating the row altogether if it has no tags.

Would it be possible to make mosaic aware of any/all of these patterns? Conceptually, the cross join should only happen if the `tags` column was part of the group by, otherwise filtering would happen without row expansion. Even if just the simplest `tags VARCHAR[]` version, a simple list of scalar values, was supported, any other data models could be transformed to that in a pre-processing step.

derekperkins:
This is an initial stab at this. I know the tests aren't passing, just wanted to double check that this is heading down the right path before spending too much time on it. Specifically, I wasn't sure if there needed to be a some new types, or if `MaybeArray<ExprValue>` was sufficient

Fixes #830 

cc @jheer


packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,58 @@
...
+import { argsList, fn } from '../util/function.js';
+import { ListNode } from "../ast/list.js";
+import {ExprNode} from "../ast/node.js";
+
+function listFn(name: string, expr: MaybeArray<ExprValue>, ...args: unknown[]) {
+  return fn(name, expr, ...(argsList(args).map(asLiteral)));
+}
+
+/**
+ * Return a SQL AST node for a literal value. The supported types are
```
jheer: This comment needs to be updated as it looks like it was copy and pasted and does not reflect changes

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,58 @@
...
+ * null, string, number, boolean, Date, and RegExp. Otherwise, the
+ * input value will be directly coerced to a string.
+ * @param values
+ */
+export function list(values: ExprNode[]) {
```
jheer: Maybe makes sense to instead have an `ExprValue[]` type, then run a `.map(asLiteral)` over it? That will allow passing in JS values (string, integer, Date) directly. (`asLiteral` treats strings as literal values, `asNode` is similar but treats strings as column references, both will pass through existing `ExprNode` instances unchanged.)

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,58 @@
...
+  return new ListNode(values);
+}
+
+/**
+ * Function that returns true if the list contains the element,
+ * false otherwise.
+ * @param list
+ * @param element
+ */
+export function list_contains(
```
jheer: Elsewhere the SQL package uses camelCase, so to stay consistent the method names would be `listContains`, `listHasAll`, and `listHasAny`.

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,58 @@
...
+  list: MaybeArray<ExprValue>,
+  element: ExprValue
+) {
+  return listFn('list_contains', list, element);
```
jheer: `listFn` supports var args and maps element to `[element]`. So I don't think you want to use `listFn` here?

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,58 @@
+import type {ExprValue, MaybeArray} from '../types.js';
+import { asLiteral } from '../util/ast.js';
+import { argsList, fn } from '../util/function.js';
+import { ListNode } from "../ast/list.js";
+import {ExprNode} from "../ast/node.js";
+
+function listFn(name: string, expr: MaybeArray<ExprValue>, ...args: unknown[]) {
```
jheer: I'm interpreting this as a helper for functions that take two lists as input. Instead of `expr` and `...args` I'd expect arguments like `list1` and `list2`, each receiving identical processing. I think the type signature for an input list should be something like `ExprValue | ExprValue[]`, supporting either a single expression or an array of list content. A string value could be mapped to a column reference. If array-typed, the array could be pre-processed (e.g., using `list`).

jheer (changes requested):

packages/mosaic/sql/src/functions/list.ts
derekperkins: I thought it was strange to be underscores, and thought I had double checked, but I think I saw the function list at the bottom of the aggregate file and assumed the js was matching the duckdb references.
https://github.com/uwdata/mosaic/blob/c381a4422193464aad796d741ad371dd460863ca/packages/mosaic/sql/src/ast/aggregate.ts#L138-L140

jheer: That list is the actual function names used in the emitted SQL. The JS method names are different (see https://github.com/uwdata/mosaic/blob/c381a4422193464aad796d741ad371dd460863ca/packages/mosaic/sql/src/functions/aggregate.ts). Similarly I'd expect `listContains()` JS code to emit `list_contains()` DuckDB SQL.

derekperkins: It makes total sense, I wasn't critiquing the code, just pointing out my airheaded mistake

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -0,0 +1,72 @@
...
+import { asLiteral, asNode } from "../util/ast.js";
+import { argsList, fn } from "../util/function.js";
+import { ListNode } from "../ast/list.js";
+import { literal } from "./literal.js";
+
+/**
+ * Create a List containing the argument values.
+ * @param values
+ */
+export function list(...values: ExprValue[]) {
```
derekperkins: I wondered if this should be called `listValue`, since there's a DuckDB helper for it, but since this is happening in js, I didn't want to use that function name
https://duckdb.org/docs/stable/sql/functions/list#list_valueany-

derekperkins:
Other than my one question about `listValue`, this should be ready for review. Happy to make any changes you want, I tried to match the existing code style, but still don't feel like I have a grasp of the full system.

derekperkins:
docs added

packages/mosaic/sql/src/functions/list.ts
```diff
@@ -33,7 +33,7 @@ export function listContains(
   list1: ExprValue | ExprValue[],
   element: ExprValue,
 ) {
-  return fn("list_contains", exprValuesToExprNode(list1), asLiteral(element));
+  return fn("list_contains", asList(list1), asList(element));
```
derekperkins: I think this should still be `asLiteral(element)`. I imagine the most common use case for this is to filter by a string, so my thought was to optimize for that, vs having it turned into a column reference, which seems very unlikely. I do understand if you want the consistency of single strings being treated as columns across the board, so whatever you think is best.

derekperkins: If you did want to leave it as a column reference for consistency, I would personally prefer to see `asNode(element)` vs `asList(element)`, since a list isn't a valid input. The array check in `asList` would handle that case without error, but it feels a little misleading looking at the code.

jheer: This was a mistake. Please ignore as I fix :) I should know better than to make spot edits on GitHub rather than locally :)

derekperkins: No worries, the rest of the fixes all looked great

PR #839, opened 2025-07-07
feat: support list functions in menu.js

derekperkins:
This builds on #837. Again, I'm just checking to see if this seems like a reasonable direction to take. I'm thinking that inputs can be a little smarter about supporting list columns, they just need a little introspection to know if they're targeting a list column. The column is just a string, so introspection would require a query. I added an additional input necessary to set `any` or `or` for the `listMatch` type.

I also added a hardcoded `UNNEST()`, but it seems like adding an `unnest` js function to the query builder would probably be useful in the future.

The menu is a simpler case, where there is still just a single selection value, but whatever patterns we decide on should be relatively easy to translate to a `select` that allows multiple selections.


derekperkins:
closing in favor of #837

PR #840, opened 2025-07-08
feat: add dataframe interoperability to `MosaicWidget` using `narwhals`

peter-gy:
This PR extends `MosaicWidget` to accept `data` input from a variety of backends using [Narwhals](https://github.com/narwhals-dev/narwhals) as the abstraction layer.

### Overview of Backends Supported by Narwhals

| Backend    | Support Level   |
|------------|----------------|
| pandas     | Full           |
| Polars     | Full           |
| cuDF       | Full           |
| Modin      | Full           |
| PyArrow    | Full           |
| Dask       | Lazy-only      |
| DuckDB     | Lazy-only      |
| Ibis       | Lazy-only      |
| PySpark    | Lazy-only      |
| SQLFrame   | Lazy-only      |
| Daft       | In progress    |

For Pandas, Polars and PyArrow datasets we keep passing the native objects to `DuckDBPyConnection.register` to leverage all of DuckDB's related optimizations.

For other backends, under the hood all we do is use `nw.from_native` to access a uniform API to convert the passed dataframe to Arrow format which can be directly registered to the DuckDB connection. The related performance overhead should be [negligible](https://narwhals-dev.github.io/narwhals/overhead/). Depending on specific backend implementation, the conversion to Arrow may or may not be a zero-copy operation. Users get a warn log about this.

As far as I am aware, DuckDB does not support registering dataframes as virtual tables in lazy format, so we must materialize those before calling `con.register`. The `_frame_to_duckdb_registrable` utility takes care of this.

### Demo Notebook

CleanShot 2025-07-08 at 12 54 16@2x


peter-gy:
@domoritz while at it, I also added a `mypy` step to CI, similarly to what we have for `duckdb-server`.

domoritz:
Thank you!

PR #842, opened 2025-07-08
feat: enable DuckDBConfig for OPFS support

derekperkins:
This exposes the native DuckDBConfig object on the connector, which right now only enables two things:
- path (OPFS support)
- accessMode

I couldn't see a good place to add a test for this, it doesn't appear like there is WASM specific testing. Are there any concerns with this approach?


jheer (approved):
LGTM. Adding some DuckDB-WASM specific tests is a good idea (would use the Node version, obviously), but could be done in a future PR.

jheer:
Thank you!

PR #848, opened 2025-07-18
ci: include `uv.lock` in `cache-dependency-glob`

peter-gy:
Might fix the [most recent](https://github.com/uwdata/mosaic/actions/runs/16366806252/job/46245811354) red CI run on main.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,8 +62,8 @@ jobs:
           cd packages/server/duckdb-server
 
           uv build
-          uv run mypy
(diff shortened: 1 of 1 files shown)


domoritz (approved):
Ah, good catch!

peter-gy:
Not good enough unfortunately... Looking into it.

peter-gy:
Seems to be related to `uv` version. I could not reproduce the CI failure locally with `0.6.4` but could do so with `0.8.0` (same `uv` version used by the GH flow)

domoritz:
I suspect it's pretty rare for uv releases to break things and it's good to be aware.

PR #854, opened 2025-07-28
duckdb-go-server: production style enhancements

### LINKED ISSUE #852: server: add schema validation / authentication
Following up on [a discussion](https://github.com/uwdata/mosaic/discussions/817) about multi-tenancy, I'm thinking about submitting a PR that's fairly opinionated, where the user can provide a flag for which "trusted headers" have to match the schema name. Here's what I envision that flow looking like:
1. Mosaic is configured with `tenant_id`, `user_id`, or something like that as the `preagg.schema`
2. Client sends through authenticated requests. Implementation is out of scope for this PR, but this assumes that there is a cookie or `Authorization` header per request, which is verified through an API Gateway or other mechanism. That middleware would inject some trusted headers, like `X-Tenant-Id: 12345`, `verified-user-id: abc`
3. Mosaic server is started with a flag: `--schema_match_headers=X-Tenant-Id,verified-user-id`
4. When the Mosaic server receives a request, it calls [json_serialize_sql](https://duckdb.org/docs/stable/data/json/sql_to_and_from_json.html) to get the full AST. It recurses through all the schema objects, and returns a 403 error if any schemas are referenced that don't match the header values
5. Finally execute the original SQL (or possibly [json_execute_serialized_sql](https://duckdb.org/docs/stable/data/json/sql_to_and_from_json) if there are performance benefits) and return the results as normal

A production implementation might pull those valid values out of a jwt or other mechanisms, headers just seems like the broadest solution for OSS. If that sounds like a reasonable fit, I'm happy to submit a PR to the Go server.

There's a larger question around the purpose of the servers (currently python, rust, and go). Are they there more as an option to have a plug and play server to use with mosaic, or if they're more intended as a proof of concept / implementation starter. The Go server I contributed was more in line with the latter, assuming that most use cases would have custom logic to run alongside the code, but maybe I'm thinking about it the wrong way.

Related Discussion:
- https://github.com/uwdata/mosaic/discussions/817

derekperkins:
Fixes #852

### DIFF
--- a/packages/server/duckdb-server-go/.golangci.yaml
+++ b/packages/server/duckdb-server-go/.golangci.yaml
@@ -7,6 +7,17 @@ run:
 linters:
   default: standard
 
+  enable:
+    - rowserrcheck
+    - sqlclosecheck
+
+  settings:
+    errcheck:
+      # List of functions to exclude from checking, where each entry is a single function to exclude.
+      # See https://github.com/kisielk/errcheck#excluding-functions for details.
+      exclude-functions:
+        - (*database/sql.Rows).Close
+
 formatters:
   enable:
     - gofmt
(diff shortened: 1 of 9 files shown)


jheer (approved):
I'm not know Go well, so I'm not equipped to make a proper review. But these changes look valuable -- and better to have than not have!

jheer:
Thanks @derekperkins!

PR #860, opened 2025-08-01
ci: fix occasional test faliures

domoritz:
This should fix these occasional issues (which were bothering me enough to fix it). 

Screenshot 2025-08-01 at 15 57 41

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -854,49 +854,6 @@
         }
       }
     },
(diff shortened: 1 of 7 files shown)


kwonoh (approved):

domoritz:
Okay. Let me merge this so you can benefit from it. We can change it if there are better approaches.

PR #863, opened 2025-08-05
feat: add more set operations

kwonoh:
Add more set operations https://duckdb.org/docs/stable/sql/query_syntax/setops.html

### DIFF
--- a/docs/api/sql/queries.md
+++ b/docs/api/sql/queries.md
@@ -55,9 +55,9 @@ In addition, the following static methods take multiple queries as input and ret
 - `Query.unionAll(...queries)`: Union results with no de-duplication of rows.
 - `Query.unionAllByName(...queries)`: Union results with no de-duplication of rows, combining rows from different tables by name, instead of by position.
 - `Query.intersect(...queries)`: Query for distinct rows that are output by both the left and right input queries.
-- `Query.intersectAll(...queries)`: Query for rows that are output by both the left and right input queries using bag semantics, so duplicates are returned.
+- `Query.intersectAll(...queries)`: Query for all rows that are output by both the left and right input queries using bag semantics, so duplicates are returned.
 - `Query.except(...queries)`: Query for distinct rows from the left input query that aren't output by the right input query.
-- `Query.exceptAll(...queries)`: Query for rows from the left input query that aren't output by the right input query using bag semantics, so duplicates are returned.
+- `Query.exceptAll(...queries)`: Query for all rows from the left input query that aren't output by the right input query using bag semantics, so duplicates are returned.
 
 Common table expressions can be applied via static method
 


derekperkins:
LGTM

docs/api/sql/queries.md
```diff
@@ -51,9 +51,13 @@ The following static methods create a new `SelectQuery` and invoke the correspon
 In addition, the following static methods take multiple queries as input and return `SetOperation` instances:
 
 - `Query.union(...queries)`: Union results with de-duplication of rows.
+- `Query.unionByName(...queries)`: Union results with de-duplication of rows, combining rows from different tables by name, instead of by position.
 - `Query.unionAll(...queries)`: Union results with no de-duplication of rows.
+- `Query.unionAllByName(...queries)`: Union results with no de-duplication of rows, combining rows from different tables by name, instead of by position.
 - `Query.intersect(...queries)`: Query for distinct rows that are output by both the left and right input queries.
+- `Query.intersectAll(...queries)`: Query for distinct rows that are output by both the left and right input queries using bag semantics, so duplicates are returned.
```
domoritz: Thanks for the pull request. I'm curious whether you have thought through the implications for using bag semantics in mosaic? Or did you add this because it's supported by DuckDB?

kwonoh: Hi @domoritz, I have use cases for `unionByName` and `unionAllByName` at this moment. I have added `intersectAll` and `exceptAll` as well, since they are available in DuckDB and other DBMS/query engines (e.g., [PostgresSQL](https://www.postgresql.org/docs/current/queries-union.html), [Trino](https://trino.io/docs/current/sql/select.html#set-operations), [Spark SQL](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html))

I believe the implications of this PR are minimal, as these are new functions and don't modify any existing behavior. Do you have any concerns about adding these?

domoritz: No concerns. Just curious how bag semantics works with Mosaic's query model.

kwonoh: I'm not sure I’m fully following your question, but:

In the context of Mosaic's Query AST, these are additional [`op` values for `SetOperation`](https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/sql/src/ast/query.ts#L522).

As for query results, here’s a quick example:
```
SELECT unnest([5, 5, 6, 6, 6, 6, 7, 8]) AS x
INTERSECT
SELECT unnest([5, 6, 6, 7, 7, 9]);

┌───────┐
│   x   │
│ int32 │
├───────┤
│     5 │
│     7 │
│     6 │
└───────┘
```


```
SELECT unnest([5, 5, 6, 6, 6, 6, 7, 8]) AS x
INTERSECT ALL
SELECT unnest([5, 6, 6, 7, 7, 9]);

┌───────┐
│   x   │
│ int32 │
├───────┤
│     6 │
│     6 │
│     7 │
│     5 │
└───────┘
```

domoritz: I understand the bag semantics but not how it would be useful with mosaic. I guess you want to use bag semantics and not set semantics before a group by aggregate so that you don't collapse items that should be separate. 

I see the value. I approve.

kwonoh:
@domoritz @jheer any concerns on merging this?

PR #865, opened 2025-08-21
fix: destroy is part of MosaicClient

severo:
From the MosaicClient class, I understand there is no need to add a destroy method manually. 

https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/core/src/MosaicClient.ts#L214-L224


domoritz:
Good catch. That was added in https://github.com/uwdata/mosaic/commit/1c29646087aa630a84d66df9d29b1d21c09d2160#diff-9c821b9302d6c6346c364803a1c4a8da10cae40065f1e0bb8cf35b6e57de7876.

PR #867, opened 2025-08-21
disable client before destroying to cancel pending queries

### LINKED ISSUE #866: destroy the client in React example throws an error
I used the React example, with DuckDBWASM (I don't think it's related), and I found that destroying the client when useEffect is re-run throws an error:

```
TypeError: can't access property "requestQuery", this._coordinator is null
```

The stack:

```
Source
requestQuery
webpack-internal:/node_modules/@uwdata/mosaic-core/dist/src/MosaicClient.js (156:1)
initialize/this._pending<
webpack-internal:/node_modules/@uwdata/mosaic-core/dist/src/MosaicClient.js (191:1)
promise callback*initialize
webpack-internal:/node_modules/@uwdata/mosaic-core/dist/src/MosaicClient.js (191:1)
connect
webpack-internal:/node_modules/@uwdata/mosaic-core/dist/src/Coordinator.js (178:1)
makeClient
webpack-internal:/node_modules/@uwdata/mosaic-core/dist/src/make-client.js (12:1)
```

I guess it's because we remove the client's connector while the initialization has not finished (`this.prepare()` is still pending):

https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/core/src/MosaicClient.ts#L210

The `destroy` method removes the connector:

https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/core/src/MosaicClient.ts#L223

Then, when `this.prepare()` resolves, it assumes `this._coordinator` exists, which is no longer the case.

https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/core/src/MosaicClient.ts#L176

A solution might be to set `this._enable = false` in destroy, because it's used as a condition in `.requestQuery`:

https://github.com/uwdata/mosaic/blob/3ef701b545160e69704888eab2d50867ef45fdc8/packages/mosaic/core/src/MosaicClient.ts#L173-L181.

---

Note that it occurs when React is in [Strict mode](https://react.dev/reference/react/StrictMode), because all the components are rendered twice immediately.

severo:
fixes #866.

### DIFF
--- a/packages/mosaic/core/src/MosaicClient.ts
+++ b/packages/mosaic/core/src/MosaicClient.ts
@@ -220,6 +220,8 @@ export class MosaicClient {
    * disconnect from the coordinator.
    */
   destroy(): void {
+    // disable client to cancel current queries.
+    this._enabled = false;
     this.coordinator?.disconnect(this);
   }
 


packages/mosaic/core/src/MosaicClient.ts
```diff
@@ -220,6 +220,8 @@ export class MosaicClient {
    * disconnect from the coordinator.
    */
   destroy(): void {
+    // disable client to cancel current queries.
```
domoritz:
```suggestion
```

domoritz: It's not really cancelling queries. It's just preventing them from crashing when the promises return.

severo: yes, indeed

PR #874, opened 2025-09-07
fix: clause.source doesn't accept a MosaicClient

donghaoren:
The symptom: clause.source doesn't accept a MosaicClient. If you define a simple client like:

```js
let client = makeClient(...);

selection.update({
  source: client,
  ...
});
```

You'll get a type error on "Type 'MosaicClient' has no properties in common with type 'ClauseSource'". Right now we have to cast the client to `ClauseSource` like `client as ClauseSource`.

I think we can fix it by changing `ClauseSource` to:

```ts
export type ClauseSource = object & { reset?: () => void; };
```

This should match any object, but also objects with a reset method (and if there is a reset method, it must match be of type `() => void`) .

### DIFF
--- a/packages/mosaic/core/src/SelectionClause.ts
+++ b/packages/mosaic/core/src/SelectionClause.ts
@@ -59,9 +59,7 @@ export interface IntervalMetadata extends ClauseMetadata {
   bin?: BinMethod
 }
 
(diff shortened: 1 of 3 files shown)


jheer (approved):
Thanks!

PR #879, opened 2025-09-10
duckdb-server-go: refactor config + add additional validators

derekperkins:
This PR adds three new flags, to continue making the server viable for safe use in production, with arbitrary users sending queries via the api.
-   `--max-cache-bytes <bytes>`: Max number of cache size in bytes (overrides max-cache-entries if both are set). Defaults to 0 (no limit).
-   `--cache-ttl <duration>`: Time-to-live for cache entries as Go duration. 0s means no expiration (e.g., '10m', '1h'). Defaults to 0s.
-   `--function-blocklist`: Comma-separated list of functions to block, useful for blocking functions that may pose security or performance risks. (e.g., 'bigquery_query,read_parquet')`


derekperkins:
This should be ready to merge, I think the node failure is just a flaky test

PR #886, opened 2025-09-17
feat: extract logic for toString into visitor

domoritz:
See https://github.com/uwdata/mosaic/pull/880#issuecomment-3303726161


packages/mosaic/sql/src/init.ts
```diff
@@ -0,0 +1,5 @@
+import { setDefaultVisitor } from './ast/node.js';
+import { duckdbVisitor } from './visit/duckdb-visitor.js';
+
+// Initialize the default visitor
+setDefaultVisitor(duckdbVisitor);
```
domoritz: This causes a side effect which is probably not good. If we are okay with the general approach in this pull request, I can remove this and change the rest of mosaic to initialize the visitor.

packages/mosaic/sql/src/visit/to-string-visitor.ts
```diff
@@ -0,0 +1,204 @@
...
+  abstract visitUnnest(node: UnnestNode): string;
+  abstract visitVerbatim(node: VerbatimNode): string;
+  abstract visitWhen(node: WhenNode): string;
+  abstract visitWindow(node: WindowNode): string;
+  abstract visitWindowClause(node: WindowClauseNode): string;
+  abstract visitWindowDef(node: WindowDefNode): string;
+  abstract visitWindowExtentExpr(node: WindowFrameExprNode): string;
+  abstract visitWindowFrame(node: WindowFrameNode): string;
+  abstract visitWindowFunction(node: WindowFunctionNode): string;
+  abstract visitWithClause(node: WithClauseNode): string;
```
derekperkins: I wonder if this makes future changes tougher at all. For instance if we add a new abstract method, that same PR would have to implement the methods for all visitor implementations, right? In general that seems ok, but if we end up with 5-10 visitors, the person adding new functionality might only be versed in one of them, and wouldn't feel comfortable editing all of them.

Maybe if this abstract class had concrete methods that returned an error for not being implemented, then each language implementation wouldn't have to be 100% in lockstep as new features get supported. Instead of a build error that fails CI, it's now a runtime error if you try to use a new feature with an unsupported grammar, but allows existing queries to complete without issue. gRPC in Go requires something similar, where a default [unimplemented concrete server](https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/README.md) returns errors if not explicitly overridden.

Another approach might be to have this be a concrete implementation of the DuckDB grammar. That solves the problem from above. It also should slim down the per-database implementation, since most of the basic `SELECT`, `ORDER BY`, etc., code wouldn't have to be copy/pasted into the concrete implementations. You would only have to override specific methods that deviate from the DuckDB output. That means that the DuckDB visitor couldn't be tree-shaken out, but that doesn't seem like an issue worth optimizing for. It also might make for slightly more brittle language implementations, in the event that DuckDB diverges in grammar, but that seems very unlikely for the shared cases, and again probably not worth bike shedding over.

Overall, I love how this is shaping up. I'm not married to either of these suggestions, just talking through possible design ramifications.

packages/mosaic/sql/src/init.ts
derekperkins: Agreed, I'd much rather see explicit visitor setup, even if it's doing this same thing in a constructor

packages/mosaic/sql/src/visit/to-string-visitor.ts
domoritz: For the first issue, I am imagining that most dialects would not be in the core mosaic repo so one would only need to implement a new method when updating mosaic. I can also imagine that people would extend the duckdb implementation rather than the base visitor. So I think we can already so that you propose but I separated out the duckdb implementation rather than merging it with the base class which means one can make their totally independent version if desired.

derekperkins: Perfect, I've been in Go and out of OOP/class driven languages too long.  No other notes on the implementation, I think it's good to merge unless @jheer has feedback. 

Are you opposed to keeping the first few dialects in the core repo for visibility?

domoritz: Nope, we can keep the first few dialects as long as we have testing for them to make sure they actually work

packages/mosaic/sql/src/visit/codegen/duckdb.ts
```diff
@@ -0,0 +1,439 @@
...
+
+    if (joinVariant !== 'CROSS') {
+      type = joinType !== 'INNER' ? `${joinType} ` : '';
+      cond = condition ? ` ON ${condition}`
+        : using ? ` USING (${using?.join(', ')})`
+        : '';
+    }
+    const samp = sample ? ` USING SAMPLE ${sample}` : '';
+    return `${left} ${variant}${type}JOIN ${right}${cond}${samp}`;
+  }
```
derekperkins, line 177: I don't know if this is a real issue or just a bug in GitHub's code highlighting, but is there a string escape mismatch here? It seems like that would've broken the build, which is why I'm skeptical that it's a real bug

Image

packages/mosaic/sql/src/visit/codegen/duckdb.ts
```diff
@@ -1,4 +1,3 @@
-import { ToStringVisitor } from './to-string-visitor.js';
```
derekperkins: I don't feel strongly about this, but having these under the `visit/codegen` folder has a connotation to me that these were generated by code from a spec, vs handwritten here to generate SQL.

packages/mosaic/sql/src/visit/codegen/duckdb.ts
domoritz: That's just GitHub

Screenshot 2025-09-18 at 14 53 20

packages/mosaic/sql/src/visit/codegen/duckdb.ts
jheer: If others feel similarly I have no problem changing this. It's just internal organization so doesn't affect the exported API.

derekperkins:
This is awesome, I'll open a PR soon for BigQuery support. I'm at a conference next week, so it may be the week after

PR #888, opened 2025-09-18
Add join SQL helpers.

jheer:
- Add `JoinNode` SQL AST node type.
- Add `join`, `cross_join`, `asof_join`, and `positional_join` helper functions.
- Add `from` helper method for creating table references for query from and join contexts.


domoritz (approved):

packages/mosaic/sql/src/functions/from.ts
```diff
@@ -0,0 +1,18 @@
+import { FromClauseNode } from "../ast/from.js";
+import { SampleClauseNode } from "../ast/sample.js";
+import { TableRefNode } from "../ast/table-ref.js";
+import { asTableRef } from "../util/ast.js";
+
+/**
+ * Create a new table FROM reference, applicable in a query or join .
```
domoritz, line 7:
```suggestion
 * Create a new table FROM reference, applicable in a query or join.
```

PR #895, opened 2025-09-19
feat: add BigQuery dialect

derekperkins:
This isn't tested yet, I just wanted to make sure I'm heading in the right direction. This may not be 100% comprehensive with all the language differences, but it should be pretty close, at least for the major functions.

### Function node
The function node is probably going to be the hairiest to support, since it's an untyped string match, and requires some arg reordering, but wasn't bad. Do you see any issues with the direct arg references?

### Types
There were a few untyped strings I ran across that would be easier to translate if they were a string union. I should've kept a better list, but these are the ones I remember:
- CastNode.cast
- IntervalNode.name

Maybe the function names in the function node could be too?

### Casing
DuckDB by convention uses lowercase, BigQuery uses uppercase. I wrote the override functions in uppercase, not sure if those should still all be lowercase. I don't feel strongly about it.

### Unsupported errors
Right now I'm checking for unsupported cases and throwing errors, following the example of a single throw in the DuckDB visitor. Not sure if there's a pattern otherwise to follow.

### Testing
We moved the visitor out, but not the tests. Should these sql tests be moved into per visitor files? I can add specific BigQuery tests once we decide on that organization.
https://github.com/uwdata/mosaic/tree/c381a4422193464aad796d741ad371dd460863ca/packages/mosaic/sql/test

### AST manipulation
There are a few operations that could be supported, but not while walking through `toString`. One example is `SEMI|ANTI JOIN`. BigQuery and Snowflake don't support them, but they can be trivially added by adding a `WHERE` condition. The database engine shouldn't leak into the AST creation, where individual nodes have to run a `supportsSemiJoin` type of scenario, which is the other place I can think to add it. Maybe an optional method like `manipulateAst` could be exposed, and executed if the visitor needed to. There's nothing urgent on this, as Mosaic just barely got join support yesterday, so I wouldn't spend any time on it unless there was a materialized view use case for the coordinator. In the meantime, I filed an issue with Google to add syntax support. (https://issuetracker.google.com/issues/446157721)

### DIFF
--- a/packages/mosaic/sql/src/visit/codegen/bigquery.ts
+++ b/packages/mosaic/sql/src/visit/codegen/bigquery.ts
@@ -1,0 +1,106 @@
+import {
+  AggregateNode,
+  CastNode,
+  FunctionNode,
+  JoinNode,
+  SampleClauseNode,
+  UnnestNode,
+} from "../../index.js";
+import { DuckDBCodeGenerator } from "./duckdb.js";
+
+/**
+ * BigQuery SQL dialect visitor for converting AST nodes to BigQuery-compatible SQL.
+ */
+export class BigQueryCodeGenerator extends DuckDBCodeGenerator {
+  visitAggregate(node: AggregateNode): string {
+    const { filter } = node;
+    // TODO: it might be possible to emulate filters using an IF statement
+    if (filter) {
+      throw new Error(`BigQuery does not support aggregate filters. Node: ${JSON.stringify(node)}`);
+    }
+    return super.visitAggregate(node);
+  }
+
+  visitCast(node: CastNode): string {
+    const { expr, cast } = node;
+    return `CAST(${this.toString(expr)} AS ${cast})`;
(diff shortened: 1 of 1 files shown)


derekperkins:
> creating a new dedicated file/folder for BigQuery tests

I'll go ahead and add a new folder. Should I copy 100% of the tests, or only the tests for the overridden methods?

domoritz:
I think for now testing only the only the functions that are different. Otherwise, we will have a lot of duplication.

domoritz:
I'll make this as draft since it needs tests and some more integration etc but I'm excited to eventually get this in.

derekperkins:
We're going full DuckDB, so closing this. I think it's pretty close and AI could land it. It's not really testable without adding remote access testing on some GCP account

domoritz:
Makes sense. I think having #1163 as a demo of another dialect would be great and so we don't need big query as an example as much anymore.

PR #898, opened 2025-09-27
chore: add visual tests to reduce regressions

domoritz:
Uses playwright to render the examples and compares them to stored snapshots. You can update the examples locally with docker. The GitHub Action also uploads the diffs so you don't need to use docker. 

I'm not a fan of using a timeout to wait for renders. Is there a reliable way to wait for render to be done?


jheer (commented):
Top level, it would be great to have tests of rendered specs so we can be more robust and against breakage, especially with new collaborators and/or AI assisted code. I don't like inflated test run times but can live with that if not egregious.

One concern is generating the snapshots. If you have to install Docker etc its a bigger lift that adds overhead to development setup. Of course, different browsers / OSes render differently so if you go through a browser you don't have a choice. At minimum I think we want to have scripts (not just README instructions) in place to minimize effort if we go this direction.

An alternative would be to do server-side rendering via jsdom or similar. That has its own drawbacks but tends to be fast and portable.

bin/visual-test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
jheer, line 1: Maybe add comment that makes it clear this is for local testing only?

jheer (approved):
Looks good! Just need to resolve conflicts.

PR #907, opened 2025-10-07
feat: update duckdb in python to 1.4.1

domoritz:
Fixes issue with changes in duckdb which now returns a `pyarrow.lib.RecordBatchReader` instead of `Table`.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -750,7 +750,7 @@ wheels = [
 
 [[package]]
 name = "ibis-framework"
-version = "10.8.0"
+version = "11.0.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "atpublic" },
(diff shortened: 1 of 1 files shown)


packages/vgplot/widget/pyproject.toml
```diff
@@ -5,9 +5,9 @@ dynamic = ["version"]
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-  "anywidget>=0.9.0",
-  "duckdb>=1.3.1",
-  "narwhals>=1.46.0",
+  "anywidget>=0.9.8",
+  "duckdb==1.4.0",
```
donghaoren: `>=` ? It's `>=` above in `duckdb-server`.

domoritz: Yeah, it's needed for now because of https://github.com/narwhals-dev/narwhals/issues/3137.

domoritz: Works now. Updated to 1.4.1

PR #910, opened 2025-10-17
fix: remove cache from fetch options

kwonoh:
Remove `cache: 'no-cache'` from RestConnector. 

Having `cache: 'no-cache'` causes repeated OPTIONS requests to check CORS status unnecessarily. This increases overall response time.

Also, I believe HTTP caching rarely applies to POST anyway. 

With `cache: 'no-cache'`:
Screenshot 2025-10-15 at 15 32 54

Without setting `cache` option:
Screenshot 2025-10-15 at 15 31 14

Alternatively, we could add a param for `fetch` call in the constructor of RestConnector similar to `ipc`


derekperkins:
I've been meaning to look into why that was happening. This is great

PR #914, opened 2025-10-22
Escape quotes in SQL identifier strings.

jheer:
- Fix: escape quotes in SQL identifier strings during mosaic-sql string serialization.
- Add test cases for mosaic-sql column references.

cc: @flekschas

### DIFF
--- a/packages/mosaic/sql/src/util/string.ts
+++ b/packages/mosaic/sql/src/util/string.ts
@@ -5,9 +5,7 @@ export function parseIdentifier(id: string) {
 
 export function quoteIdentifier(value: string) {
   // escape double quotes if they occur in identifier
-  return value.includes('"')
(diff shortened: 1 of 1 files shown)


packages/mosaic/sql/src/util/string.ts
```diff
@@ -3,8 +3,11 @@ export function parseIdentifier(id: string) {
   return id.split('.');
 }
 
-export function quoteIdentifier(value: unknown) {
-  return `"${value}"`;
+export function quoteIdentifier(value: string) {
+  // escape double quotes if they occur in identifier
```
domoritz, line 7: Why not always replace?

jheer: No good reason. Will simplify.

PR #929, opened 2025-11-04
Add vgplotr: R package for vgplot visualizations in Quarto

kcarnold:
This commit introduces vgplotr, an R htmlwidget package that brings Mosaic vgplot visualizations to R, Quarto documents, and RMarkdown.

Key Features:
- Full DuckDB WASM integration for scalable data processing
- Interactive crossfilter selections and linked brushing
- Grammar of graphics API inspired by ggplot2
- Direct loading of data from URLs (Parquet, CSV, etc.)
- Self-contained Quarto/RMarkdown outputs
- Shiny application support

Package Structure:
- R API functions for marks, plots, selections, and layouts
- htmlwidgets binding for browser-based rendering
- JavaScript build system using esbuild
- DuckDB WASM for client-side query execution
- Mosaic coordinator for reactive updates

Examples:
- flights-demo.qmd: 10M flight records with linked histograms
- getting-started.qmd: Basic usage and common patterns

Documentation:
- README.md: Complete API reference and examples
- INSTALL.md: Build and installation instructions
- Comprehensive inline R documentation

The package enables users to create interactive visualizations like:

```r
brush <- Selection$crossfilter()
make_plot <- function(col) {
  plot(
    rectY(from("data", filterBy = brush), list(x = bin(col), y = count())),
    intervalX(as = brush),
    xDomain(Fixed)
  )
}
mosaic_spec(vconcat(make_plot("x"), make_plot("y")))
```

PR #930, opened 2025-11-07
Add UMD bundle

kcarnold:
I've been trying to help @rpruim make an R package, and one of the challenges that emerged was that some of the runtime environments don't support `import()`, or need to run offline. See, for example, https://github.com/uwdata/mosaic/pull/929#issuecomment-3495067353

This PR is an attempt at building a self-contained UMD bundle. I don't know if this is even a direction you'd like to go, yet alone whether this approach coheres with your vision of how things should modularize or how the build process should work.

I actually made two different attempts at this; this one has the simpler changes but I've only actually tested my other one: https://github.com/uwdata/mosaic/compare/main...kcarnold:mosaic:add-umd-build?expand=1

Comments welcome.

### DIFF
--- a/packages/vgplot/vgplot/package.json
+++ b/packages/vgplot/vgplot/package.json
@@ -34,7 +34,7 @@
     "prebuild": "npm run clean",
     "build": "npm run build:types && npm run build:umd",
     "build:types": "tsc --build",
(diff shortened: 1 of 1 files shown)


domoritz:
Closing unless we change our stance on supporting UMD going forward.

PR #935, opened 2025-11-23
feat: export types from mosaic-spec

### LINKED ISSUE #934: export types from mosaic spec
Hi again. I'm building a light chart builder, and I think it would be really helpful if the types from https://github.com/uwdata/mosaic/tree/main/packages/vgplot/spec/src/spec are exported. Unless I'm missing something and these types are already exported, it would help builders create charts easier.

Light2Dark:
Fixes #934 

This is optional, but I also added a test to make sure everything in the spec folder is exported.
Makes it so that adding new ts files in the future would always need to be exported.
I'm open to removing this test too.


packages/vgplot/spec/src/index.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ * @typedef {import('./spec/Spec.js').Spec} Spec A Mosaic declarative specification.
```
domoritz: I don't think we need this anymore, do we?

packages/vgplot/spec/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+export type * from "./spec/PlotInteractor.js";
+export type * from "./spec/PlotLegend.js";
+export type * from "./spec/PlotMark.js";
+export type * from "./spec/PlotTypes.js";
+export type * from "./spec/Spec.js";
+export type * from "./spec/Transform.js";
+export type * from "./spec/VConcat.js";
+export type * from "./spec/VSpace.js";
+
+// Marks
```
domoritz: What if instead we add `./spec/marks/index.ts` which exports types?

packages/vgplot/spec/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+export type * from "./spec/marks/Raster.js";
+export type * from "./spec/marks/Rect.js";
+export type * from "./spec/marks/Regression.js";
+export type * from "./spec/marks/Rule.js";
+export type * from "./spec/marks/Text.js";
+export type * from "./spec/marks/Tick.js";
+export type * from "./spec/marks/Vector.js";
+export type * from "./spec/marks/Waffle.js";
+
+// Interactors
```
domoritz: What if instead we add `./spec/interactors/index.ts` which exports types?

domoritz (approved):
Thank you. At some point it would make sense to convert the rest of the js files in this package to ts but that's an orthogonal issue.

packages/vgplot/spec/src/index.ts
Light2Dark: Yes, we can do this. A small tradeoff is that I'll remove the test to guarantee all .ts files are exported. I think that's fine..

domoritz: I agree. That's fine.

Light2Dark: Got it, made the change. Thanks! 
also agree to eventually convert all files to ts

packages/vgplot/spec/src/index.ts
domoritz, line 1: Don't we still need the `Spec` import? What was it for before? Please check the compiled files.

Light2Dark: Are you referring to the typedef comment?
```typescript
/**
 * @typedef {import('./spec/Spec.js').Spec} Spec A Mosaic declarative specification.
 */
```
I'm not very familiar with typedef, but I believe this is a jsdoc syntax for declaring types. Since this is now a ts file, I don't think it's needed. We also do export the Spec type 
`export type * from "./Spec.js";` in packages/vgplot/spec/src/spec/index.ts

Issue #943, opened 2025-12-02
Integrated dev tools


derekperkins:
When doing dev on mosaic, it would be really handy to have a query log, when things get materialized or not, current state of selectors, etc. TanStack has some great framework agnostic devtools for their various projects, which are now being opened up to make it easier for other tools to build. Would you be interested in having something like this contributed upstream here?

https://tanstack.com/devtools/

cc @tannerlinsley @SeanCassiere @ChristianCasazza

derekperkins:
To give a better feel for things, here are a few examples of what I'd have in mind:

### TanStack Query
Show all the queries, and you can click into them to see the data. We could potentially do this per mosaic query, limiting retention to a handful of rows maybe.

<img width="1477" height="588" alt="Image" src="https://github.com/user-attachments/assets/c74088d1-f5cb-47a8-9ad6-e529c3762e41" />

### TanStack Router

Shows the current state of routing params. Maybe this could be a minimal built in data viewer?

<img width="1396" height="629" alt="Image" src="https://github.com/user-attachments/assets/c0afc750-78ee-47b3-a33e-372015752d24" />

### OSS Data Tool (https://github.com/theVedanta/meridian)

This is what originally triggered my thoughts around this, just being able to see what queries were being made, and annotate them with details about how/why a materialized view was made, or maybe couldn't be made. Show when there's a materialized view hit, etc. You could filter by client name, so you could see if you have a component that's overly trigger happy.

<img width="359" height="354" alt="Image" src="https://github.com/user-attachments/assets/18f4004d-403f-44d5-8d59-a920e6c69c51" />

domoritz:
I'm a huge fan of having good debug tooling. Right now, we have the browser console/terminal logs as the main means of debugging. I'd be interested in upstreamed tooling if it doesn't add a lot of dependencies and is lightweight.

derekperkins:
Agreed. The idea would be to have a separate package to opt into the dev tools, and only set it up as a dev dependency
- e.g. https://www.npmjs.com/package/@tanstack/query-devtools
- https://tanstack.com/devtools/latest/docs/production

PR #946, opened 2025-12-15
feat: Devtools -- Queries, Client State, and other relevant info

theVedanta:
This PR is in response to #943 
Currently, the devtools (uses tanstack) is coupled with the core, which may increase bundle size. So, we are looking to move to a pub-sub system. The QueryManager and Coordinator basically integrate event emitter right next to the logger's code and some UI in the HTML connects to the Event Bus.

I would love some thoughts on the pub-sub system and how to keep the bundle size down by decoupling the core with devtools and what the "ideal" API design there looks like.

https://github.com/user-attachments/assets/2a2b97d6-f619-4b9b-a248-83cf695c0a7e


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
domoritz, line 1: Don't we already have prettier working?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -105,6 +114,52 @@ export class QueryManager {
 
       if (cache) this.clientCache!.set(sql!, data);
 
+      // Emit 'queries:end' event
```
domoritz, line 117: Let's not add comments than explain code. Add comments that explain why, not what.

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -291,6 +290,24 @@ export class Coordinator {
 
     // connect filter selection
     connectSelection(this, client.filterBy!, client);
+
+    clientEvents.emit('state', {
+      clients: Array.from(this.clients).map((client) => ({
+        // @ts-expect-error - client id exists, just is not typed here
```
domoritz, line 296: let's type it

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -291,6 +290,24 @@ export class Coordinator {
...
+        clientId: client.id, // Not sure why isn't a part of the type here
+        connected: true
+        // lastQuery: client.lastQuery ?? client._lastQuery,
```
domoritz, line 299: outdated code?

packages/devtools/src/devtools.ts
```diff
@@ -0,0 +1,448 @@
...
+                : 'not materialized';
+            }
+
+            header.appendChild(right);
+
+            item.appendChild(header);
+
+            // details (hidden by default)
+            const details = document.createElement('div');
+            details.style.marginTop = '8px';
```
domoritz, line 222: Can we move styling to css?

dev/query/index.html
```diff
@@ -28,8 +28,9 @@
     <button id="submit">Query</button>
     <br/>
     <div id="output"></div>
+    <div id="devtools"></div>
```
domoritz, line 31: It took me forever to find the button for the dev tools in the bottom left. I think it should be more prominent. Also, can we not have it tanstack branded?

packages/devtools/src/devtools.ts
```diff
@@ -0,0 +1,448 @@
...
+  let devtoolsDiv = document.getElementById('devtools');
+  if (!devtoolsDiv) {
+    devtoolsDiv = document.createElement('div');
+    devtoolsDiv.id = 'devtools';
+    document.body.appendChild(devtoolsDiv);
+  }
+
+  const { queryEventClient, clientEventClient } = createEventClients();
+
+  const devtools = new TanStackDevtoolsCore({
```
domoritz, line 110: Most of this seems to be about logging which browser consoles already support pretty well (with console.log and even grouping etc (e.g. https://developer.mozilla.org/en-US/docs/Web/API/console/group_static).

domoritz (commented):
Thanks for the prototype. I can definitely see the value of having a debug UI that I customized for mosaic. Similar to how we added debugging features in the Vega editor (https://vega.github.io/editor/). One thing I would like to learn more about is how we provide more value beyond logging (and maybe some way to customize what types of events get logged). 

* "No additional plugins available. You have all compatible devtools installed and registered!" then we shouldn't have the marketplace. 
* clicking query logs shows a bunch of clients logs?

Image

* not sure what the social previews are for
* there are a lot of messages in the console which make it hard to see the mosaic logs
* I'm not sure I get the benefit of this yet over just logs. I thing we can have a valuable debug UI but I'd say let's build it from scratch rather than pulling in tanstack which seems to offer a lot more than what we need and doesn't make it clearer to support what we want. Building a log with some augmentation for mosaic relevant metadata is easy enough to build from scratch rather than relying on a framework that developers need to learn. Having said that, I think it's great that we can learn from tanstack's design to see what's possible.

.prettierrc
theVedanta: I didn't see a config for it and my editor was going crazy with the edits so I thought it would be helpful to add. Can remove if you want me to :)

packages/mosaic/core/src/QueryManager.ts
theVedanta: My bad, fixing now

packages/devtools/src/devtools.ts
theVedanta: For sure, this is just a simple POC in HTML that is probably not what the devtools would look like. The idea is to have an interactive set of useful utilities (perhaps some react/solid examples will better showcase them)

dev/query/index.html
theVedanta: It took me a while at first too haha; and I'll ask about a custom devtools trigger

domoritz: I think a custom UI makes more sense t this point just to keep things light.

packages/devtools/src/devtools.ts
domoritz: Yeah, sorry, I dove to deep into the code. Should have stayed more high level.

packages/mosaic/core/src/Coordinator.ts
domoritz: later

packages/mosaic/core/src/Coordinator.ts
domoritz: later

packages/mosaic/core/src/QueryManager.ts
domoritz: can do later when we have something we actually want to merge

.prettierrc
domoritz: huh, if we want to change prettier, let's do that in a separate pull request.

PR #949, opened 2025-12-24
chore: add prettier

theVedanta:


.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "none"
```
domoritz: I think we want a print width of 120

domoritz (changes requested):
At the moment, we don't use prettier. 

To add this we would need to
* make prettier a dev dependency
* add prettier support to eslint
* ignore schema files

We should make sure that there are no degredations after we add prettier.

theVedanta:
> At the moment, we don't use prettier.
> 
> To add this we would need to
> 
> * make prettier a dev dependency
> * add prettier support to eslint
> * ignore schema files
> 
> We should make sure that there are no degredations after we add prettier.

Addressed these. Hope I added all the necessary schema files in `prettierignore`

Also added printWidth

domoritz:
Do we need a prettierignore? Doesn't eslint do that? Or is that for editors with auto formatter?

domoritz:
Ci is failing btw. We probably need to reformat as well.

theVedanta:
yep, lemme check it out

eslint.config.js
```diff
@@ -1,6 +1,7 @@
 import js from '@eslint/js';
 import tseslint from 'typescript-eslint';
 import globals from 'globals';
+import prettier from 'eslint-config-prettier';
```
domoritz: shouldn't this be `import eslintConfigPrettier from "eslint-config-prettier/flat";`? That's what https://www.npmjs.com/package/eslint-config-prettier says.

theVedanta: sorry I forgot tseslint returns the flat config array. Fixed now!

derekperkins:
Should we add a command to package.json to write the changes?

package.json
```diff
@@ -19,6 +19,7 @@
     "clean": "lerna run clean",
     "lint": "lerna run lint",
     "test": "tsc --build && lerna run schema && vitest run",
+    "write": "prettier --write '**/*.{js,ts,json,yaml}'",
```
domoritz: No, please add `format` or `lint:fix` and let ESLint fix things.

docs/public/contours.yaml
```diff
@@ -4,29 +4,29 @@ params:
   bandwidth: 50
   thresholds: 10
 plot:
```
domoritz, line 6: Isn't this generated? Then we should not reformat.

package.json
```diff
@@ -19,6 +19,9 @@
     "clean": "lerna run clean",
     "lint": "lerna run lint",
     "test": "tsc --build && lerna run schema && vitest run",
+    "format": "prettier --write '**/*.{js,ts,tsx,json,css,md}'",
+    "lint:fix": "eslint '**/*.{js,ts,tsx}' --fix",
+    "format:check": "prettier --check '**/*.{js,ts,tsx,json,css,md}'",
```
domoritz: That should be in lint already.

package.json
```diff
@@ -19,6 +19,9 @@
     "clean": "lerna run clean",
     "lint": "lerna run lint",
     "test": "tsc --build && lerna run schema && vitest run",
+    "format": "prettier --write '**/*.{js,ts,tsx,json,css,md}'",
```
domoritz: Lint fix and format should be the same

packages/examples/vanilla-example/src/main.js
```diff
@@ -5,17 +5,10 @@ import * as vg from '@uwdata/vgplot';
...
-      {x: "Date", y: "Close", tip: true}
-    ),
-    vg.width(680),
-    vg.height(200)
-  )
-)
+document
+  .getElementById('plot')
+  .replaceChildren(
+    vg.plot(vg.lineY(vg.from('aapl'), { x: 'Date', y: 'Close', tip: true }), vg.width(680), vg.height(200))
```
domoritz: This is worse in terms of readability. Any ideas how we can improve?

packages/examples/vega-example/src/main.js
```diff
@@ -10,48 +10,48 @@ const result = await embed(vegaPlot, spec, { actions: false });
...
-        ),
-        vg.intervalX({ as: selection }),
-        vg.xyDomain(vg.Fixed),
-        vg.width(350),
-        vg.height(240)
+document
+  .querySelector('#plots')
+  .replaceChildren(
+    vg.vconcat(
+      vg.hconcat(vg.hspace('2em'), vg.menu({ from: 'weather', column: 'weather', label: 'Weather', as: selection })),
```
domoritz: This also looks worse. But maybe okay as a trade off.

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -96,10 +88,10 @@ export class Coordinator {
     const { clients = true, cache = true } = options;
     this.manager.clear();
     if (clients) {
-      this.filterGroups?.forEach(group => group.disconnect());
-      this.filterGroups = new Map;
-      this.clients?.forEach(client => this.disconnect(client));
-      this.clients = new Set;
+      this.filterGroups?.forEach((group) => group.disconnect());
```
domoritz: Let's remove the extra (). Iirc prettier has an option for this.

packages/mosaic/core/src/MosaicClient.ts
```diff
@@ -154,7 +155,8 @@ export class MosaicClient {
    * @param error
    * @returns this
    */
-  queryError(error: Error): this { // eslint-disable-line @typescript-eslint/no-unused-vars
+  queryError(error: Error): this {
```
domoritz: Why is this in a new line?

packages/mosaic/core/src/QueryConsolidator.ts
```diff
@@ -103,8 +102,11 @@ function consolidationKey(query: QueryType, cache: Cache): string {
   const sql = `${query}`;
   if (isSelectQuery(query) && !cache.get(sql)) {
     if (
-      query._where.length || query._qualify.length || query._having.length ||
```
domoritz, line 106: Was there a reason for the grouping? Maybe use comments.

packages/mosaic/core/src/preagg/sufficient-statistics.ts
```diff
@@ -410,12 +429,7 @@ function regrSumXYExpr(
   avg: (field: ColumnRefNode) => ExprNode
 ): ExprNode {
   const [y, x] = node.args;
-  const sxy = sum(
-    mul(
-      sub(x, avg(x as ColumnRefNode)),
-      sub(y, avg(y as ColumnRefNode))
-    )
-  );
+  const sxy = sum(mul(sub(x, avg(x as ColumnRefNode)), sub(y, avg(y as ColumnRefNode))));
```
domoritz: Harder to read

package.json
theVedanta: would having prettier-format not address more of files that eslint cannot go over though?

docs/public/contours.yaml
theVedanta: what other files would this go for? just making sure I know what else is generated here

packages/examples/vanilla-example/src/main.js
theVedanta: umm.. I would say
1. printWidth should be shorter like 80 or so (I tested 80 and works okay for me)
2. es5 for the trailing commas

Doing that now...

docs/public/contours.yaml
domoritz: I think the yaml and esm example are all generated. Try building the library and website without formatting and see whether anything gets reset.

package.json
domoritz: Maybe. But then we should still have them in one command. I am also less worried about files other than ts.

packages/examples/vanilla-example/src/main.js
domoritz: Let's try different line widths and see what gives us good results and fewer changes. 

Why would trailing commas help?

domoritz: Maybe 100 is good

theVedanta: Yeah 100 works well for me too :)
We dont wanna use `"all"` for it since trailing commas are not seen in the project, so to keep things close to the previous version and reduce changes, I thought `"es5"` would be a good idea

packages/mosaic/core/src/Coordinator.ts
theVedanta: Done! For keeping this here, it was `"arrowParens": "avoid"`

package.json
theVedanta: what should be in lint? Sorry I don't understand?

package.json
theVedanta: That makes sense. would you like me to make a combined command for them or just get rid of the `prettier --write` entirely?

packages/mosaic/core/src/MosaicClient.ts
theVedanta: good catch! This is interesting because the printwidth will wrap it to the next line. We have the option to:
1. Move all these comments above the functions manually with `eslint-disable-next-line` (time-taking)
2. Rename all unused args to start with `_` perhaps? and then ignore them in global eslint config with something like `"@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]`

package.json
domoritz: Lint should be a superset of format check

package.json
domoritz: How much do we benefit from combined vs remove?

packages/mosaic/core/src/MosaicClient.ts
domoritz: 1 seems cleaner. Even better might be to disable the check and use typescript with @ts-expect-error. The advantage of that is that this will error when the variable is actually used.

.prettierrc.json
```diff
@@ -1,7 +1,7 @@
 {
   "singleQuote": true,
   "tabWidth": 2,
-  "trailingComma": "none",
-  "printWidth": 120,
```
domoritz: I think I prefer 100.

theVedanta: Gotcha, will set it to 100

packages/mosaic/core/src/MosaicClient.ts
theVedanta: Pardon me, but why would you say that is an advantage?

package.json
theVedanta: How about we do:
```json
    "lint:fix": "eslint '**/*.{js,ts,tsx}' --fix && prettier --write '**/*.{js,ts,tsx,json,css,md}'",
```

and then have a `format:write` command separately for the `--write`? Best of both worlds imo

theVedanta: I don't suppose it's any harm removing the `format` command though; if we dont care about the files other than ts

packages/mosaic/core/src/MosaicClient.ts
domoritz: ts expect error will silence an error but also will error if there is no error

package.json
theVedanta: gotcha, removing.

bin/publish-schema.js
```diff
@@ -6,7 +6,8 @@ import { copyFile, readFile, symlink, unlink } from 'node:fs/promises';
 async function link(schemaFile, linkFile) {
   try {
     await unlink(linkFile);
-  } catch (err) { // eslint-disable-line @typescript-eslint/no-unused-vars
+    // eslint-disable-next-line
```
domoritz: I don't think this is needed anymore.

packages/mosaic/core/src/util/AsyncDispatch.ts
```diff
@@ -83,8 +83,10 @@ export class AsyncDispatch<T> {
    *  function, or null if all unemitted event values should be filtered.
    */
   emitQueueFilter(
-    _type: string, // eslint-disable-line @typescript-eslint/no-unused-vars
-    _value: unknown // eslint-disable-line @typescript-eslint/no-unused-vars
+    // eslint-disable-next-line
+    _type: string, // @typescript-eslint/no-unused-vars
```
domoritz: We can handle _ automatically. Please read https://typescript-eslint.io/rules/no-unused-vars/

packages/vgplot/plot/src/interactors/util/get-datum.js
```diff
@@ -9,7 +9,7 @@ export function getDatum(el) {
   if (el.tagName === 'a') {
     el = el.children[0];
   }
-  // @ts-ignore
+  // @ts-expect-error
```
domoritz: Let's include a description of what error is expected here.

packages/vgplot/plot/src/interactors/util/get-field.js
```diff
@@ -3,10 +3,10 @@ import { isNode } from '@uwdata/mosaic-sql';
 function extractField(field) {
   if (isNode(field)) {
     if (field.type === 'COLUMN_REF') {
-      // @ts-ignore
+      // @ts-expect-error
```
domoritz: Same for this and all others.

packages/vgplot/plot/src/marks/util/extent.js
```diff
@@ -40,7 +40,6 @@ export function filteredExtent(filter, column) {
 
   [filter].flat().forEach(p => walk(p, (node) => {
     if (node instanceof BetweenOpNode && `${node.expr}` === column) {
-      // @ts-ignore
```
domoritz, line 44: Nice

packages/vgplot/spec/src/ast/ASTNode.js
```diff
@@ -24,16 +24,16 @@ export class ASTNode {
    * @param {import('../ast-to-esm.js').CodegenContext} ctx The code generator context.
    * @returns {string|void} The generated ESM code for the node.
    */
-  codegen(ctx) { // eslint-disable-line no-unused-vars
-    // @ts-ignore
+   // eslint-disable-next-line
+   codegen(ctx) { // no-unused-vars
```
domoritz: Are we not removing no-unused-vars?

bin/publish-schema.js
domoritz: Update the rest as well. We should probably have something like

```
export default tseslint.config({
  rules: {
    // Note: you must disable the base rule as it can report incorrect errors
    "no-unused-vars": "off",
    "@typescript-eslint/no-unused-vars": "error"
  }
});
```

packages/vgplot/spec/src/ast/ASTNode.js
```diff
@@ -14,8 +14,8 @@ export class ASTNode {
    * @param {import('../ast-to-dom.js').InstantiateContext} ctx The instantiation context.
    * @returns {*} The instantiated value of this node.
    */
-  instantiate(ctx) { // eslint-disable-line no-unused-vars
-    // @ts-ignore
+   // eslint-disable-next-line
+   instantiate(ctx) { // no-unused-vars
```
domoritz: Are we not removing no-unused-vars?

package.json
```diff
@@ -19,6 +19,8 @@
...
+    "lint:fix": "eslint '**/*.{js,ts,tsx}' --fix",
```
domoritz: Let's combine lint fix and format. They are the same to me.

domoritz:
Oh okay. I want to see the changes before I approve. 

I couldn't run the code locally since the installation was broken. 

I can check locally when the installation works.

packages/mosaic/core/src/util/AsyncDispatch.ts
theVedanta: Done!

bin/publish-schema.js
theVedanta: understood.

packages/vgplot/plot/src/interactors/util/get-field.js
theVedanta: To be honest, I don't really have context on why all of these produce the errors since I did not write that code, so please feel free to correct me \:)

packages/vgplot/spec/src/ast/ASTNode.js
theVedanta: missed a few, removing...

packages/mosaic/core/src/util/AsyncDispatch.ts
```diff
@@ -67,7 +67,7 @@ export class AsyncDispatch<T> {
    * @param value The event value.
    * @returns The (possibly transformed) event value to emit.
    */
-  willEmit(type: string, value: T): T {
+  willEmit(_type: string, value: T): T {
```
domoritz, line 70: Let's not use _ for unused variables. In this case we missed that the documentation actually uses the name `type` rather than `_type`. Keeping the names and docs consistent if you all off a sudden need the variable in the function is too much overhead.

packages/vgplot/plot/src/interactors/util/get-datum.js
```diff
@@ -9,7 +9,7 @@ export function getDatum(el) {
   if (el.tagName === 'a') {
     el = el.children[0];
   }
-  // @ts-ignore
+  // @ts-expect-error -- __data__ is a non-standard property set by the rendering runtime
```
domoritz, line 12: Great!

packages/mosaic/core/src/util/AsyncDispatch.ts
theVedanta: would it be worth it to update the docs? or should we just revert to having the unused variables with a `disable` line?

domoritz: We should revert. Updating the docs isn't a good idea since we do use the parameter name in some cases.

packages/vgplot/plot/src/interactors/util/get-field.js
domoritz: I see. We can handle this separately then. Can we provide the error that happens so that eslint can automatically fail if the specific error goes away?

packages/vgplot/spec/src/ast/ASTNode.js
domoritz: Thanks

packages/mosaic/core/src/util/AsyncDispatch.ts
theVedanta: Ok sounds good. I will go back and make changes to my commit. But just to refresh your memory about this conversation: https://github.com/uwdata/mosaic/pull/949#discussion_r2764377173, we did finalize using this approach so I'm just making sure if I should go ahead. Thank you for your help \:)

domoritz (changes requested):
Please merge main and resolve the merge conflicts.

packages/mosaic/core/src/util/AsyncDispatch.ts
domoritz: What I said in that comment is that we don't need to annotate variables that already have an _. I didn't say that we should switch to using _ as a prefix for unused variables everywhere.

PR #950, opened 2025-12-25
feat: Devtools - A separate, lightweight event bus -- utilizing Dispatch (Sync)

theVedanta:
This commit resolves some tasks discussed in the meeting with @domoritz. In response to the issue https://github.com/uwdata/mosaic/issues/943
For an overview, these are some of the immediate action items discussed:

## Action items

- [x] Eliminate existing logger implementation
- [x] Add a `.observe` method to a single place on the coordinator  
  - [x] Params: `observer`, `Set` of event types (not provided subscribes to all events)
- [x] Initial implementation only includes bare minimum events to replicate current logging experience
- [x] Put code into coordinator for now, may split it into a separate file later
- [x] Remove `logQueries` bool from coordinator
- [x] Add abstract class `Dispatch`  
  - [x] existing `AsyncDispatch` extends it  
  - [x] Add new `ObserveDispatch` also extends it  
  - [x] `.observe` routes to observe dispatch `.addEventListener`

The UI-related example, written in Solid JS, will be in a separate PR. The current goal is to switch out the logger in Coordinator and QueryManager to consume our event bus for uncertain event types (which will be crystallized later for devtools).

https://docs.google.com/document/d/1cyw2sv9G5-CzGG-JgeOEcISSu7zmEqdzsa-J2oh2F6c/edit?tab=t.0


dev/index.html
```diff
@@ -154,7 +154,8 @@
   }
 
   function setQueryLog() {
-    vg.coordinator().manager.logQueries(qlogToggle.checked);
+    // logQueries is no longer in use
+    // vg.coordinator().manager.logQueries(qlogToggle.checked);
```
domoritz: This is where we should add or remove our custom observer that redirects events to console.

packages/mosaic/core/src/util/AsyncDispatch.ts
```diff
@@ -91,11 +83,69 @@ export class AsyncDispatch<T> {
...
+  }
+
+  /**
+   * Remove an event listener callback for the provided event type.
+   * @param type The event type.
+   * @param callback The event handler
+   *  callback function to remove.
+   */
+  override removeEventListener(type: string, callback: EventCallback<T>): void {
+    super.removeEventListener(type, callback);
```
domoritz: do we need this override? Looks equivalent

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -126,6 +130,11 @@ export class Coordinator {
     if (arguments.length) {
       this._logger = logger || voidLogger();
       this.manager.logger(this._logger);
+      // subscribe logger to events
```
domoritz: Looks good as a stepping stone but I think we can go ahead and make the logger a separate observer that users need to explicitly add. We can remove the logger entirely now imo.

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -246,12 +257,12 @@ export class Coordinator {
     priority: number = Priority.Normal
   ): Promise<unknown> {
     client.queryPending();
-    return client._pending = this.query(query, { priority })
+    return client._pending = this.query(query, { priority, clientId: (client as any).id })
```
domoritz: as any is worrying me a bit here

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -246,12 +257,12 @@ export class Coordinator {
...
       .then(
         data => client.queryResult(data).update(),
-        err => { this._logger?.error(err); client.queryError(err); }
+        err => { this.eventBus.emit(EventType.Error, { message: err, timestamp: Date.now() }); client.queryError(err); }
```
domoritz: I wonder whether we really need timestamps since the event bus could maybe automatically add it for all emitted events? Could make the callsites simpler.

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -280,6 +291,17 @@ export class Coordinator {
       throw new Error('Client already connected.');
     }
 
+    // assign client id if not present
```
domoritz: Not a fan of this. What is this needed? Can we add this later?

packages/mosaic/core/src/EventBus.ts
```diff
@@ -0,0 +1,49 @@
...
+  ClientConnect = 'client-connect',
+  ClientStateChange = 'client-state-change',
+  Error = 'error'
+}
+
+export interface QueryStartEvent {
+  query: string;
+  materialized: boolean;
+  clientId?: string;
+  timestamp: number;
```
domoritz: All events have a timestamp so maybe we should extract the type and logic.

packages/mosaic/core/src/EventBus.ts
```diff
@@ -0,0 +1,49 @@
...
+  clientId: string;
+  timestamp: number;
+}
+
+export interface ErrorEvent {
+  message: unknown;
+  timestamp: number;
+}
+
+export class EventBus {
```
domoritz: Why do we need this class? It seems to just be dispatching events to the ObservableDispatch anyway. If we make the latter generic, we may not need this class at all.

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -77,23 +78,43 @@ export class QueryManager {
...
+          timestamp: Date.now()
+        });
+      }
+
       // check query cache
       if (cache) {
         const cached = this.clientCache!.get(sql!);
         if (cached) {
           const data = await cached;
           this._logger.debug('Cache');
```
domoritz: Let's remove the logger and replace the same location with the event bus stuff.

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -77,23 +78,43 @@ export class QueryManager {
...
           result.ready(data);
+          // emit QueryEnd for cached
+          if (this.eventBus) {
```
domoritz: It's odd to me that this is after `result.ready(data)` when previously the logging was happening before.

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -107,7 +128,22 @@ export class QueryManager {
 
       this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
       result.ready(type === 'exec' ? null : data);
+
+      if (this.eventBus) {
```
domoritz: same as above

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -107,7 +128,22 @@ export class QueryManager {
...
+        this.eventBus.emit(EventType.QueryEnd, {
+          query: sql || '',
+          materialized: cache,
+          clientId,
+          timestamp: Date.now()
+        });
+      }
     } catch (err) {
+      if (this.eventBus) {
```
domoritz: Is this needed to replicate the logic of the logger? If not, let's not add it yet.

domoritz (commented):
Very cool start. looking good. Here are some comments.

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -107,7 +128,22 @@ export class QueryManager {
 
       this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
       result.ready(type === 'exec' ? null : data);
+
+      if (this.eventBus) {
+        this.eventBus.emit(EventType.QueryEnd, {
+          query: sql || '',
```
domoritz: If we have an undefined sql, we can just output that.

packages/mosaic/core/src/EventBus.ts
theVedanta: I wondered about just using ObservableDispatch too, but I feel that a separate EventBus would actually be extensible in the future for various reasons like maintaining a history of events between renders (for frameworks like react/preact/solid/etc). That said, if you would still like for me to just use ObserveDispatch directly, to keep things light, we can definitely do that

packages/mosaic/core/src/Coordinator.ts
theVedanta: absolutely, lemme go ahead and do that

packages/mosaic/core/src/Coordinator.ts
theVedanta: understandable, this is only for giving IDs to clients (I did not find a native client-id assignment in the project (let me know if there is one! :) )

packages/mosaic/core/src/Coordinator.ts
theVedanta: I'm thinking we get rid of the clientId for now?

packages/mosaic/core/src/util/AsyncDispatch.ts
theVedanta: I think I had some extensibility in mind for later, but on at another glance, I think this feels unnecessary. Removing...

packages/mosaic/core/src/EventBus.ts
theVedanta: Makes sense. Done. Are there any other fields we wanna abstract?

packages/mosaic/core/src/EventBus.ts
domoritz: Let's keep things light.

packages/mosaic/core/src/Coordinator.ts
domoritz: Let's do that

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -51,7 +52,7 @@ export class Coordinator {
   public clients = new Set<MosaicClient>;
   public filterGroups = new Map<Selection, FilterGroupEntry>;
   protected _logger: Logger = voidLogger();
-  public eventBus: EventBus;
+  public eventBus: ObserveDispatch<any>; // temporary measures until we finalize event types
```
domoritz: Let's use the types we already have. We can expand it later.

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,30 @@
...
+  QueryStart = 'query-start',
+  QueryEnd = 'query-end',
+  ClientConnect = 'client-connect',
+  ClientStateChange = 'client-state-change',
+  Error = 'error'
+}
+
+export interface MosaicEvent {
+  timestamp: number;
+  // Extend later with more fields
```
domoritz: No need to say that here.

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,30 @@
...
+  query: string;
+  materialized: boolean;
+}
+
+export interface ClientConnectEvent extends MosaicEvent {
+  clientId?: string;
+}
+
+export interface ErrorEvent extends MosaicEvent {
+  message: unknown;
```
domoritz: Should it be unknown or string?

packages/mosaic/core/src/util/ObserveDispatch.ts
```diff
@@ -0,0 +1,54 @@
+import { Dispatch, EventCallback } from "./AsyncDispatch.js";
+
+/**
+ * Synchronous event dispatcher that pushes events directly to callbacks
+ * without queuing or Promise handling. Suitable for simple observer patterns.
+ */
+export class ObserveDispatch<T> extends Dispatch<T> {
+  constructor() {
+    super();
+  }
```
domoritz: Can remove since it's a noop

packages/mosaic/core/src/Coordinator.ts
```diff
@@ -280,6 +266,11 @@ export class Coordinator {
       throw new Error('Client already connected.');
     }
 
+    // emit ClientConnect
+    this.eventBus.emit(EventType.ClientConnect, {
+      // additional arguments for later -- for comprehensive client data
```
domoritz: remove if not yet used

.prettierrc.json
domoritz, line 1: let's do in the other pull request

theVedanta: oh sorry i was just using this locally, my bad

domoritz: All good. Let me know when you addressed the comments above. I think we are getting close.

packages/mosaic/core/src/Events.ts
theVedanta: the extend later part?

packages/mosaic/core/src/Events.ts
theVedanta: string for sure; I don't remember why I typed this as unknown

packages/mosaic/core/src/Coordinator.ts
theVedanta: gotcha, we'll add identifiers to the clients later

.prettierrc.json
theVedanta: Done \:) Anything else I may have missed?

packages/mosaic/core/src/Events.ts
domoritz: Yeah, the whole comment.

packages/mosaic/core/src/Coordinator.ts
domoritz: My question is, can we just not have no ids at all.

dev/query/index.html
```diff
@@ -57,7 +57,7 @@
       await coordinator.exec(`${input.value}`);
       output.replaceChildren();
     } else {
-      await coordinator().exec(
+      await vg.coordinator().exec(
```
domoritz, line 67: why do we need to add `vg.`?

dev/query/index.html
```diff
@@ -66,6 +66,21 @@
     }
     enable();
   });
+
+  function setQueryLog() {
```
domoritz, line 77: Shouldn't this logic (which replicates the old logger, right?) be in mosaic core so that people can easily get the old logger behavior again?

dev/index.html
```diff
@@ -154,7 +154,16 @@
   }
 
   function setQueryLog() {
-    vg.coordinator().manager.logQueries(qlogToggle.checked);
+    const bus = vg.coordinator().eventBus;
+    
+    bus.observe("QueryStart", (event) =>
```
domoritz: duplicated logic?

packages/mosaic/core/src/util/AsyncDispatch.ts
```diff
@@ -131,7 +178,7 @@ export class AsyncDispatch<T> {
       if (callbacks?.size) {
         // broadcast update to callbacks, which may return promises
         // wait until promises resolve, then process pending updates
-        const callbackValues = Array.from(callbacks, cb => cb(event));
+        const callbackValues = Array.from(callbacks, (cb) => cb(event));
```
domoritz: why change this here?

packages/mosaic/core/src/util/AsyncDispatch.ts
```diff
@@ -73,14 +65,14 @@ export class AsyncDispatch<T> {
 
   /**
    * Lifecycle method that returns a filter function for updating the
-   * queue of unemitted event values prior to enqueueing a new value.
+   * queue of un-emitted event values prior to enqueueing a new value.
```
domoritz: let'd not change spelling here

packages/mosaic/core/src/util/ObserveDispatch.ts
```diff
@@ -0,0 +1,50 @@
+import { Dispatch, EventCallback } from "./AsyncDispatch.js";
+
+/**
+ * Synchronous event dispatcher that pushes events directly to callbacks
+ * without queuing or Promise handling. Suitable for simple observer patterns.
+ */
+export class ObserveDispatch<T> extends Dispatch<T> {
+  /**
+   * Subscribe to events of the given type. Essentially
+   * a shorthand/wrapper for addEventListener().
```
domoritz: I don't see how it's different. If it's the same, can we just use addEventListener?

packages/mosaic/core/src/util/ObserveDispatch.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @returns This ObserveDispatch instance for method chaining.
+   */
+  observe(type: string, callback: EventCallback<T>): this {
+    this.addEventListener(type, callback);
+    return this;
+  }
+
+  /**
+   * Unsubscribe from events of the given type. Essentially
+   * a shorthand/wrapper for removeEventListener().
```
domoritz: same as above. The code looks identical.

packages/mosaic/core/src/util/ObserveDispatch.ts
```diff
@@ -0,0 +1,50 @@
...
+   */
+  unobserve(type: string, callback: EventCallback<T>): this {
+    this.removeEventListener(type, callback);
+    return this;
+  }
+
+  /**
+   * Emit an event value to listeners for the given event type.
+   * Events are dispatched *synchronously* to all registered callbacks.
+   * This is different from the AsyncDispatch class, which use queues to
```
domoritz: I don't get why we distinguish from async so much. Async is just a different class that inherits from the same base. Sync is more basic so it feels odd to distinguish so strongly. Let's simplify the comments.

packages/mosaic/core/src/util/ObserveDispatch.ts
```diff
@@ -0,0 +1,50 @@
+import { Dispatch, EventCallback } from "./AsyncDispatch.js";
+
+/**
+ * Synchronous event dispatcher that pushes events directly to callbacks
+ * without queuing or Promise handling. Suitable for simple observer patterns.
```
domoritz: See below's comments and comments.

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,43 @@
...
+export type EventMap = {
+  [EventType.QueryStart]: Omit<QueryStartEvent, keyof MosaicEvent>;
+  [EventType.QueryEnd]: Omit<QueryEndEvent, keyof MosaicEvent>;
+  [EventType.ClientConnect]: Omit<ClientConnectEvent, keyof MosaicEvent>;
+  [EventType.Error]: Omit<ErrorEvent, keyof MosaicEvent>;
+};
+
+export interface MosaicEvent {
+  timestamp: number;
+  // Extend later with more fields
```
domoritz: rm comment

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export type EventMap = {
+  [EventType.QueryStart]: Omit<QueryStartEvent, keyof MosaicEvent>;
+  [EventType.QueryEnd]: Omit<QueryEndEvent, keyof MosaicEvent>;
+  [EventType.ClientConnect]: Omit<ClientConnectEvent, keyof MosaicEvent>;
+  [EventType.Error]: Omit<ErrorEvent, keyof MosaicEvent>;
+};
+
+export interface MosaicEvent {
```
domoritz: Should this be private/not exported?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -1,35 +1,34 @@
...
-  private _logQueries: boolean;
   private _consolidate: ReturnType<typeof consolidator> | null;
   /** Requests pending with the query manager. */
   public pendingResults: QueryResult[];
   private maxConcurrentRequests: number;
   private pendingExec: boolean;
+  public eventBus?;
 
-  constructor(maxConcurrentRequests: number = 32) {
+  constructor(maxConcurrentRequests: number = 32, eventBus?: ObserveDispatch<Omit<MosaicEvents, keyof MosaicEvent>>) {
```
domoritz: why do we need the omit?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -52,7 +51,7 @@ export class QueryManager {
         if (result.state === QueryState.ready) {
           result.fulfill();
         } else if (result.state === QueryState.done) {
-          this._logger.warn('Found resolved query in pending results.');
+          this.eventBus?.emit(EventType.Error, { message: 'Found resolved query in pending results.' });
```
domoritz: Shouldn't this be a warning?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -80,22 +79,27 @@ export class QueryManager {
...
+        materialized: cache,
+      });
+
       // check query cache
       if (cache) {
         const cached = this.clientCache!.get(sql!);
         if (cached) {
-          const data = await cached;
-          this._logger.debug('Cache');
+          const data = cached;
```
domoritz, line 117: Do we not need to await anymore?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -105,9 +109,16 @@ export class QueryManager {
...
       result.ready(type === 'exec' ? null : data);
+
+      this.eventBus?.emit(EventType.QueryEnd, {
+        query: sql || '',
+        materialized: cache,
+      });
     } catch (err) {
+      this.eventBus?.emit(EventType.Error, {
+        message: err,
+      });
```
domoritz: Is this strictly needed to replicate the behavior of the old logger?

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,43 @@
+export enum EventType {
+  QueryStart = "query-start",
+  QueryEnd = "query-end",
+  ClientConnect = "client-connect",
+  ClientStateChange = "client-state-change",
+  Error = "error",
+}
+
+export type EventMap = {
```
domoritz: I've never seen this pattern before. Why is this needed?

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,43 @@
+export enum EventType {
+  QueryStart = "query-start",
+  QueryEnd = "query-end",
+  ClientConnect = "client-connect",
+  ClientStateChange = "client-state-change",
```
domoritz: Should this still be here?

packages/mosaic/core/src/Events.ts
```diff
@@ -0,0 +1,43 @@
+export enum EventType {
```
domoritz, line 1: Instead of having this enum here. Wouldn't it be easier to have classes for the events that can be instantiated with a constructor?

domoritz (changes requested):

packages/mosaic/core/src/Events.ts
theVedanta: This was mainly for type-safety (for us and the clients)

packages/mosaic/core/src/Events.ts
theVedanta: That is probably true. It will get rid of this "broken"-ish type interface for the events. But I remember that we agreed on enums (denoted by strings instead of numbers) to provide the user a simple list of event types in autocomplete

packages/mosaic/core/src/QueryManager.ts
theVedanta: await will have no effect here

packages/mosaic/core/src/QueryManager.ts
theVedanta: Well, all it's doing it just pipe whatever the message was (which logger emitted) into the bus, which is now consumable by anyone, including the logger. I don't know if Im missing something, did you have a simpler approach in mind?

packages/mosaic/core/src/QueryManager.ts
theVedanta: yep, adding a warning event type now... (perhaps we need more for other logger-related types -- warning, info, etc)

packages/mosaic/core/src/QueryManager.ts
theVedanta: I chose this approach because we dont pass timestamp here at compile time but we need to provide the user with the timestamp type. Open to hear ideas that can do this better!

packages/mosaic/core/src/Events.ts
theVedanta: I feel like we'll need this type when omitting pre-set keys in the coordinator/query-manager

packages/mosaic/core/src/util/AsyncDispatch.ts
theVedanta: sorry, might have been a prettier thing on my local machine

dev/query/index.html
theVedanta: honestly, I'm not really sure. this example was broken when I tested it; the coordinator is not accessible as an object here, but only as a property of `vg`

packages/mosaic/core/src/util/ObserveDispatch.ts
theVedanta: We can, but in the meeting I remember us discussing that we need terminology to differ in the event bus. The listener function should be called `.observe` based on our discussion. But I'm open to switching back to `addEventListener` for consistency

dev/query/index.html
theVedanta: ahh. we can definitely do that. My mistake, I felt that when we said 'remove the logger from core entirely,' we give the client the option to write that logic. I can put this function in core to give back a logger util to the user/client (perhaps we can discuss that idea some more (?))

packages/mosaic/core/src/Events.ts
theVedanta: sorry? I don't understand

dev/query/index.html
domoritz: Ahh, good catch. I'll fix it separately. Let's keep PRs focused on one thing.

domoritz: Done in https://github.com/uwdata/mosaic/pull/965. Let's remove this here.

dev/query/index.html
domoritz: IIRC the idea was to remove the logger but give users an equivalent tool via the new bus.

packages/mosaic/core/src/QueryManager.ts
domoritz: Let's make a separate pull request. IIRC there was some reason where cached results can be promises but I am not 100% sure right now.

packages/mosaic/core/src/Events.ts
domoritz: Let's not do this and instead use the class approach I suggested?

packages/mosaic/core/src/Events.ts
domoritz: I don't see `client-state-change` used anywhere.

packages/mosaic/core/src/Events.ts
domoritz: Yeah, we can still use enums in the event types but in order to create an event object, we could use a class. Maybe it's cleaner to actually put the event type into the message so that we can easily type check a message in isolation. Let's do that.

packages/mosaic/core/src/QueryManager.ts
domoritz: I just didn't see a log message (we removed) here before so I am wondering whether the emit is needed.

dev/query/index.html
theVedanta: Awesome!

packages/mosaic/core/src/util/ObserveDispatch.ts
theVedanta: Sure!

domoritz:
Closing in favor of #1028

PR #965, opened 2026-02-04
fix: correct query playground

domoritz:


domoritz:
@jheer I think this is needed to fix the query playground which otherwise throws an error.

PR #966, opened 2026-02-07
chore: replace npm with pnpm

theVedanta:
Simply replaces npm (package-lock) with pnpm (pnpm-lock) and adds a workspace.yaml file

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,109 +9,112 @@ importers:
   .:
     devDependencies:
       '@eslint/js':
-        specifier: ^9.39.2
-        version: 9.39.2
+        specifier: ^10.0.1
+        version: 10.0.1(eslint@10.1.0)
       ajv:
-        specifier: ^8.17.1
-        version: 8.17.1
+        specifier: ^8.18.0
+        version: 8.18.0
       ajv-formats:
         specifier: ^3.0.1
-        version: 3.0.1(ajv@8.17.1)
+        version: 3.0.1(ajv@8.18.0)
       eslint:
-        specifier: ^9.39.2
-        version: 9.39.2
+        specifier: ^10.1.0
+        version: 10.1.0
+      globals:
+        specifier: ^17.4.0
+        version: 17.4.0
       jsdom:
-        specifier: ^27.4.0
-        version: 27.4.0
+        specifier: ^29.0.1
+        version: 29.0.1
       lerna:
-        specifier: ^9.0.3
-        version: 9.0.3(@types/node@20.19.32)
+        specifier: ^9.0.7
+        version: 9.0.7(@types/node@20.19.32)
       nodemon:
-        specifier: ^3.1.11
-        version: 3.1.11
+        specifier: ^3.1.14
+        version: 3.1.14
       rimraf:
-        specifier: ^6.1.2
-        version: 6.1.2
+        specifier: ^6.1.3
+        version: 6.1.3
       timezone-mock:
-        specifier: ^1.3.6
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,11 +44,5 @@
...
     "vitest": "^4.0.17",
     "yaml": "^2.8.2"
-  },
-  "workspaces": [
-    "packages/mosaic/*",
-    "packages/server/*",
-    "packages/vgplot/*",
-    "packages/examples/*"
-  ]
+  }
```
domoritz: Do we need to add something here to tell people to use pnpm?

theVedanta: you're right, let's add a packageManager field

theVedanta: added 10.5.2 as the pnpm version to use

theVedanta: also please set pnpm in the CI with a `--frozen-lockfile` flag \:)

domoritz: Can you do that?

theVedanta: of course

package.json
```diff
@@ -9,7 +9,9 @@
     "type": "git",
     "url": "https://github.com/uwdata/mosaic.git"
   },
+  "packageManager": "pnpm@10.5.2",
```
domoritz: I think we should allow more than one specific version.

.github/workflows/test.yml
```diff
@@ -20,18 +20,25 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: "10.5.2"
```
domoritz: This is already outdated. Let's use `latest` here. We will notice early if things break then.

theVedanta: I don't think we should use latest here, since this will unexpectedly result in lockfile changes (format or otherwise) and install behaviors. I would prefer we set both to a stable version and upgrade both manually when needed (if at all)

derekperkins: dependabot should pick it up in monthly runs?

theVedanta: Yes, though I've never set up dependabot or renovate before, might have to mess around a bit

derekperkins: It's already setup and running, so I don't think you have to do anything. I agree with pinning a specific version vs using latest

https://github.com/uwdata/mosaic/blob/09334ee4a8f8b4b5a1b100e8468b8069db0feca4/.github/dependabot.yml#L3

domoritz: https://github.com/uwdata/mosaic/blob/09334ee4a8f8b4b5a1b100e8468b8069db0feca4/.github/dependabot.yml#L3 updates actions but not sure it updates the part inside "with".

derekperkins: I don't know either, that was the reason behind my question mark earlier. I would hope so?

Looking at it, I think we just leave this line out and it'll pull the version from our package.json
https://github.com/pnpm/action-setup?tab=readme-ov-file#version

theVedanta: yep, sounds good. I would love to get this PR in since it is a pain right now to install dependencies and test locally. If there's any problems with pinning down a pnpm version, we can always make another PR @domoritz

package.json
domoritz: Do we need to replace npm run?

.github/workflows/test.yml
domoritz: Sounds good. I found a few more things that need to be fixed (missing a lot of places where we still use npm) and the ci needs to pass.

PR #967, opened 2026-02-09
mosaic-core: latest-only stream scheduling for interactive updates

Datta-sai-vvn:
Problem
Interactive UI interactions (brush/drag/slider) can enqueue many similar queries quickly. With Mosaic’s current request queueing and socket protocol, those requests are executed serially and cannot be cancelled server-side. This causes “stale query storms” where the system spends time executing queries that are already obsolete, increasing end-to-end latency for the user’s latest intent.

Change
Add latest-only stream scheduling at the Coordinator/QueryManager layer:

New request options: stream?: string, latest?: boolean

Semantics when stream + latest are set:

Increment per-stream generation counter

Prune older queued requests for the same stream before they reach the connector

Suppress stale results using generation IDs when responses arrive

What this is NOT

This does not cancel queries already executing in DuckDB/server (protocol + node-duckdb binding do not expose interruption). It reduces wasted work by preventing stale requests from being sent when still queued and by ignoring stale responses.

Evidence

Reproduced backlog: blasting 30 heavy queries yields ~7.6s tail latency due to serial execution.

With latest-only enabled, only the latest request resolves; stale requests are pruned/suppressed, reducing tail latency substantially.

Added a unit test proving pruning behavior and small number of connector calls.

Files

packages/mosaic/core/src/types.ts

packages/mosaic/core/src/Coordinator.ts

packages/mosaic/core/src/QueryManager.ts

packages/mosaic/core/test/querymanager-latest-only.test.ts

packages/mosaic/core/docs/latest-only-streams.md

How to use
Use stream to group interaction requests (e.g., "brush", "pan", "slider"). Set latest: true to keep only the most recent pending request for that stream.

### DIFF
--- a/packages/mosaic/core/docs/latest-only-streams.md
+++ b/packages/mosaic/core/docs/latest-only-streams.md
@@ -1,0 +1,24 @@
+# Latest-Only Stream Scheduling
+
+## Problem
+Interactive applications often generate high-frequency query streams (e.g., brushing, slider dragging). Sending every intermediate query to the server can cause:
+1.  **Backlog Latency:** The server queues up stale requests, delaying the "latest" relevant result.
+2.  **Wasted Work:** Computing results for chart states that the user has already passed.
+
+## API
+The `Coordinator.query()` method accepts two new options:
+
+```typescript
+coord.query(sql, {
+  stream: 'my-interaction-id', // Group requests by this ID
+  latest: true                 // Enable pruning of stale requests
+});
+```
+
+## Semantics
+When `latest: true` is set:
+1.  **Queue Pruning:** Any *pending* (queued but not yet sent) requests with the same `stream` ID are removed from the client-side queue.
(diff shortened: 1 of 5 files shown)


Datta-sai-vvn (commented):
Hi! This PR is ready for review. Since it’s from a fork, GitHub Actions is currently blocked (“workflow awaiting approval”).
Could a maintainer please approve the workflow run so CI can execute, and then review when convenient?
Local core tests pass, and the change is limited to mosaic-core (Coordinator/QueryManager/types + test + short doc). Thanks!

derekperkins:
Doesn't [requestUpdate](https://idl.uw.edu/mosaic/api/core/client.html#requestupdate) already handle this?

derekperkins:
I wanted to follow up and see what the status is on this. Is it worth picking up this PR, or is that student work on a dependency graph far enough along that it'll make sense to push that through?

cc @SeanCassiere

Issue #977, opened 2026-03-10
Filtering on unnested data doesn't get quoted properly


mgd722:
I have data like:
```sql
(id INT, type VARCHAR, tags VARCHAR[])
(1, 'Edit', ['Roads', 'Lakes'])
(2, 'Tool', ['Buildings', 'Dimension Lines'])
```
I want to make an event type chart and an unnested tags chart that crossfilter each other. The tag chart renders as expected with this code:
```js
const tagPlot = vg.plot(
  vg.barX(vg.from('events', { filterBy: $filter, unnest: 'tags' }), {
    x: vg.count(),
    y: 'tags',
    fill: '#1f77b4',
    sort: { y: '-x', limit: 15 },
  }),
  vg.toggleY({ as: $filter }),
  vg.width(tagChartContainer.clientWidth || 500),
  vg.height(300),
  vg.marginLeft(120),
);
```
but when I hover/click a bar to filter the chart I get an error:
```
Error: Parser Error: syntax error at or near "Lines"
LINE 2: ... "events" AS "source" WHERE ("tags" IN (Dimension Lines)) GROUP BY "type"
                                                             ^
```
It seems that the unnested tag isn't getting quoted in the where clause.

### [Codepen example](https://codepen.io/mgd722/pen/raMMoLM)
_The unnesting doesn't seem to work at all in my example-- I would expect each tag to have its own bar in the chart?_

mgd722:
Thanks @derekperkins. I'm working (struggling) to better understand the codebase to see if I'll be able to submit a PR with a fix.

To make sure I'm in the ballpark, if unnesting works as intended, we would expect this chart from [my Codepen](https://codepen.io/mgd722/pen/raMMoLM) to list each tag individually (i.e. one bar for Buildings, one bar for Lakes, etc.):

<img width="741" height="386" alt="Image" src="https://github.com/user-attachments/assets/e1a457e1-bb15-4fd2-a274-8fbec346646d" />

If so, does the syntax in my example look correct?

PR #978, opened 2026-03-12
Fix unnesting in vgplot

mgd722:
Closes issue #977. 

## Summary of changes:

- __`Mark.js`__: Added `isUnnested()` method and modified `query()` to wrap unnested fields in `UNNEST()` in the select clause so DuckDB expands arrays into individual rows. This prevents the chart from rendering the entire array stringified (like `Roads,Lakes`) as bar labels.

- __`Toggle.js` & `Nearest.js`__: When the interactor's field is from an unnested source (`mark.isUnnested(field)`), switch from `clausePoints()` to `clauseList()`, which generates `list_has_any("tags", ['Roads'])` SQL instead of an unquoted `IN` predicate.

- __`SelectionClause.ts`__: Fixed `clauseList()` to pass array values directly to `listHasAny`/`listHasAll` (which wraps them in a properly-quoted `ListNode`), rather than through `literal()` which coerces arrays to unquoted strings via `Array.toString()`.

## Testing

When I ran `npm run test` everything passed. I tried to get this PR built in StackBlitz to drop a link to test it there, but couldn't get it to work. After updating the vgplot import in my standalone example HTML file from the issue ([unnest-bug.html](https://github.com/user-attachments/files/25950551/unnest-bug.html)) and putting it in the `dev` directory, when you go to `http://localhost:5173/dev/unnest-bug.html` the tags chart works as expected:

https://github.com/user-attachments/assets/11a1bfd8-3629-44f1-9c13-c75c765f0a12

### DIFF
--- a/packages/mosaic/core/src/SelectionClause.ts
+++ b/packages/mosaic/core/src/SelectionClause.ts
@@ -352,6 +352,10 @@ export function clauseList(
   }
 ): SelectionClause {
   const listFn = listMatch === 'all' ? listHasAll : listHasAny;
-  const predicate = value !== undefined ? listFn(field, literal(value)) : null;
+  const predicate = value != null
+    ? Array.isArray(value)
+      ? listFn(field, value as ExprValue[])
+      : listFn(field, literal(value))
+    : null;
   return { source, clients, value, predicate };
 }
--- a/packages/vgplot/plot/src/interactors/Toggle.js
+++ b/packages/vgplot/plot/src/interactors/Toggle.js
@@ -1,5 +1,5 @@
 /** @import { ClauseSource } from '@uwdata/mosaic-core' */
-import { clausePoints } from '@uwdata/mosaic-core';
+import { clauseList, clausePoints } from '@uwdata/mosaic-core';
 import { getDatum } from './util/get-datum.js';
 import { neq, neqSome } from './util/neq.js';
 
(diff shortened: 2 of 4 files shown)


packages/vgplot/plot/src/interactors/Toggle.js
```diff
@@ -42,10 +42,16 @@ export class Toggle {
...
-      clients: this.peers ? mark.plot.markSet : new Set().add(mark)
-    });
+    const clients = this.peers ? mark.plot.markSet : new Set().add(mark);
+    const source = /** @type {ClauseSource} */(this);
+
+    // if the field is unnested, use clauseList to generate list_has_any() predicate
+    if (fields.length === 1 && mark.isUnnested(fields[0])) {
+      const v = value?.length ? value.map(v => v[0]) : undefined;
+      return clauseList(fields[0], v, { source, clients });
+    }
```
derekperkins, line 52: in menu.js, I think there is a `listMode` or something that got passed through that let the user define if it used `list_has_any` or `list_has_all`. Is there a similar pattern that would work here?

packages/mosaic/core/src/SelectionClause.ts
```diff
@@ -352,6 +352,10 @@ export function clauseList(
   }
 ): SelectionClause {
   const listFn = listMatch === 'all' ? listHasAll : listHasAny;
-  const predicate = value !== undefined ? listFn(field, literal(value)) : null;
+  const predicate = value != null
```
derekperkins, line 355: Why did this switch from `value !== undefined` to `value != null`?

mgd722:
Unforeseen circumstances pulled me out of the office for the past few weeks. The project I was involved with has since moved to using a different charting library with `MosaicClient` instead of using vgplot.

I won't have time to finish this so I'll close it.

domoritz:
Curious to hear what you use if you don't mind sharing.

mgd722:
We're using d3 directly for the charts. Ultimately we're migrating a dashboard off of dc.js, so d3 was a familiar interface that allowed us more control than vgplot affords for now.

PR #1009, opened 2026-04-09
feat: Add client initialized flag, fix selection updates and vgplot interval fields.

jheer:
- Add `initialized` flag to Mosaic clients, is true when initialization is complete.
- Fix selection event dispatch to await clients that are still initializing, preventing query access race condition.
- Add tests for client prepare and query event ordering.
- Fix vgplot field extraction for bin transform, by supporting custom AST nodes with a `column` property.

### DIFF
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { SocketConnector } from './connectors/socket.js';
 import { type Connector } from './connectors/Connector.js';
-import { PreAggregator, type PreAggregateOptions } from './preagg/PreAggregator.js';
+import { PreAggregator, type PreAggregateInfo, type PreAggregateOptions } from './preagg/PreAggregator.js';
 import { voidLogger } from './util/void-logger.js';
 import { QueryManager, Priority } from './QueryManager.js';
 import { type Selection } from './Selection.js';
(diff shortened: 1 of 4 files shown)


packages/mosaic/core/src/Coordinator.ts
```diff
@@ -381,16 +381,22 @@ function updateSelection(
   const { preaggregator, filterGroups } = mc;
   const { clients } = filterGroups!.get(selection)!;
   const { active } = selection;
-  return Promise.allSettled(Array.from(clients, (client: MosaicClient) => {
+  return Promise.allSettled(Array.from(clients, async (client: MosaicClient) => {
+    // if client is not enabled, register a request for later
     if (!client.enabled) return client.requestQuery();
+
+    // if client is initializing, wait for it to complete
+    if (!client.initialized) await client.pending;
```
domoritz, line 389: Can pending ever be true when initialized? If not, then we could just await the pending.

packages/mosaic/core/src/MosaicClient.ts
```diff
@@ -31,7 +31,13 @@ export class MosaicClient {
   _coordinator: Coordinator | null;
   _pending: Promise<unknown>;
   _enabled: boolean;
-  _initialized: boolean;
+  /**
+   * Initialization state. One of:
+   *  - `0` uninitialized
+   *  - `-1` preparing
+   *  - `1` initialized
+   */
```
domoritz, line 37: Let's use an enum?

```
enum InitState {
  Uninitialized = 0,
  Preparing = -1,
  Initialized = 1,
}
```

domoritz (approved):
Tests would be good so we make sure that the logic is and stays correct.

packages/mosaic/core/src/Coordinator.ts
jheer: Yes, the pending promise allows tracking a pending request. It is often set when initialized.

packages/mosaic/core/src/MosaicClient.ts
jheer: I'll move to stricter types. An enum seems like overkill given that this is solely internal and IIRC TypeScript generates a somewhat complex object for enums.

PR #1007, opened 2026-04-05
feat: vgplot Python API

karthik-anand:
This PR introduces a Python API for `vgplot`, letting users build Mosaic visualizations programmatically in Python instead of writing raw JSON/YAML specs by hand. It adds a new `packages/vgplot/vgplot-python` package (published to PyPI as `vgplot`) and wires Python spec output into the docs and examples.

## Key additions

- **`vgplot` Python package** (`packages/vgplot/vgplot-python`) — a fluent builder API mirroring the vgplot JavaScript API, using Python-idiomatic snake_case naming (e.g. `line_y`, `x_domain` instead of `lineY`, `xDomain`). Python keywords are escaped with a trailing underscore (`from_`, `for_`).
- **Composable `View` objects** — `vg.plot()`, `vg.vconcat()`, and `vg.hconcat()` return `View` objects that render directly in Jupyter/marimo and export to JSON.
- **Automatic data/param discovery** — data sources (`vg.parquet()`, `vg.csv()`, `vg.table()`, `vg.spatial()`) and params/selections declared as local variables are discovered from the caller's frame at render/export time; no explicit registration or passing required.
- **AST-to-Python code generator** (`packages/vgplot/spec/src/ast-to-python.js`) — a JavaScript-side utility that translates parsed Mosaic spec ASTs into equivalent Python, used to regenerate the `specs/python/` and `docs/public/specs/python/` examples. Regenerate with `npm run docs:examples`.
- **Python specs for all examples** — every example now ships a generated Python version alongside the existing JSON/YAML/ESM, surfaced as a **Python** tab in the docs code groups.
- **Round-trip parity tests** (`test_full_round_trip.py`) — verify that running each generated Python spec produces JSON identical to its `specs/json/` reference fixture, and that the Python/JSON/ESM example sets stay in sync.

## How it works

The package is hand-written fluent helpers over Mosaic's spec format (not codegen from a schema):

1. **Builder functions** (`plot.py`, `data.py`, `params.py`, `encodings.py`) construct marks, directives, data sources, params/selections, and input widgets. Common marks and attributes have explicit helpers; anything else is reachable by its snake_case name via a dynamic `__getattr__` fallback (e.g. `vg.regression_y(...)`, `vg.x_tick_rotate(45)`) that maps snake_case → camelCase.

[...]

PR #1014, opened 2026-04-22
feat: Added step to build and test Spec Classes

karthik-anand:
Moves the spec classes build and test step into this workflow from a separate PR.
Adds --color=yes to the pytest invocation in the build and test Spec Classes CI step, matching the style of the server and widget steps for readable output in GitHub Actions logs.


domoritz:
How does this differ from #1007?

domoritz:
The goal was originally https://github.com/uwdata/mosaic/pull/1007#discussion_r3047755117. 

Let's redo this. What we want is a simple pull request that adds a check that generating specs in various languages does not introduce any changes.

PR #1017, opened 2026-04-29
chore: Bump to DuckDB v1.5.2.

jheer:
- Bump DuckDB and DuckDB-WASM dependencies to use DuckDB v1.5.2.


domoritz (approved):
We should bump the rust and go versions as well to make sure they are in sync. Merge #1018.

PR #1018, opened 2026-04-29
chore: update duckdb rust and go versions

domoritz:
For https://github.com/uwdata/mosaic/pull/1017


jheer (approved):

packages/server/duckdb-server-go/go.mod
```diff
@@ -5,20 +5,20 @@ go 1.25.0
 require (
 	github.com/apache/arrow-go/v18 v18.5.2
 	github.com/coder/websocket v1.8.14
-	github.com/duckdb/duckdb-go/v2 v2.10501.0
```
domoritz, line 8: Just fyi, duckdb uses a special version format now where they encode the duckdb version. It's pretty nice.

PR #1023, opened 2026-05-02
feat: Expand IN operator node coverage.

jheer:
- Add `InSubqueryNode` for inclusion test of values within subqueries.


domoritz (approved):
This works. 

Should we consider merging the in subquery and in operator? https://duckdb.org/docs/current/sql/expressions/in shows that in can even be a list of values and subqueries so breaking into two operators isn't fully generic/expressive. 

Image

jheer:
> Should we consider merging the in subquery and in operator?

Good question. I'm not sure yet. DuckDB's internal parser currently treats them as entirely different AST node types, but we don't have to follow that. We hard-wired a "tuple" interpretation into the existing `IN` operator node, so we'd need to make it more versatile, which might require a breaking change (e.g., to support tuple, list, or subquery). I'll look into it...

jheer:
@domoritz I redesigned `InOpNode` to be more flexible, and added the needed `Tuple` node type (which could also prove valuable elsewhere later). The `IN` operator node now accepts a tuple, list, or subquery. This is a breaking change to the `InOpNode` constructor signature, so will require a minor version bump following semver with `v0` major version.

Meanwhile, I kept the `isIn` helper function unchanged. If desirable we can add `isInList` or `isInQuery` helpers later, as I think it is useful / safer for the helper function API to reflect those semantic differences.

domoritz (approved):
Looks good. Test coverage is a bit thin.

domoritz:
Added two small tests in https://github.com/uwdata/mosaic/pull/1024.

Issue #892, opened 2025-09-18
pivot support


derekperkins:
We have a use case for pivoting data into columns for comparison purposes. The example from the DuckDB docs looks like this

### source data
<img width="358" height="432" alt="Image" src="https://github.com/user-attachments/assets/6edba329-edc7-466c-903b-5eb03bf52724" />

### displayed data after pivoting

<img width="378" height="178" alt="Image" src="https://github.com/user-attachments/assets/805dd2f2-b7ad-4119-8635-26d15a7b390e" />

Currently, mosaic-sql doesn't support the [PIVOT operator](https://duckdb.org/docs/stable/sql/statements/pivot.html). That shouldn't be too hard to add, but I'm curious if you have thoughts about what should be handled by the db/mosaic, and what should be a display side concern. My intuition is that using PIVOT in a query is probably the wrong place to do it, and that we'd be better off keeping it in the format of the source data above, which probably matches other metric displays on the page, and pivoting client side. Thoughts?

PR #1025, opened 2026-05-06
feat: DuckDB PIVOT query support

SeanCassiere:
Building on the discussion in #892, this PR is basically my first stab at adding support for DuckDB’s simplified `PIVOT` syntax to the sql package.

This add `Query.pivot(source)` as a new query-level AST node, with chainable support for `ON`, `IN`, `USING`, and `GROUP BY`:

```ts
Query
  .pivot('sales')
  .on('year')
  .in(2020, 2021)
  .using({ total: sum('amount') })
  .groupby('region')
```

The new node is exported publicly, works with the existing SQL visitor/codegen path, can be used as a `FROM` source in `SelectQuery`, and is covered by focused tests for construction, rendering, cloning, and clause behavior.

---

This PR intentionally sticks to DuckDB’s simplified `PIVOT` form, but of-course should be extendable to support the SQL specific implementation. It does not add SQL-standard PIVOT, runtime validation that `USING` expressions are aggregates, or rendered `LIMIT/OFFSET` behavior for pivot queries. So any feedback here would be great!


derekperkins:
Looking forward to this!

>  bunch of our optimizations would have to be carefully reviewed to make sure they still work

@domoritz brought this up in the issue, I'm not sure the best way to go about confirming this works with those materializations

derekperkins:
Comparing the DuckDB simplified syntax vs the SQL standard syntax for convenience here.

image

https://duckdb.org/docs/current/sql/statements/pivot#simplified-pivot-syntax

image

https://duckdb.org/docs/current/sql/statements/pivot#sql-standard-pivot-syntax

My gut feeling is that there's no need to support the less convenient syntax in the AST, even if it is the standard. Since we have the pivot AST node, translations to other SQL syntax should still be possible, and the convenience of the DuckDB setup is 100x better for the user.

packages/mosaic/sql/src/visit/codegen/duckdb.ts
```diff
@@ -223,6 +224,16 @@ export class DuckDBCodeGenerator extends SQLCodeGenerator {
...
   }
 
+  visitPivotQuery(node: PivotQuery): string {
+    const { source, _on, _in, _using, _groupby } = node;
+    const ref = isQuery(source) ? `(${this.toString(source)})` : this.toString(source);
+    const on = _on.length ? ` ON ${this.mapToString(_on).join(', ')}` : '';
+    const values = _in.length ? ` IN (${this.mapToString(_in).join(', ')})` : '';
+    const using = _using.length ? ` USING ${this.mapToString(_using).join(', ')}` : '';
+    const groupby = _groupby.length ? ` GROUP BY ${this.mapToString(_groupby).join(', ')}` : '';
+    return `PIVOT ${ref}${on}${values}${using}${groupby}`;
```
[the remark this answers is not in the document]
SeanCassiere: Applied 👍🏼

packages/mosaic/sql/test/pivot.test.ts
```diff
@@ -0,0 +1,324 @@
...
+describe('PivotQuery', () => {
+  it('constructs a pivot query from a table name', () => {
+    const query = Query.pivot('sales');
+
+    expect(query).toBeInstanceOf(PivotQuery);
+    expect(query.type).toBe('PIVOT_QUERY');
+    expect(isQuery(query)).toBe(true);
+    expect(isPivotQuery(query)).toBe(true);
+  });
+
```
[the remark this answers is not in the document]
SeanCassiere: Applied 👍🏼

packages/mosaic/sql/src/ast/query.ts
```diff
@@ -255,6 +271,138 @@ export class Query extends ExprNode {
...
+   */
+  clone(): this {
+    const { source, ...rest } = this;
+    // @ts-expect-error creates pivot query
+    return Object.assign(new PivotQuery(source), rest, {
+      _on: this._on.slice(),
+      _in: this._in.slice(),
+      _using: this._using.slice(),
+      _groupby: this._groupby.slice()
+    });
```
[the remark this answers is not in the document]
SeanCassiere: Applied 👍🏼

derekperkins:
@jheer @domoritz how does this look as an approach? We can track down any performance interactions if you think this is viable and can point us in the right direction.

packages/mosaic/sql/src/ast/query.ts
domoritz, line 1: To make this file a bit cleaner, please move the code related to pivot to a new file `ast/pivot.ts`.

packages/mosaic/sql/src/ast/query.ts
```diff
@@ -255,6 +271,140 @@ export class Query extends ExprNode {
...
+  /** The expressions that define pivot row groups. */
+  _groupby: ExprNode[] = [];
+
+  /**
+   * Instantiate a new pivot query.
+   * @param source The source relation to pivot.
+   */
+  constructor(source: PivotSource) {
+    super(PIVOT_QUERY);
+    this.source = maybeTableRef(source);
```
domoritz: Do we need to run `maybeTableRef` here? Source is already a resolved node, no?

packages/mosaic/sql/src/ast/query.ts
```diff
@@ -255,6 +271,140 @@ export class Query extends ExprNode {
...
+  clone(): this {
+    const { source, ...rest } = this;
+    // @ts-expect-error creates pivot query
+    return Object.assign(new PivotQuery(source), rest, {
+      _with: this._with.slice(),
+      _on: this._on.slice(),
+      _in: this._in.slice(),
+      _using: this._using.slice(),
+      _groupby: this._groupby.slice(),
+      _orderby: this._orderby.slice()
```
domoritz: Can you confirm that an order by clause gets serialized to SQL correctly?

domoritz (approved):
Looks good. Left three comments.

packages/mosaic/sql/src/ast/query.ts
SeanCassiere: I think we still need `maybeTableRef` here. `Query.pivot` and `WithClause.pivot` accept `PivotSource`, which can be a table name string, string path, or an already-built SQL node. The constructor is where table-name inputs like `Query.pivot('sales')` get converted into `TableRefNode`s.

For an already-built SQL node, `maybeTableRef` just returns it unchanged. Keeping this also makes sure `source` is always an AST node before codegen calls `toString(source)`.

packages/mosaic/sql/src/ast/query.ts
SeanCassiere: Good catch! Thanks, I missed this earlier.

SeanCassiere:
@domoritz thanks for your feedback Gone ahead and added them.

packages/mosaic/sql/src/ast/query.ts
domoritz: What about limit and offset? Where are those handled?

SeanCassiere: @domoritz at the moment, `PivotQuery` inherits the shared `Query` API, so `limit()` / `offset()` are available, but as I mentioned in the PR description I hadn’t added rendered `LIMIT/OFFSET` behavior for pivot queries yet.

I wasn’t totally sure what the right semantics should be here. Pagination with `PIVOT` feels a little subtle, since applying `LIMIT/OFFSET` after the pivot would page the resulting grouped/wide rows, while limiting the source relation before the pivot, which would mean something quite different.

Do you think `PivotQuery` should mirror `Query` and serialize `LIMIT/OFFSET` after the pivot? should those inherited methods be treated as unsupported/no-ops for pivot queries? or would you prefer another approach?

I'm learning towards adding them in (like with Query) and having it be user choice, but this is one of those things where I think y'all might have actual "right" answer.

SeanCassiere: Provided, its supposed to be like `Query`, https://github.com/uwdata/mosaic/pull/1025/commits/0eba48a14e229eb4883bf5736bbde10c4277f971 should suffice.

domoritz: I think since Sql supports limit and offset, it makes sense to support them. At the very least we should not silently drop them. 

Now the code for rendering SQL looks pretty long and repeated. Do you think we can reuse the Query serialization code?

SeanCassiere: Yes, we can reuse shared query-generation code, but we should share only dialect-agnostic query envelope mechanics and small clause helpers - since the statements for both are different enough in their placement of items in the generated statements.

The current AST models DuckDB simplified pivot, while standard pivot has a different placement inside `FROM`. Because of that, Pivot and Select should remain sibling renderers, even if they use common helpers for shared statement-level clauses.

Edit: added these changes in https://github.com/uwdata/mosaic/pull/1025/commits/cb9007924b992d8eeab572f984ef51b809f78aa6

packages/mosaic/sql/src/ast/query.ts
```diff
@@ -15,6 +16,15 @@ import { VerbatimNode } from './verbatim.js';
...
 import { WithClauseNode } from './with.js';
 
+// PivotQuery lives in pivot.ts to keep this module focused on the shared query
+// API. It registers this factory to avoid a runtime import cycle: PivotQuery
+// extends Query, while Query.pivot needs to construct a PivotQuery.
+let createPivotQuery: ((source: PivotSource) => PivotQuery) | undefined;
+
+export function setPivotQueryFactory(factory: (source: PivotSource) => PivotQuery): void {
+  createPivotQuery = factory;
+}
```
domoritz: I tried locally to remove this and didn't run into any issues with circular imports. can we remove this?

domoritz (approved):
Nice. Just one comment on the import. I didn't see any issues.

packages/mosaic/sql/src/ast/query.ts
SeanCassiere: Yup, thanks, I've nuked it.

domoritz:
Thank you!

domoritz:
To follow up here @SeanCassiere, can we close #892 or is there more to do. Also, could you help with docs for this feature: https://github.com/uwdata/mosaic/issues/1048.

SeanCassiere:
@derekperkins I think we can close out #892.

@domoritz sure thing! I'll throw up something for docs the addition of the `PIVOT` operator and tag yall for feedback. One thing to note, is that I'd be calling out that its for DuckDB's Simplified PIVOT and NOT the Standard SQL one.

domoritz:
Thank you. That makes sense.

PR #1030, opened 2026-05-14
ci: add generated-file consistency checks

karthik-anand:
Added two new CI checks that catch committed generated files drifting out of sync with their sources


derekperkins:
Can you rebase this on main?

domoritz:
Yes, please rebase so the diff is small and only checks that there are no new generated files. And please use semantic commit titles (e.g. fix: ci: feat:).

PR #1036, opened 2026-05-22
ci: add spec output verification to CI workflow

domoritz:
Ensure that we don't have changes in the generated specs


domoritz:
Replaces #1030

jheer (approved):

jheer:
Need to pass Node tests, but otherwise the changes look good to me.

domoritz:
Looks like we missed a change before so it's good that we have a test now. 

```diff
docs/public/specs/json/area-sine.json --- JSON
4        "description": "A test specification to compare M4 optimized and unoptimized area charts over a dense dual-tone sine wave.\n"
5      },
6      "data": {
7        "wave": {
8          "type": "csv",
12      "params": {
13        "brush": {
14          "select": "intersect"
15        }
16      },
   11   "vconcat": [
   12     {
   13       "plot": [
   14         {
```

docs/public/specs/json/area-sine.json
domoritz, line 1: Made this consistent with triangle-wave.json

domoritz:
@jheer can you add the sine example dataset to the Cloudflare bucket? https://idl.uw.edu/mosaic/examples/area-sine.html#area-sine-wave fails at the moment because the file is missing. 

Screenshot 2026-05-22 at 14 47 00

PR #1037, opened 2026-05-27
feat: add widget.sql and widget.data() for reading filtered rows in Jupyter

sittingthyme:
## Summary

Adds Python-side access to the widget's current filter state after user
interaction in Jupyter: inspect the generated SQL via `widget.sql` and read
the filtered rows via `widget.data()`, which returns a lazy DuckDB relation
to materialize with `.df()` (pandas), `.pl()` (Polars), `.arrow()`, or
`.fetchall()`.

- Combine the active selection predicates from the synced `params` traitlet
  into a SQL `WHERE` clause and run it on the widget's DuckDB connection
- Infer the source table from the spec's top-level `data` entries and the
  frames registered via the `data` constructor argument; when they name more
  than one table, `widget.data(table)` takes it explicitly and `widget.sql`
  returns `None` with a warning
- `filter_by` restricts the query to specific selections (names accepted
  with or without the leading `$`; unknown names raise)
- Join multi-clause selection predicates with `AND` in the JS param sync so
  they serialize as valid SQL
- Document usage in the Jupyter widget guide

## API

```python
widget.sql           # SQL for the current interactive state
widget.data().df()   # filtered rows as a pandas DataFrame

widget.data("weather")                    # explicit table (multi-table widgets)
widget.data(filter_by="range")            # apply a subset of selections
widget.data("weather", filter_by=["click", "range"])

### CODE REVIEW

File: packages/vgplot/widget/mosaic_widget/__init__.py [hunk 1]
```diff
@@ -104,3 +111,69 @@ def _handle_custom_msg(self, content: dict, buffers: list) -> None:
...
+            table: Source table name. Defaults to the unique table discovered
+                in the spec; required when the spec references multiple
+                tables.
+            filter_by: Optional selection name or sequence of names to limit
+                which selection predicates are applied. By default all
+                selections used as ``filterBy`` on ``table`` in the spec are
+                combined with ``AND``.
+        """
+        resolved = self._resolve_table(table)
+        return self.con.query(self._build_sql(resolved, filter_by=filter_by)).df()
```
domoritz: This always returns pandas. I wonder whether we should instead return the query result and let the user decide how they want the result? Or we could try to be smart about returning the data in the format is was originally provided to mosaic.

packages/vgplot/widget/mosaic_widget/spec_tables.py
```diff
@@ -0,0 +1,71 @@
...
+    Selection names appear in document order and are de-duplicated.
+    """
+    result = {}
+
+    def visit(node):
+        if isinstance(node, dict):
+            data = node.get("data")
+            if isinstance(data, dict) and _is_literal_table(data.get("from")):
+                table = data["from"]
+                names = result.setdefault(table, [])
```
domoritz: why not use a set?

packages/vgplot/widget/mosaic_widget/__init__.py
```diff
@@ -104,3 +111,69 @@ def _handle_custom_msg(self, content: dict, buffers: list) -> None:
...
+            )
+        if len(tables) > 1:
+            raise ValueError(
+                f"Spec references multiple source tables: {sorted(tables)}. "
+                "Pass a table name to widget.data(table)."
+            )
+        return next(iter(tables))
+
+    def _build_sql(self, table, filter_by=None):
+        if not _IDENTIFIER_RE.match(table):
```
domoritz: what does this defend against?

packages/vgplot/widget/mosaic_widget/__init__.py
```diff
@@ -104,3 +111,69 @@ def _handle_custom_msg(self, content: dict, buffers: list) -> None:
...
+                tables.
+            filter_by: Optional selection name or sequence of names to limit
+                which selection predicates are applied. By default all
+                selections used as ``filterBy`` on ``table`` in the spec are
+                combined with ``AND``.
+        """
+        resolved = self._resolve_table(table)
+        return self.con.query(self._build_sql(resolved, filter_by=filter_by)).df()
+
+    def _resolve_table(self, table):
```
domoritz: What's the logic here? Is this just returning the first table name? 

Maybe we should only allow not specifying a table name if there is only one table?

domoritz (commented):
Thanks for the pull request. I need to dig more into the logic to really understand why it needs to be this complex. Did you write this all by hand?

packages/vgplot/widget/mosaic_widget/__init__.py
sittingthyme: if the user inputs a table, then we verify that the table is actually part of the spec. if there is no table input, than we make sure that the spec actually has a table and there is only one table in the spec we can choose from.

packages/vgplot/widget/mosaic_widget/__init__.py
sittingthyme: it makes sure that the table names don't have malicious characters in it.

packages/vgplot/widget/mosaic_widget/spec_tables.py
sittingthyme: good idea

packages/vgplot/widget/mosaic_widget/__init__.py
domoritz: Yeah, I get that but what makes a name malicious and why are we worried about that?

packages/vgplot/widget/mosaic_widget/__init__.py
domoritz: Let's add a doc comment.

packages/vgplot/widget/mosaic_widget/spec_tables.py
```diff
@@ -0,0 +1,71 @@
...
+def collect_table_filters(spec):
+    """Walk a Mosaic spec and map each source table to its ``filterBy`` selections.
+
+    Mark ``data`` blocks of the form ``{"from": "<table>", "filterBy": "$sel"}``
+    contribute a ``table -> [selection]`` entry. Tables referenced without a
+    ``filterBy`` still appear in the result with an empty list. Dynamic ``from``
+    values that are param references (``"$tableParam"``) are skipped.
+
+    Selection names are de-duplicated and returned in sorted order.
+    """
```
domoritz: describe the why, not the what in comments. This is a very detailed comment that I could also get from reading the code.

packages/vgplot/widget/mosaic_widget/spec_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                visit(item)
+
+    visit(spec)
+    return result
+
+
+def resolve_predicates(params, selection_names):
+    """Collect non-empty ``predicate`` strings for the given selection names.
+
+    Names without a corresponding entry in ``params``, or whose entry lacks a
```
domoritz: same as above

packages/vgplot/widget/mosaic_widget/__init__.py
```diff
@@ -104,3 +111,69 @@ def _handle_custom_msg(self, content: dict, buffers: list) -> None:
             logger.warning(f"DONE. Slow query {uuid} took {total} ms.\n{sql}")
         else:
             logger.info(f"DONE. Query {uuid} took {total} ms.\n{sql}")
+
+    @property
+    def sql(self):
+        """SQL for the current interactive state of the (single) source table.
+
+        Combines the live selection ``predicate`` strings on ``params`` into a
```
domoritz: Describe the why, not the what. This comment is too detailed and I can read the exact logic from the code below.

domoritz (changes requested):

packages/vgplot/widget/mosaic_widget/__init__.py
sittingthyme: Like if someone were to download a spec from someone else and the table name had like a sql command in the name then it would mess up our query. Not sure if thats even a realistic scenario or not. if it's not then it can be removed.

domoritz: Someone could already put raw sql in specs so that's not a threat model I'm worried about.

packages/vgplot/widget/mosaic_widget/__init__.py
```diff
@@ -104,3 +103,70 @@ def _handle_custom_msg(self, content: dict, buffers: list) -> None:
...
+            table: Source table name. Defaults to the unique table discovered
+                in the spec; required when the spec references multiple
+                tables.
+            filter_by: Optional selection name or sequence of names to limit
+                which selection predicates are applied. By default all
+                selections used as ``filterBy`` on ``table`` in the spec are
+                combined with ``AND``.
+        """
+        resolved = self._resolve_table(table)
+        return self.con.query(self._build_sql(resolved, filter_by=filter_by))
```
sittingthyme: Changed it to return the query result. User can add .df() or whatever after the .data() to specify the result type

domoritz (changes requested):
Please make sure that tests pass

domoritz (approved):

PR #1038, opened 2026-05-27
docs: LLM docs

domoritz:
Using https://github.com/okineadev/vitepress-plugin-llms. 

We will investigate how much better this makes LLMs at building Mosaic apps. 

There will be two new files. 

/mosaic/llms.txt
and /mosaic/llms_full.txt


derekperkins:
@SeanCassiere do you have any insights from TanStack docs?

SeanCassiere:
> @SeanCassiere do you have any insights from TanStack docs?

From the distribution of docs standpoint, the `llms.txt` method would be the main thing you can do here.

Having the `.md` suffix work correctly is pretty important, irrespective of whether `llms` comes in or not. The suffixes for the guides/pages, is something Vite, Vercel, React, TanStack, unjs, and many more have implemented.

---

In terms of using the libs with LLMs, Skills would probably be where you should be looking into. Depending on what you want out of your skills, you may want to publish to [skills.sh](https://www.skills.sh). The problem is distribution and updating, since these are not tied to your package manager's update-system so its quite easy for these skills to become stale.

You could pair this with the `llms.txt` approach to have Mosaic `SKILL.md` essentially just be a couple lines about IF you are working with Mosaic then use a sub-agent to fetch the llms file from the website to then index into that. Its not the cleanest thing in the work, but its a pretty resistant method of not having to deal with the skill file rotting.

TanStack uses our own solution called [Intent](https://tanstack.com/intent), which TLDR, is essentially a way to bundle your skills in the NPM releases so that the skills are version and are locked to the packages themselves. We've got a few other big libraries using distribution flow: Redux, Apollo, tRPC, Resend, Electric SQL, GoDaddy, and a few more.

domoritz:
A skill would be great. Thanks for your notes on how to keep the skill up to date. I think I like the idea of having a simple skill that just points to the docs. Why do you suggest fetching the docs in a sub-agent? Wouldn't that mean that the main agent doesn't have the docs context? Do you have an example of a skill that uses this method? 

I'm leaving this as draft for now since my student @sittingthyme is going to evaluate the effect of these docs.

docs/.vitepress/config.js
```diff
@@ -7,7 +8,12 @@ export default defineConfig({
   description: 'Scalable, interactive data visualization',
   base: '/mosaic/',
   vite: {
-    resolve: viteConfig.resolve
+    resolve: viteConfig.resolve,
+    plugins: [
+      llmstxt({
+        description: 'Scalable, interactive data visualization'
```
domoritz: This needs to be expanded to be useful.

CONTRIBUTING.md
domoritz, line 1: Needs a complete pass from me.

SeanCassiere:
> ... Why do you suggest fetching the docs in a sub-agent? Wouldn't that mean that the main agent doesn't have the docs context? Do you have an example of a skill that uses this method?

It _could_ be dependent on the type of task being worked on, but its about what you are storing in primary agent's context. 

Since you'll be treating the llms file akin to a directory of a book, you'd be fetching the llms file, read/parse it, then gather which sub-guides/pages need to be looked at, then fetch those files, read/parse those files and bring them into context.

If this was being done by the primary agent, then all those extra messages for the tools calls will remain in context whilst functionally not really having any purpose going forward.

If it were to carried out by a sub-agent, then all it needs to do, is to issue the instructions to the sub-agent constructor, wait for the response, then catalog and parse the response.

I don't have a specific skill in mind that uses this. The closest one would be the Vercel the web-design-guidelines skill but even it doesn't explicitly call out the "sub-agent". That being said, it was created at a time when sub-agents were a bit newer.

---

@domoritz all in all, it might be early optimization that's not needed. It's mostly to do with how big llms file is and can you reasonably figure out the type of content being covered on each page/guide without having to fetch the target file.

domoritz:
Thanks for the explanation. We should look at how big the docs end up. I also see that pulling in all the examples may not be always what you need.

domoritz:
It looks like the examples account for about ~26% of the `llms_full.txt` file. But agents can use the individual files or handle the larger file if they have enough context. Let's leave this as is for now. Ready for a review and merge.

jheer (approved):

PR #1044, opened 2026-06-06
fix: Support filter pushdown in multi-table from clauses.

jheer:
- Support filter pushdown in multi-table `FROM` clauses.

This improves pushdown, for example to support implicit joins.

### DIFF
--- a/packages/mosaic/sql/src/transforms/filter-query.ts
+++ b/packages/mosaic/sql/src/transforms/filter-query.ts
@@ -21,20 +21,23 @@ export function filterPushdown(
   filter: FilterExpr
 ) {
   const clone = deepClone(query);
-  const tableRef = asTableRef(table)!;
-  walk(clone, (node) => {
-    if (node.type === SCALAR_SUBQUERY) {
-      return 1; // don't recurse
-    } else if (
-      isSelectQuery(node) &&
-      node._from.length === 1 &&
-      node._from[0] instanceof FromClauseNode &&
-      isTableRef(node._from[0].expr) &&
-      arrayEquals(node._from[0].expr.table, tableRef.table)
-    ) {
-      node.where(filter);
-    }
-  });
+  const tableRef = asTableRef(table);
+  if (tableRef) {
+    walk(clone, (node) => {
+      if (node.type === SCALAR_SUBQUERY) {
+        return 1; // don't recurse
+      } else if (isSelectQuery(node)) {
+        for (const source of node._from) {
+          if (source instanceof FromClauseNode &&
+            isTableRef(source.expr) &&
+            arrayEquals(source.expr.table, tableRef.table)
+          ) {
+            node.where(filter);
+          }
+        }
+      }
+    });
+  }
   return clone;
 }
 
--- a/packages/mosaic/sql/src/transforms/filter-query.ts
+++ b/packages/mosaic/sql/src/transforms/filter-query.ts
@@ -38,7 +38,7 @@ export function filterPushdown(
     // filtered table not present in query, nothing to do
     return clone;
   }
-  let filteredName = `_${tableRef?.name}`;
+  let filteredName = `_${tableRef.name}`;
   while (names.has(filteredName)) {
     filteredName = `_${filteredName}`;
   }
--- a/packages/mosaic/sql/src/transforms/filter-query.ts
+++ b/packages/mosaic/sql/src/transforms/filter-query.ts
@@ -1,7 +1,6 @@
 import { SCALAR_SUBQUERY } from '../constants.js';
 import type { FilterExpr } from '../types.js';
-import { FromClauseNode } from '../ast/from.js';
-import { isSelectQuery, type Query } from '../ast/query.js';
+import { Query } from '../ast/query.js';
 import { isTableRef, type TableRefNode } from '../ast/table-ref.js';
 import { asTableRef } from '../util/ast.js';
 import { deepClone } from '../visit/clone.js';
@@ -10,7 +9,7 @@ import { walk } from '../visit/walk.js';
 /**
  * Perform filter pushdown on a query: clones the given query and adds a
  * WHERE clause for the specified base table. Ignores scalar subqueries,
- * but will recurse into CTEs, joins, and other subqueries.
+ * but will recurse into CTEs, joins, and FROM subqueries.
  * @param query The query to clone and extend.
  * @param table The base table as a table name or table reference node.
  * @param filter The filter predicate expression to add.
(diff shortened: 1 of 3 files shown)


packages/mosaic/sql/test/filter-pushdown.test.ts
```diff
@@ -22,6 +22,18 @@ describe('filterPushdown', () => {
...
+  it('updates implicit joins', () => {
+    const q = Query
+      .select('a', 'b', column('v', 'O'))
+      .from(
+        new FromClauseNode(new TableRefNode('table'), 'T'),
+        new FromClauseNode(new TableRefNode('other'), 'O'),
+      )
+      .where(eq(column('id', 'T'), column('id', 'O')));
+    const f = filterPushdown(q, 'table', gt('x', 2));
+    expect(String(f)).toBe('SELECT "a", "b", "O"."v" AS "v" FROM "table" AS "T", "other" AS "O" WHERE ("T"."id" = "O"."id") AND ("x" > 2)');
```
domoritz: The where clause here does not specify which table it refers to (in this case `T`) and leave the ambiguity for the database to resolve. This assumes that pushed filter only references columns of the target table. If the column exists in both tables, we may get an error. The solution would be to alias the filer (`AND ("T"."x" > 2)`).

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -21,20 +21,23 @@ export function filterPushdown(
...
+    walk(clone, (node) => {
+      if (node.type === SCALAR_SUBQUERY) {
+        return 1; // don't recurse
+      } else if (isSelectQuery(node)) {
+        for (const source of node._from) {
+          if (source instanceof FromClauseNode &&
+            isTableRef(source.expr) &&
+            arrayEquals(source.expr.table, tableRef.table)
+          ) {
+            node.where(filter);
```
domoritz: If we have the same table aliased twice, we would get the same filter twice. This isn't a big deal since I don't expect us to have the same table aliased twice and even if we do, the database should figure out that the filter is redundant but wanted to flag in case I'm overlooking how it may be relevant.

packages/mosaic/sql/test/filter-pushdown.test.ts
domoritz: Maybe this is a non-issue in Mosaic specifically.

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -21,20 +21,23 @@ export function filterPushdown(
...
-    }
-  });
+  const tableRef = asTableRef(table);
+  if (tableRef) {
+    walk(clone, (node) => {
+      if (node.type === SCALAR_SUBQUERY) {
+        return 1; // don't recurse
+      } else if (isSelectQuery(node)) {
+        for (const source of node._from) {
+          if (source instanceof FromClauseNode &&
```
domoritz: FromClauseNode is not what JoinNode uses (https://github.com/uwdata/mosaic/blob/3df7b508be08b030e1730f865f6a98b95f6d8f9a/packages/mosaic/sql/src/ast/join.ts#L14) so this doesn't cover explicit joins. On purpose?

packages/mosaic/sql/src/transforms/filter-query.ts
jheer: I don’t expect this to happen often (if ever), but if so it should still be a correct query, so I’m not too concerned.

packages/mosaic/sql/src/transforms/filter-query.ts
jheer: That’s right. We need additional logic to cover explicit joins. We could try to add that here or make it the subject of a future PR.

packages/mosaic/sql/test/filter-pushdown.test.ts
jheer: Yeah I should test this to see if we get a binder error in that case. We’d need to add logic to rewrite the filter expression with table references. I don’t expect this to occur often but it would be good to cover it.

jheer: Just checked and this does result in a binder error. So given multiple tables we should add a table identifier, or find another solution (eg, create a CTE that filters the source table and then update source table references to use the CTE instead — which could handle explicit joins too and avoid repeated filters in the case of self joins).

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -21,20 +21,23 @@ export function filterPushdown(
...
-      node.where(filter);
-    }
-  });
+  const tableRef = asTableRef(table);
+  if (tableRef) {
+    walk(clone, (node) => {
+      if (node.type === SCALAR_SUBQUERY) {
+        return 1; // don't recurse
+      } else if (isSelectQuery(node)) {
+        for (const source of node._from) {
```
domoritz: One more thing. If there is no table that matches, then the filter is silently dropped. We can fix that in a follow up pull request.

jheer: Why do we need to “fix” that? We could throw an error, but if you ask to filter a table that isn’t there then doing nothing seems like a reasonable result as well.

domoritz: Not strictly needed but I think it would be nice/helpful. 

If you do a pushdown without a matching table, then the filter gets silently dropped. This means a query that previously failed (with a binding error) would not fail anymore (but produce incorrect results if you expected the filter to work). I think giving some feedback to the user (shoring an error or at least logging a warning) could help downstream users avoid mistakes that are hard to debug.

jheer:
@domoritz I changed the implementation to use a CTE to define the filtered table once and then reference that within the cloned query. In addition to multi-table queries, this now also handles joins. I also added more tests, including to make sure filterPushdown and pre-aggregation compose correctly. Let me know what you think.

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -21,21 +20,42 @@ export function filterPushdown(
...
+  walk(clone, (node) => {
+    if (isTableRef(node)) {
+      names.add(node.name);
+    }
+  });
+  if (!names.has(tableRef.name)) {
+    // filtered table not present in query, nothing to do
+    return clone;
+  }
+  let filteredName = `_${tableRef?.name}`;
```
domoritz:
```suggestion
  let filteredName = `_${tableRef.name}`;
```

packages/mosaic/sql/test/filter-pushdown.test.ts
```diff
@@ -1,31 +1,81 @@
...
+    expect(String(f)).toMatchInlineSnapshot(`"WITH "_table" AS (SELECT * FROM "table" WHERE ("x" > 2)) SELECT "v" FROM (SELECT "x" AS "v" FROM "_table" AS "table")"`);
   });
 
   it('updates CTEs', () => {
     const q = Query
       .with({ c: Query.select({ v: 'x' }).from('table') })
       .select('v').from('c');
     const f = filterPushdown(q, 'table', gt('x', 2));
-    expect(String(f)).toBe('WITH "c" AS (SELECT "x" AS "v" FROM "table" WHERE ("x" > 2)) SELECT "v" FROM "c"');
+    expect(String(f)).toMatchInlineSnapshot(`"WITH "c" AS (SELECT "x" AS "v" FROM "_table" AS "table"), "_table" AS (SELECT * FROM "table" WHERE ("x" > 2)) SELECT "v" FROM "c""`);
```
domoritz: Isn't the order wrong here? `_table` is defined after it's being used. I think this is why #1046 could be useful ;-)

packages/mosaic/sql/test/filter-pushdown.test.ts
```diff
@@ -1,31 +1,81 @@
...
+  });
+
+  it('updates self joins', () => {
+    const q = Query
+      .select('*')
+      .from(
+        cross_join('table', 'table')
+      );
+    const f = filterPushdown(q, 'table', gt('x', 2));
+    expect(String(f)).toMatchInlineSnapshot(`"WITH "_table" AS (SELECT * FROM "table" WHERE ("x" > 2)) SELECT * FROM "_table" CROSS JOIN "_table""`);
```
domoritz: The original query already doesn't run in duckdb. Self joins need an alias. Should we add an alias to the original query so it works before the filter pushdown? 

Image

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -10,7 +9,7 @@ import { walk } from '../visit/walk.js';
 /**
  * Perform filter pushdown on a query: clones the given query and adds a
  * WHERE clause for the specified base table. Ignores scalar subqueries,
- * but will recurse into CTEs, joins, and other subqueries.
+ * but will recurse into CTEs, joins, and FROM subqueries.
```
domoritz: Not correct anymore since we use CTEs now. 

```
Perform filter pushdown on a query: clone a query and push a filter into the given base table via a CTE, so every reference to the table sees the filtered rows. Skips scalar subqueries.
```

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -10,7 +9,7 @@ import { walk } from '../visit/walk.js';
...
  * @param query The query to clone and extend.
  * @param table The base table as a table name or table reference node.
  * @param filter The filter predicate expression to add.
```
domoritz: Maybe worth documenting that this filter should reference the un-aliased table name. Probably expected already but the logic can fail if someone uses an aliased filter. 

```suggestion
 * @param filter The filter predicate expression to add. May only reference columns of the table itself, not an alias assigned to it elsewhere in the query.
```

domoritz (approved):

PR #1050, opened 2026-06-19
feat: Preaggregation of having and qualify clauses.

### LINKED ISSUE #1047: Preaggregator fails on HAVING clauses that contain aggregates not covered in SELECT clause
Consider this candidate new test case for `/packages/mosaic/core/test/preaggregator.test.ts`:

```js
  it('supports queries with having clause', async () => {
    const query = (predicate: FilterExpr = []) => {
      return Query.from('testData')
        .select({ measure: sum('y'), avg_x: avg('x') })
        .groupby('dim')
        .where(predicate)
        .having(gt('avg_x', 2));
        // .having(gt(avg('x'), 2));
    };
    expect(await run(query)).toStrictEqual([13, true]);
  });
```

This succeeds as written, but if we switch to use the commented out `having` clause, then preaggregation queries will fail:

```
Error: Binder Error: Referenced column "x" not found in FROM clause!
Candidate bindings: "dim"

LINE 1: ... ("dim" IS NOT DISTINCT FROM 'b') GROUP BY "dim" HAVING (avg("x") > 2)))
                                                                         ^
```

The problem is that the preaggregator does not currently analyze and rewrite the contents of the `having` clause, so such clauses will only succeed if referencing columns produced by the `SELECT` clause.

jheer:
- Add preaggregation optimization of aggregates in `HAVING` and `QUALIFY` clauses.
- Add `setHaving` and `setQualify` methods to the `SelectQuery` class.
- Add new preaggregator tests.

Close #1047.

### DIFF
--- a/packages/mosaic/core/src/preagg/preagg-columns.ts
+++ b/packages/mosaic/core/src/preagg/preagg-columns.ts
@@ -8,6 +8,8 @@ export interface PreAggColumnsResult {
   groupby: Record<string, ExprNode>;
   preagg: Record<string, ExprNode>;
   output: Record<string, ExprNode>;
+  having: ExprNode[];
+  qualify: ExprNode[];
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/mosaic/core/src/preagg/preagg-columns.ts
```diff
@@ -57,43 +59,80 @@ export function preaggColumns(client: MosaicClient): PreAggColumnsResult | null
...
+    return null;
   }
+}
 
-  // bail if the query has no aggregates
-  if (!aggrs.size) return null;
+/**
+ * Analyzes an expression and returns a rewritten expression if preaggregation
+ * optimization is supported. Returns undefined if the expressions does not
+ * contain any aggregates. Throws if the expression contains an unsupported
```
domoritz, line 107: Feels a bit odd to throw if the aggregate cannot be optimized since the caller then cannot distinguish between a not supported aggregate or an error in the analyze expression function code.

packages/mosaic/core/test/preaggregator.test.ts
```diff
@@ -217,6 +217,28 @@ describe('PreAggregator', () => {
     expect(await run(query)).toStrictEqual([13, true]);
   });
 
+  it('supports queries with having clause', async () => {
+    const query = (predicate: FilterExpr = []) => {
+      return Query.from('testData')
+        .select({ measure: sum('y') })
+        .groupby('dim')
+        .where(predicate)
+        .having(gt(avg('x'), 2));
```
domoritz, line 226: Should check whether aliases also work here. (e.g. using `.having(gt('measure'), 2)`)

domoritz (approved):

packages/mosaic/core/test/preaggregator.test.ts
domoritz, line 226: Is this clause actually being used? Tests are still passing even after I comment out this line.

domoritz: The tests below already captures aliases.

domoritz (approved):

packages/mosaic/core/test/preaggregator.test.ts
jheer: You're right that the clause is not selective on the data, but it *is* evaluated, which is the main point of the test.

PR #1056, opened 2026-06-22
docs: document and demonstrate `PIVOT` query support

### LINKED ISSUE #892: pivot support

### LINKED ISSUE #1048: Document pivot
https://idl.uw.edu/mosaic/sql/#query should document how to use pivot introduced in #1025

SeanCassiere:
Documents DuckDB's simplified `PIVOT` support in mosaic-sql, and adds an example that uses it.

- API reference for `Query.pivot()`, the `PivotQuery` methods (`on`, `in`, `using`, `groupby`, `setGroupby`), and the `isPivotQuery` guard, plus a short overview in the SQL guide.
- A new Seattle weather example that pivots daily observations into a yearly cross-tab (rows = year, columns = weather type counts), shown in a table.

The example spec uses an explicit `ON ... IN (...)` list because DuckDB requires it when the PIVOT is registered as a `CREATE TABLE ... AS` relation.

Followup to
- #1025

Closes #892 
Closes #1048


SeanCassiere:
@domoritz added both the reference documentation and reused the Seattle Weather dataset for demoing PIVOT.

I was in two minds about adding the example, since I wasn't sure how refined y'all wanted them be. Let me know if you want me to nuke that commit.

domoritz:
Do you also see that the ci has not run?

derekperkins:
i see that it hasn't run, and I don't see the button to start it

SeanCassiere:
Perhaps it needs a reviewer to approve it? Or perhaps the "Approve workflows to run" button needs to be pressed?

image

Edit: Could also be GitHub having one of those days. I had Dependabot, on one of the other repos, just take the day off today - so perhaps that's related.

domoritz:
No, there is no pending approval. I suspect GitHub had a day and you may need to push something to trigger builds again.

Screenshot 2026-06-23 at 18 03 18

SeanCassiere:
> ... you may need to push something to trigger builds again.

@domoritz that seems to have done the trick 👍🏼

domoritz (approved):

domoritz:
Thank you!

PR #1057, opened 2026-06-23
Python api vgspec removal

karthik-anand:
vg.plot(), vg.vconcat(), and other layout helpers now return a View directly. The frame inspection that vg.spec() used to discover DataDef and param variables has moved into View.show() / View.to_dict(), so we go straight from building a view to displaying it with view.show() or view as the last notebook cell.
plot_defaults and config move onto the outermost layout call (vg.vconcat(..., plot_defaults={...})), matching how JS handles them as top-level spec fields.


domoritz:
This pull request should target cmudig:pythonAPI_improvements, not main, right?

domoritz:
Replaced by https://github.com/cmudig/mosaic/pull/28

PR #1060, opened 2026-06-27
docs: add LLM authoring guide, llms.txt, and spec reference for code generation

sittingthyme:
## Summary
This PR improves Mosaic documentation for LLM-assisted code generation, addressing common failures like invented APIs and mixed spec/JS syntax.
- Added an **[LLM Authoring Guide](/llm-authoring)** with a wrong→right API table, selection wiring rules, and JavaScript and Jupyter examples
- Added **[Spec Format vs JavaScript API](/api/vgplot/spec-to-js)** mapping page with side-by-side examples and a common mistakes section
- Added a generated **[Spec Schema Reference](/api/spec/schema-reference)**, which is the list of valid spec option names

Relates to #1038


domoritz:
Thanks for the pull request. It would be great to collect some evidence to show that the changes actually help and llm make a better mosaic app. 

I would like to achieve something specific here. Rather than focus on the spec format vgplot, I'd like LLMs to be able to make mosaic apps using the architecture. So maybe we want to test making applications with raw D3.

docs/.vitepress/config.js
```diff
@@ -7,7 +8,12 @@ export default defineConfig({
   description: 'Scalable, interactive data visualization',
   base: '/mosaic/',
   vite: {
-    resolve: viteConfig.resolve
+    resolve: viteConfig.resolve,
+    plugins: [
+      llmstxt({
+        description: 'Scalable, interactive data visualization'
```
domoritz, line 14: We probably need to expand this a bit.

docs/api/spec/format.md
```diff
@@ -96,7 +96,7 @@ To refer to a param or selection later in the specification, use a `$`-prefixed
   "singleSelection": { "select": "single" },
   "unionSelection": { "select": "union" },
   "intersectSelection": { "select": "intersect" },
-  "crossfilterSelection": { "select": "crossfiltter" }
```
domoritz, line 99: Let's fix this in a separate pull request.

PR #1065, opened 2026-07-05
fix: unnested array columns don't expand into rows and crash on cross-filter click 

### LINKED ISSUE #977: Filtering on unnested data doesn't get quoted properly

SuMayaBee:
# Description

I've been exploring Mosaic (love that it can push millions of rows straight to DuckDB and still feel snappy) and ran into this bug while building a chart on an array column. Turns out there was already an attempt at a fix in #978, closed by the author because of the time issue, so I picked it up, brought it up to date with current main, fixed a couple of things.

## Summary of changes:

- **`Mark.js`**: Wraps unnested fields in `UNNEST()` in `query()`, so arrays expand into individual rows instead of one stringified bar. `isUnnested()` only matches real column references, not just anything with a `.column` property.

- **`Toggle.js` & `Nearest.js`**: Unnested fields now use `clauseList()` instead of `clausePoints()`, generating `list_has_any("tags", ['Roads'])` instead of an unquoted `IN`. Both take a `listMatch` option (`'any' | 'all'`), same as `Menu`.

- **`SelectionClause.ts`**: `clauseList()` now passes arrays straight to `listHasAny`/`listHasAll` instead of through `literal()`, which was coercing them to unquoted strings. Also fixed empty-array clears filtering out every row.

- **`selection-clause.test.ts`** (new): Tests for all of the above, including the exact values from the issue that used to crash.

## Testing

Ran the new tests, they pass. Also checked it by hand against the repro from the issue: tag bars now render individually instead of as one combined array, and clicking a bar cross-filters cleanly with no console errors.

- **Before**: you can also try the bug live here: https://codepen.io/mgd722/pen/raMMoLM

https://github.com/user-attachments/assets/b227fb14-0581-4316-a752-20f25a543495

- **After**:

https://github.com/user-attachments/assets/4192dd67-9a6f-4b84-ad39-3895e30a0f52

Closes #977.

cc @derekperkins for review.


jheer (approved):
Overall, this is looking good to me.

@derekperkins if you can take a quick look as well to confirm changes to `clauseList` behavior, that would be great! 🙏

derekperkins (approved):

derekperkins:
> SelectionClause.ts: clauseList() now passes arrays straight to listHasAny/listHasAll instead of through literal(), which was coercing them to unquoted strings. Also fixed empty-array clears filtering out every row.

this is a good fix

SuMayaBee:
Thanks for the reviews @jheer @derekperkins @domoritz! Just pulled in the latest main and sorted out the conflict, so this should be all clear now.

domoritz:
This change introduced a regression: https://github.com/uwdata/mosaic/issues/1080. I sent a fix in https://github.com/uwdata/mosaic/pull/1081. It seems to be limited to toggling in legends so I will not deploy an emergency release until I have a review from Jeff.

PR #1067, opened 2026-07-06
feat: Python API for vgplot

### LINKED ISSUE #395: Python API
Design and develop a Python API for generating Mosaic specifications. Similar in spirit to [Vega-Altair](https://altair-viz.github.io/), the Python API should enable programmatic construction of a Mosaic JSON specification and enable easy operation with data from Pandas, Polars, DuckDB, etc.

domoritz:
Extends https://github.com/uwdata/mosaic/pull/1007 with auto-generation of the bulk of the API. 

Closes #395


packages/vgplot/vgplot-python/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[project]
+name = "vgplot"
+dynamic = ["version"]
+description = "A Python API for authoring Mosaic vgplot visualizations"
+readme = "README.md"
+requires-python = ">=3.9"
```
dangotbanned: `3.9` is past end-of-life, and the `3.10` EOL is in 3 months https://devguide.python.org/versions/

```suggestion
requires-python = ">=3.10"
```

Bumping can be helped out by the other suggestion 🙂

packages/vgplot/vgplot-python/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[dependency-groups]
+dev = [
+    "mosaic-widget",
+    "pytest",
+    "pytest-cov",
+    "ruff",
+    "ty",
+]
+
```
dangotbanned, line 24: Together, these rules will detect and autofix the use deprecated/outdated python APIs.

```suggestion
[tool.ruff.lint]
extend-select = [
    "FA",  # https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa
    "TC",  # https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc
    "UP",  # https://docs.astral.sh/ruff/rules/#pyupgrade-up
]
extend-safe-fixes = [
    "FA",
    "TC",
    "UP",
]

```

The most common case I saw was like here (https://github.com/cmudig/mosaic/blob/e7abc7461d8ab09ce208c2224e41d07a126a8305/packages/vgplot/vgplot-python/vgplot/plot.py#L4):

```py
from typing import Any, Dict, List, Optional, Union
```

`typing.Any` is the only import you'd need after re-running `ruff`

dangotbanned (commented):
RE: (https://github.com/uwdata/mosaic/issues/395#issuecomment-4894743403)

I would be creating a lot of automated noise with either suggestion as a  PR (e.g. https://github.com/narwhals-dev/narwhals/pull/3533).

So here's a comment instead of dumping a (possibly) huge diff on your lap

packages/vgplot/vgplot-python/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+[dependency-groups]
+dev = [
+    "mosaic-widget",
+    "pytest",
+    "pytest-cov",
+    "ruff",
+    "ty",
+]
+[tool.ruff.lint]
```
dangotbanned: My bad, sorry for breaking CI!

```suggestion
[tool.ruff]
fix = true

[tool.ruff.lint]
```

packages/vgplot/vgplot-python/vgplot/_generated/_types.py
```diff
@@ -0,0 +1,19 @@
+# DO NOT EDIT. Generated from the Mosaic JSON schema by bin/generate-python-api.js.
+# Regenerate with: pnpm run generate:python-api
+from typing import Any
+
+# Permissive value aliases. Channels/attributes accept column names, constants,
+# $param references, and SQL expressions; kept broad to avoid false type errors.
+ChannelValue = Any
+AttrValue = Any
+MarkData = Any
```
dangotbanned: > kept broad to avoid false type errors

You can avoid false positives, while still preserving some utility in the typing with:

```diff
- ChannelValue = Any
+ ChannelValue = <whatever-is-actually-valid> | Any
```

I'm not sure what the type for `<whatever-is-actually-valid>` is *yet*, but here's a translation of the comment:

```py
from __future__ import annotations

import datetime as dt
import decimal
from typing import Any, TypeAlias, TypedDict

Incomplete: TypeAlias = Any
"""https://github.com/python/typeshed/blob/a7bcce8b6df32dd85dc31691cef30af61e7b1371/stdlib/_typeshed/__init__.pyi#L45-L50"""

ColumnName: TypeAlias = str

NumericLiteral: TypeAlias = int | float | decimal.Decimal
TemporalLiteral: TypeAlias = dt.date | dt.datetime | dt.time | dt.timedelta
NonNestedLiteral: TypeAlias = NumericLiteral | TemporalLiteral | str | bool | bytes | None
NestedLiteral: TypeAlias = list[Any] | tuple[Any, ...] | dict[str, Any]
PythonLiteral: TypeAlias = NonNestedLiteral | NestedLiteral
"""https://github.com/narwhals-dev/narwhals/blob/99322f8605e3c310ceece07b502838dcdb9198b4/src/narwhals/typing.py#L272-L278"""

# NOTE: Change this to whatever {sub,super}set of the above is meant by "constants"
Constant: TypeAlias = PythonLiteral

# NOTE: I'm not sure if this is a kind of string, or a class - but that would go here instead
_ParamReference: TypeAlias = Incomplete

# NOTE: [`typing.LiteralString`](https://typing.python.org/en/latest/spec/literal.html#literalstring) is available from `3.11`
# But it is also available as a backport https://github.com/python/typing_extensions
_SQLString: TypeAlias = str


class SQLExpr(TypedDict):
    """(What should be) The return type of [`vg.sql`].

    [`vg.sql`]: https://github.com/uwdata/mosaic/blob/0623a0ed4f58653be73f18b57067a105952f983a/packages/vgplot/vgplot-python/vgplot/encodings.py#L5-L6
    """

    expr: _SQLString


# Permissive value aliases. Channels/attributes accept column names, constants,
# $param references, and SQL expressions; kept broad to avoid false type errors.
ChannelValue: TypeAlias = ColumnName | Constant | _ParamReference | SQLExpr | Incomplete
AttrValue: TypeAlias = ColumnName | Constant | _ParamReference | SQLExpr | Incomplete
MarkData: TypeAlias = Incomplete
```

> [!NOTE]
> Would need to add this kinda stuff wherever the codegen kicks off

dangotbanned (commented):
Ah I forgot to hit send on this comment I wrote up yesterday.

Will keep dipping back into this any chance I get ❤️

packages/vgplot/vgplot-python/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[dependency-groups]
+dev = [
+    "mosaic-widget",
+    "pytest",
+    "pytest-cov",
```
dangotbanned, line 20:
```suggestion
    "pytest",
    "pytest-cov",
    "pytest-xdist",
```

Not sure how many tests there are yet, but (https://github.com/pytest-dev/pytest-xdist) can keep things moving quickly as the test suite grows

- https://github.com/narwhals-dev/narwhals/issues/2818#issuecomment-3065140863
- https://github.com/narwhals-dev/narwhals/pull/2826#issuecomment-3068745928

domoritz: Thank for the pointers. Tests are fast enough for now.

packages/vgplot/vgplot-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+fallback, so a name that is missing from the generated API raises
+`AttributeError` (and is flagged by type checkers) rather than silently
+appearing.
+
+## Tests
+
+```sh
+cd packages/vgplot/vgplot-python
+uv run --group dev pytest      # round-trip + API + schema-coverage tests
+uv run --group dev ty check
```
dangotbanned, line 45: I'm not sure of the solution yet [^1], but @domoritz you probably want to run a type checker against the `specs/` directories:

```sh
cd packages/vgplot/vgplot-python
uv run --group dev ty check ../../../docs/public/specs/python/
```

Which turns up:

```
Found 103 diagnostics
```

And most look like this:

```rust
error[invalid-argument-type]: Argument to function `plot` is incorrect
  --> /mosaic/docs/public/specs/python/density1d.py:19:9
   |
19 |         vg.interval_x(bind=brush),
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^ Expected `Mark | Directive`, found `dict[str, Any]`
   |
info: Function defined here
  --> vgplot\plot.py:89:5
   |
89 | def plot(
   |     ^^^^
90 |     *items: Mark | Directive,
   |     ------------------------ Parameter declared here
```

[^1]: I certainly am not going to remember the path 😂

packages/vgplot/vgplot-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
+# Contributing to the vgplot Python API
+
+Most of the API — every **mark** and **plot attribute** — is generated from the
+Mosaic JSON schema, which is itself generated from the JavaScript spec
+(`packages/vgplot/spec/src/spec/Spec.ts`). The JS API is the source of truth.
+
+Hand-written (not generated): the runtime core (`plot.py`, `spec.py`), data
+sources (`data.py`), params (`params.py`), SQL encodings (`encodings.py`), and
+the interactor / input / legend helpers.
```
dangotbanned:
```suggestion
Mosaic JSON schema, which is itself generated from the JavaScript spec ([`Spec.ts`]). 
The JS API is the source of truth.

Hand-written (not generated): the runtime core (`plot.py`, `spec.py`), data
sources (`data.py`), params (`params.py`), SQL encodings (`encodings.py`), and
the interactor / input / legend helpers.

[`Spec.ts`]: ../../../packages/vgplot/spec/src/spec/Spec.ts
```

packages/vgplot/vgplot-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+
+```sh
+pnpm run generate:python-api
+```
+
+This rebuilds the schema, runs `bin/generate-python-api.js`, and formats the
+output into `vgplot/_generated/`. Commit the regenerated files.
+
+CI fails if the committed `vgplot/_generated/` output does not match a fresh
+regeneration, so the generated code always matches the schema.
```
dangotbanned:
```suggestion
This rebuilds the schema, runs [`bin/generate-python-api.js`], and formats the
output into [`vgplot/_generated/`]. Commit the regenerated files.

CI fails if the committed [`vgplot/_generated/`] output does not match a fresh
regeneration, so the generated code always matches the schema.

[`bin/generate-python-api.js`]: ../../../bin/generate-python-api.js
[`vgplot/_generated/`]: ./vgplot/_generated/__init__.py
```

packages/vgplot/vgplot-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Adding a new mark or attribute
+
+Add it on the JS side (`Spec.ts`), then run `pnpm run generate:python-api`.
```
dangotbanned:
```suggestion
Add it on the JS side ([`Spec.ts`]), then run `pnpm run generate:python-api`.
```

dangotbanned (commented):
Still learning my way around here

docs/public/specs/python/normalize.py
```diff
@@ -0,0 +1,42 @@
+import vgplot as vg
+
+stocks = vg.parquet("data/stocks.parquet")
+labels = vg.table(
+    "SELECT MAX(Date) as Date, ARGMAX(Close, Date) AS Close, Symbol FROM stocks GROUP BY Symbol"
+)
+
+point = vg.param({"date": "2013-05-13"})
```
dangotbanned: I just discovered this after experimenting with some typing on (https://github.com/dangotbanned/mosaic/commit/33ac9766ea92ebe51e19e197d2bb5c8e75782df8)


Rather than directly mirroring this:

https://github.com/uwdata/mosaic/blob/23313946dcbb885a437f4ef01416a9b6efe84e84/packages/vgplot/spec/src/spec/Param.ts#L24-L31

You should be able to write:

```py
import datetime as dt
import vgplot as vg


point = vg.param(dt.date(2013, 5, 13)})
point = vg.param(dt.datetime(2013, 5, 13, 0, 1, 2, 3))
```

Both can be converted with [`datetime.date.isoformat`](https://docs.python.org/3/library/datetime.html#datetime.date.isoformat)

> The initial parameter value as an ISO date/**time**

Same story with [`datetime.time.isoformat`](https://docs.python.org/3/library/datetime.html#datetime.time.isoformat)


## Related
- https://github.com/vega/altair/pull/3654
- https://github.com/vega/altair/pull/3653

domoritz: Oh, good catch. Yeah, natively supporting dates would be awesome. @SayeVikram @karthik-anand

dangotbanned: Nice one!

Okay so maybe more controversial of a take - but what would you think about explicitly *not supporting* the current version (in python)?

> ```py
> vg.param({"date": "2013-05-13"})
> ```

If you ever want to support struct-like (`dict`) input as a parameter, this would be kinda gnarly as the key `"date"` would change the data type 🤔

The JS example is using `new Date` anyway, so I'm hoping that is preferred 🙂 

https://github.com/uwdata/mosaic/blob/23313946dcbb885a437f4ef01416a9b6efe84e84/docs/public/specs/esm/normalize.js#L8

packages/vgplot/vgplot-python/CONTRIBUTING.md
dangotbanned: Apologies if I missed the explanation, but is there a reason for 2 versions of the `specs/` directories?

docs/public/specs/python/normalize.py
domoritz: Done

packages/vgplot/vgplot-python/vgplot/_generated/_types.py
domoritz: Done to some extent

packages/vgplot/spec/src/spec/Transform.ts
```diff
@@ -415,7 +415,7 @@ export interface Lead extends WindowOptions {
    * argument, default `null`). Both offset and default are evaluated with
    * respect to the current row.
    */
-  lag: Arg3Opt;
+  lead: Arg3Opt;
```
domoritz, line 418: Found a bug in the types

dangotbanned:
I've got a chunky comment/proposal that I just need to proofread tomorrow.

Have been enjoying myself poking around everywhere

domoritz:
Thanks everyone. Let's get this merged and into the hands of people and we can iterate in follow-up pull requests.

PR #1068, opened 2026-07-08
fix: Avoid typing issues in non-generated modules

dangotbanned:
Mainly trying to get less red squiggles in VS code with `pyright`.

But it also reduces the `ty` errors mentioned in (https://github.com/uwdata/mosaic/pull/1067#discussion_r3540122687):

```diff
- Found 103 diagnostics
+ Found 7 diagnostics
``` 

If any changes look surprising, try reading the commits in order for explanation 😄

### DIFF
--- a/packages/vgplot/vgplot-python/vgplot/__init__.py
+++ b/packages/vgplot/vgplot-python/vgplot/__init__.py
@@ -134,4 +134,4 @@ __all__ = [
     "toggle_y",
     "vconcat",
     "vspace",
-] + list(_generated_all)
+] + list(_generated_all)  # pyright: ignore[reportUnsupportedDunderAll]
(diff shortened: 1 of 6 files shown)


packages/vgplot/vgplot-python/vgplot/__init__.py
```diff
@@ -134,4 +134,4 @@
     "toggle_y",
     "vconcat",
     "vspace",
-] + list(_generated_all)
+] + list(_generated_all)  # pyright: ignore[reportUnsupportedDunderAll]
```
dangotbanned, line 137: I didn't fix this, but it is important to address before a release.

I've solved it before in the altair codegen - so that's a starting point

https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/tools/generate_schema_wrapper.py#L1116-L1166

PR #1070, opened 2026-07-08
fix: Handle case-insensitive column collisions in preagg dims.

jheer:
- Update preaggregator to handle case-insensitive column name collisions. This prevents preaggregation select queries from requesting column names that do not exist because they got re-written with suffixes (`_1`) by DuckDB to avoid case-insensitive name collisions.
- Refactor preaggregate column identification for cleaner, more straightforward handling.
- Add internal utility to resolve positional references to select clauses.
- Add more preaggregator tests.


packages/mosaic/core/test/preaggregator.test.ts
```diff
@@ -251,4 +252,26 @@ describe('PreAggregator', () => {
     };
     expect(await run(query)).toStrictEqual([3.5, true]);
   });
+
+  it('supports queries with renamed groupby dimensions', async () => {
+    const query = (predicate: FilterExpr = []) => {
+      return Query.from('testData')
+        .select({ measure: avg('x'), Cat: 'cat' })
+        .groupby("cat")
```
domoritz, line 261: Should we fix collisions in the group by as well? Like `.groupby('cat', 'CAT')`. This still produces `dims: ["cat", "CAT"]`.

domoritz: Similarly, `groupby('cat', 'cat')` are now resulting in dupes.

domoritz: I addressed these in 9e560bf7

domoritz (approved):
I addressed the comments already. Looks good to go now.

PR #1076, opened 2026-07-13
fix: Improve query consolidator.

jheer:
- Fix query consolidator to resolve positional expression references.
- Fix query consolidator to consolidate identical groupings with different input orders of group terms.
- Add more query consolidator test cases.

### DIFF
--- a/packages/mosaic/core/src/util/positional.ts
+++ b/packages/mosaic/core/src/util/positional.ts
@@ -1,5 +1,12 @@
 import type { ExprNode, LiteralNode, SelectClauseNode } from '@uwdata/mosaic-sql';
 
+/**
+ * Resolve a positional (1-based index) reference to a select clause entry.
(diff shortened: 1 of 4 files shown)


packages/mosaic/core/src/util/positional.ts
```diff
@@ -1,5 +1,12 @@
 import type { ExprNode, LiteralNode, SelectClauseNode } from '@uwdata/mosaic-sql';
 
+/**
+ * Resolve a positional (1-based index) reference to a select clause entry.
+ * Returns the corresponding select clause node if found, otherwise returns
+ * undefined.
+ * @param expr The SQL exprssion node to check.
```
domoritz:
```suggestion
 * @param expr The SQL expression node to check.
```

domoritz (approved):
Positional references can get out of sync with the consolidated query so we should resolve them earlier. I added a fix and test.

PR #1077, opened 2026-07-13
feat!: Use field expressions in selection clause.

### LINKED ISSUE #1071: Avoid high-cardinality materializations for expression-valued point selections.
Point selections may match either column values or the results of an expression.

To handle expressions, the current logic is:
- Extract all column references from the active predicate
- Include these extracted columns as groupby dimensions of the materialized preaggregations
- Upon selection, apply the predicate over the extracted columns in the preagg table.

This produces correct results, but it can explode the dimensionality of preaggregated materialized views.

Consider a binarizing expression like `CASE WHEN value > 100 THEN 'large' ELSE 'small' END`. If we materialize this expression and group, there are only two groups: `large` and `small` (well, actually three groups if there are `NULL` values). We'd like our preaggregates to leverage this small, reduced cardinality.

However, using our logic above, we actually group by all observed values of the `value` column, which could be arbitrarily large. Not good! We should consider mechanisms to better handle such expression-valued point selections. We might need to expand our selection clause metadata to do so?

jheer:
- Add required `fields` property to the `SelectionClause` interface, simplifying reasoning about selection structure.
- Add `clauseNone` to simplify creation of a clause that clears any predicates for a given clause source.
- Update point selection pre-aggregation to leverage fields, ensuring more efficient binning schemes.

*Note*: This PR makes a **breaking** change, in that it now requires that Selections include a defined `fields` array listing the columns or expressions that are filtered over. When using the standard the `clauseXXX` helper methods, no changes are needed. However, when generating a SelectionClause object directly, you will need to include the defined `fields` (or an empty array if none). These fields are used to improve Mosaic's automatic optimizations.

Fix #1071.


domoritz:
I added a test for the canonical case from https://github.com/uwdata/mosaic/issues/1071 so that we catch any kind of regression on the improved pre aggregation.

domoritz (approved):

PR #1078, opened 2026-07-14
feat: support dataframes directly in the spec

domoritz:


packages/vgplot/vgplot-python/vgplot/spec.py
```diff
@@ -41,25 +41,34 @@ def _collect_params(node: Any) -> list[_ParamBase]:
...
     if isinstance(node, list):
         found = []
         for item in node:
-            found.extend(_collect_datadefs(item))
+            found.extend(_collect_data_sources(item))
         return found
     return []
 
 
+def _frame_data_names(caller_locals: dict[str, Any]) -> dict[int, str]:
```
SayeVikram, line 63: if the same frame is bound to multiple names, it could lead to the names getting overwritten which the user wouldn't be able to diagnose. Maybe a comment here saying the behavior of this function would be good

packages/vgplot/widget/mosaic_widget/frame_interop.py
```diff
@@ -6,6 +6,20 @@
 logger.addHandler(logging.NullHandler())
 
 
+# Mirror of vgplot.data.is_frame; duplicated to avoid a widget→vgplot dependency.
```
SayeVikram: This could maybe lead to issues if vgplot's version of this function was to change? maybe moving it to a shared package could be better to avoid this issue

packages/vgplot/vgplot-python/vgplot/spec.py
domoritz: Good catch, we would get the same name when originally they were different but we are still referring to the same data so it shouldn't cause errors. I'll still add a comment.

packages/vgplot/widget/mosaic_widget/frame_interop.py
domoritz: The implementations are the same but their intention is slightly different. `vgplot.data.is_frame` answers "is this a dataframe-like object I should route into the spec's data section," and `is_registrable_frame` answers "is this something DuckDB can register." I clarified the code by removing the comment.

PR #1079, opened 2026-07-14
fix: improve widget handling of updates and teardown

domoritz:
This improves the performance of initial load and of updates in interactive specs.

### DIFF
--- a/packages/vgplot/widget/src/index.js
+++ b/packages/vgplot/widget/src/index.js
@@ -70,8 +70,13 @@ export default {
       return parts.length > 1 ? parts.map(s => `(${s})`).join(' AND ') : (parts[0] ?? '');
     }
 
+    let appliedSpecJSON;
+
     async function updateSpec() {
       const spec = getSpec();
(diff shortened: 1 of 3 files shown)


packages/vgplot/widget/src/index.js
```diff
@@ -70,8 +70,13 @@ export default {
...
+    let appliedSpecJSON;
+
     async function updateSpec() {
       const spec = getSpec();
+      const specJSON = JSON.stringify(spec);
+      if (specJSON === appliedSpecJSON) return;
+      appliedSpecJSON = specJSON;
       reset();
       logger.log('Setting spec:', spec);
       const dom = await instantiateSpec(spec);
```
SayeVikram, line 82: might not be substantial, but wrapping this with a try/catch might be good incase instantiating throws an error, in that case we can clear the dom/rety, etc.

domoritz: We can try that but it's orthogonal to this change

PR #1082, opened 2026-07-16
feat: types for interactor mark to avoid future regressions

domoritz:
Depends on #1065 and avoids future regressions.

### DIFF
--- a/packages/vgplot/plot/src/legend.js
+++ b/packages/vgplot/plot/src/legend.js
@@ -137,7 +137,11 @@ function interactorMark(legend) {
   const field = legend.field ?? findField(plot.marks, channel) ?? 'value';
   if (field) {
     const f = { field };
-    return { plot, channelField: c => channel === c ? f : undefined };
+    return {
+      plot,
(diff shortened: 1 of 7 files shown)


jheer (approved):

packages/vgplot/plot/src/legend.js
```diff
@@ -132,12 +132,20 @@ function getInteractor(legend, type) {
 }
 
 // generate a faux mark to pass to an interactor
+/**
+ * @import { InteractorMark } from './marks/Mark.js'
+ * @returns {InteractorMark | undefined}
```
SuMayaBee, line 137: ah this is what would've caught the side effect from #1065. good learning 👍

domoritz: Yeah, when I saw the regression I figured this could have been caught by types so I added them and a few more that seemed useful.

SuMayaBee: ahh gotcha, that's smart! Love that you covered the extra ones too 😄

PR #1084, opened 2026-07-17
fix: Pre-aggregate ORDER BY aggregates.

### LINKED ISSUE #1083: Pre-aggregator should handle aggregates within ORDER BY clause
DuckDB allows aggregate expressions in a query `ORDER BY` clause, but the pre-aggregator does not check for those. We should extend the pre-aggregate column analysis to include ORDER BY clauses.

jheer:
- Apply pre-aggregate analysis to ORDER BY expressions.
- Update test cases to include ORDER BY aggregate.

Fix #1083


domoritz (commented):
Looks good.

Issue #1075, opened 2026-07-10
Add a python equivalent of `mosaic-spec`


dangotbanned:
Mosaic has me more excited about data visualization than I have been in a while!  
I've been busy experimenting on [my fork](https://github.com/dangotbanned/mosaic/branches) and this idea fell out.

> [!NOTE]
> Apologies in advance, but I promise this wall of text was 100% human generated  ... 

## Background
While diving through `vgplot-python`, the most notable gap for me has been the typing.  

<details><summary>Why should I care what this guy has to say?</summary>
<p>

That's where I spent a lot of my time [contributing to `altair`] and more recently [in `narwhals` too].

I weasled my way in by in fixing typing issues.
In the process, I built up a greater understanding of how a project works and then expanded to proposing/implementing more "fun" features 😉

*Naturally*, I fell back into this familiar pattern again (https://github.com/uwdata/mosaic/pull/1067#discussion_r3530434157), (https://github.com/uwdata/mosaic/pull/1067#discussion_r3530787287), (#1068), (#1069)

[contributing to `altair`]: https://github.com/vega/altair/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Amerged+typing+author%3Adangotbanned
[in `narwhals` too]: https://github.com/narwhals-dev/narwhals/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Amerged+label%3Atyping+involves%3Adangotbanned
</p>
</details> 

I'm happy to continue down the road of incremental improvements.  
But I can see an see an alternative which - *while larger in scope* - has clear boundaries and could even be optional for `vgplot-python` itself.

## `mosaic-spec-python`
I write 0% TS/JS but have understood the majority of [`mosaic-spec`](https://github.com/uwdata/mosaic/tree/main/packages/vgplot/spec/src/spec) on first look.  
There are some more exotic things that haven't made their way to python *yet* ([PEP 827 - Type Manipulation]), which I needed to read up on.

[PEP 827 - Type Manipulation]: https://peps.python.org/pep-0827/

However, I believe that it would be possible to define a python counterpart to [`mosaic-spec`], primarily using [`TypedDict`] where you have [Interfaces] [^2].

[^2]: Python's [`Protocol`](https://typing.python.org/en/latest/spec/protocol.html) is conceptually an `interface`, but [`TypedDict`] was [motivated by representing JSON objects](https://peps.python.org/pep-0589/#motivation) with `dict`.

[...]

PR #1086, opened 2026-07-18
feat(python): Make `UNSET` a PEP 661 `sentinel`

dangotbanned:
## Description
As defined on `main`, [`UNSET`](https://github.com/uwdata/mosaic/blob/920bc67bd3134e6df947eeac30e65e2998cd2e23/packages/vgplot/vgplot-python/vgplot/_types.py#L22-L29) has a few issues - most of which are not that visible.

This PR introduces a conditional-backport for [`sentinel` (python `3.15`)].
Mosaic gets both the runtime and typing benefits - while requiring no new dependencies .
When available [^1], [`typing_extensions.sentinel`] is prefered - before defining a minimal [^2] vendored version [^3]

image1

The new tests are a combination of:
- tests adapted from the `typing_extensions` test suite ([1](https://github.com/python/typing_extensions/blob/83400e979b8e3b0b647f9a6a57f0275230e5f19f/src/test_typing_extensions.py#L9694-L9701), [2](https://github.com/python/typing_extensions/blob/83400e979b8e3b0b647f9a6a57f0275230e5f19f/src/test_typing_extensions.py#L9711-L9713))
- examples from the [`sentinel` (python `3.15`)] docs
- and a handful of runtime & static cases I didn't steal 😅

[^1]: [`typing_extensions>=4.16.0`](https://github.com/python/typing_extensions/releases/tag/4.16.0) is installed
[^2]: I've documented [the original source](https://github.com/python/typing_extensions/blob/f29cd28d8ed7642cafb1d18daf5aa41be6a5c0aa/src/typing_extensions.py#L176-L271) and the differences are the exclusion of deprecation paths, 
      which serve no benefit to mosaic.
[^3]: We've used this technique [a few times](https://github.com/narwhals-dev/narwhals/blob/deffd1d1e11dd593675119c1445dc7d5291426c3/src/narwhals/_typing_compat.py) in `narwhals`, but not (yet) for `sentinel`

## What's fixed?
(4dcdb4304e91ddc29f0dcd2d8ec40c3c4a3446c0)

`UNSET` can be used directly in annotations and is understood by type checkers:

```py
from vgplot._types import UNSET

def function(arg: int | str | UNSET = UNSET) -> None:
    #                       ^^^^^^^
    #      would cause a runtime `TypeError` on main
    from typing_extensions import assert_never, assert_type

    if arg is not UNSET:  # would not perform type narrowing on `main`, 
                          # requires a solution like https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/schemapi.py#L1012-L1023
        assert_type(arg, int | str)
    elif isinstance(arg, (int, str)):
        assert_never(arg)
        #            ~~~ branch is unreachable because there is only 1 `UNSET`
    else:
        assert_type(arg, UNSET)
```

`UNSET` maintains identity when copied or pickled:

```py
import copy
import pickle

from vgplot._types import UNSET

print(
    copy.copy(UNSET)
    is copy.deepcopy(UNSET)
    is pickle.loads(pickle.dumps(UNSET))
    is UNSET
)
# True
```

`UNSET` now has a visible docstring:

image2

## Related
- [PEP 661 – Sentinel Values](https://peps.python.org/pep-0661/)
- [`sentinel` (python `3.15`)]
- [`typing_extensions.sentinel`]
- https://github.com/python/typing/pull/2277

[`sentinel` (python `3.15`)]: https://docs.python.org/3.15/library/functions.html#sentinel
[`typing_extensions.sentinel`]: https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.sentinel


dangotbanned:
I've had this sitting ready since shortly after opening #1075

But didn't want to create too much noise - so here's a present now if anyone wants it 😄

packages/vgplot/vgplot-python/vgplot/_compat.py
```diff
@@ -0,0 +1,84 @@
...
+import sys
+import typing
+
+# ruff: noqa: A002
+from contextlib import suppress
+from importlib.util import find_spec
+from typing import TYPE_CHECKING, Any, ClassVar
+
+
+def _sentinel_backport_pre_typing_extensions_4_16() -> Any:
```
domoritz, line 29: Can we remove this in the future? Would be good to add a note saying when we can remove this?

domoritz (approved):

packages/vgplot/vgplot-python/vgplot/_compat.py
```diff
@@ -0,0 +1,84 @@
...
+            return getattr(typing_extensions, "sentinel", typing_extensions.Sentinel)
+
+    return _sentinel_backport_pre_typing_extensions_4_16()
+
+
+if TYPE_CHECKING:
+    # Was renamed in https://github.com/python/typing_extensions/releases/tag/4.16.0
+    from typing_extensions import Sentinel as sentinel  # noqa: N813
+else:  # noqa: PLR5501
+    if sys.version_info >= (3, 15):
```
dangotbanned, line 97: @domoritz

> Can we remove this in the future? Would be good to add a note saying when we can remove this?

So this part would be detected by Ruff via [outdated-version-block (`UP036`)](https://docs.astral.sh/ruff/rules/outdated-version-block/#outdated-version-block-up036)

Anything that uses `sys.version_info` works the same and will be flagged in-step with https://github.com/uwdata/mosaic/blob/cdb956b8e9dc26c4963727e40d5b2a37d5e3764b/packages/vgplot/vgplot-python/pyproject.toml#L6

But on this part:

> when we can remove this?

Without dependencies?
[October 2031](https://peps.python.org/pep-0790/#lifespan)

With [`"typing_extensions>=4.16"`](https://github.com/python/typing_extensions/releases/tag/4.16.0)?
Any time 😄

dangotbanned: I could add some of that here I suppose 😅

image

dangotbanned: I went with a module doc instead, since I wrote one in (https://github.com/narwhals-dev/narwhals/blob/deffd1d1e11dd593675119c1445dc7d5291426c3/src/narwhals/_typing_compat.py)

image

domoritz: I'm generally a fan of updating version requirements when it helps us clean up stuff. We don't need to support the oldest Python versions so feel free to update version requirements.

dangotbanned: If we were to update requirements, adding this would be my preference as it is a single file dependency and very common:

```toml
dependencies = [
  "typing-extensions>=4.16 ; python_full_version < '3.15'",
]
```

Depending on the latest python version *can* simplify maintainence, but it will hurt adoption of Mosaic.
It would mean that no downstream packages can offer backwards compatibility, if they have `vgplot` as a required dependency

PR #1090, opened 2026-07-22
feat: Add QueryError, preagg update fallback.

jheer:
- Add `QueryError` class, which provides access to query SQL and more complete error reporting.
- Use `QueryError` as typed input to client `queryError` method.
- Update selection updates to fallback to standard queries if preaggregate updates fail with a query error.

### DIFF
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -251,7 +251,6 @@ export class Coordinator {
         data => client.queryResult(data).update(),
         err => {
           const e = new QueryError(err, query);
-          console.warn(`${e}`);
           this._logger?.error(e);
           client.queryError(e);
           return e;
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -402,13 +402,13 @@ function updateSelection(
 
     if (info) {
       // skip due to cross-filtering
-      if (info?.skip) return;
+      if (info.skip) return;
       // generate and issue preaggregate update query
-      const query = (info as PreAggregateInfo)?.query(active);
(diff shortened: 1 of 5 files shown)


packages/mosaic/core/src/Coordinator.ts
```diff
@@ -248,7 +249,13 @@ export class Coordinator {
     return client._pending = this.query(query, { priority })
       .then(
         data => client.queryResult(data).update(),
-        err => { this._logger?.error(err); client.queryError(err); }
+        err => {
+          const e = new QueryError(err, query);
+          console.warn(`${e}`);
```
domoritz: Isn't this a duplicate of the logging? The next line logs an error so why log a warning here as well?

packages/mosaic/core/src/MosaicClient.ts
```diff
@@ -166,7 +167,7 @@ export class MosaicClient {
    * @param error
    * @returns this
    */
-  queryError(error: Error): this { // eslint-disable-line @typescript-eslint/no-unused-vars
+  queryError(error: QueryError): this { // eslint-disable-line @typescript-eslint/no-unused-vars
```
domoritz, line 170: I haven't checked but do we need to update make-client.ts as well?

domoritz (approved):

packages/mosaic/core/src/Coordinator.ts
jheer: That's leftover from testing and should not be there. Thanks for the catch!

packages/mosaic/core/src/MosaicClient.ts
jheer: Good catch -- yes, we do. I've made the updates.

Issue #1088, opened 2026-07-19
Improve python tooling


dangotbanned:
## Description
In a similar vein to (#1087) I've run into a few other issues.
The general theme for fixes are:

- Configuration that makes existing tooling more useful
- Ensuring the (python) test suites are green[^2] by default

The end result? More things that just work™ and we get helpful feedback for things that don't/may not

[^1]: Small changes, *not* fully configuring a linter
[^2]: Or at least yellow with expectations

[`RUF10*`]: https://docs.astral.sh/ruff/rules/unused-noqa/

## Issues
I may split some of these out into other issues if they start looking big or a theme emerges.
Most can be individual PRs, just wanting to see them organized for now:

### Pytest
- [x] #1091
	- [`strict_xfail = true`](https://docs.pytest.org/en/stable/reference/reference.html#confval-strict_xfail) ([Related](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-strict-tutorial)) 
- [x] #1092
    - [`importorskip`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip-ref) for tests with optional dependencies
- [ ] [`filterwarnings = ["error"]`](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings) to convert *unhandled* warnings into errors, and either
	- Fixing the source, so that they don't surface to *users* during correct usage
	- Expecting the warnings in tests, so that they don't produce noise for *contributors*
- [x] #1094
    - Avoid working-directory-sensitive filepaths (tests don't need to depend on this)

### Ruff
Not sure if this was intentional, but I will mention it here: 
**most of what ruff can do is disabled.**

<details><summary>Is that so bad?</summary>
<p>

I'd consider that a risky combination with LLM usage, or *at least* more expensive to fix than having Ruff do it 😄

</p>
</details>

- [x] #1141 

### Typing
- [x] #1138
- [ ] `pyright`
    - [x] #1095
    - [ ] Fix `__all__` definitions (see (https://github.com/uwdata/mosaic/pull/1095#issue-4967568772))
        - #1154

### Misc
- [x] #1102 
    - Currently [`3.12`](https://github.com/uwdata/mosaic/blob/0426fba2afdf23e5b8b88f66a6e5dd77dc2f5a6e/.python-version#L1) and the [final beta for `3.15`](https://discuss.python.org/t/python-3-15-0-beta-4-is-here/108208) is out
[...]

PR #1092, opened 2026-07-23
test(python): Avoid requiring optional dependencies

dangotbanned:
## Description
I was planning to use [`pytest.importorskip`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip-ref) everywhere to skip optional dependencies.

After I did that for `pandas`, it looked like a more useful route would be *not* skipping most of the suite without `pandas` [^1].
So this extends those tests to cover a wider range (Pandas, Polars, PyArrow, Modin) and does it using Narwhals 🐳

[^1]: read the commits in order to see how that differs

## Related
- Part of #1088
- Child of #1091
- Follow-up tracking #1097

## Tasks
- [x] `vgplot-python`
- [x] `mosaic_widget`
  - [x] `test_frame_interop`
  - [x] `test_spec_input`
  - [x] `test_sql_data`


packages/vgplot/widget/mosaic_widget/tests/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+"""Columnar data for a dataframe backend."""
+
+NwDataFrame: TypeAlias = nw.DataFrame["NativeDataFrame"]
+"""A constructed [`narwhals.DataFrame`][], with a known backend."""
+
+DataFrameConstructor: TypeAlias = Callable[[Data], NwDataFrame]
+"""A constructor for a [`narwhals.DataFrame`][]."""
+
+
+# TODO @dangotbanned: Fix `modin` warnings
```
dangotbanned, line 112: FYI, this is insanely noisy currently:

```
29 passed, 1 xfailed, 22 warnings in 3.51s
```

The next part of #1088 will convert those into **errors** - so I'll address that there

packages/vgplot/widget/mosaic_widget/tests/test_spec_input.py
```diff
@@ -65,45 +72,45 @@ def to_dict(self, _context=None):
...
 
     # File-backed data stays in the spec; only the in-memory frame is registered.
     assert widget.spec["data"] == {"athletes": file_def}
     assert widget._registered_tables == {"weather"}
 
 
-def test_explicit_data_takes_precedence(pandas_frame):
-    override = pandas_frame.iloc[:1]
+# TODO @dangotbanned: Need an alternative to materializing as pandas
+# - E.g. Map `Implementation` -> `DuckDBPy{Connection,Relation}` export methods
```
dangotbanned, line 108: This would be useful to make the tests shorter and work for every eager backend independently.

But I could also see something like `MosaicWidget.collect` (e.g. leaning on [`nw.LazyFrame.collect`](https://narwhals-dev.github.io/narwhals/api-reference/lazyframe/#narwhals.dataframe.LazyFrame.collect)) being a nice API too

packages/vgplot/widget/mosaic_widget/tests/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+    Backend("dask", "dask.dataframe", Warn.ALL),
+)
+
+
+@pytest.fixture(
+    scope="session",
+    params=tuple(b for b in _BACKENDS if b.is_eager_allowed() and b.is_available()),
+    ids=str,
+)
+def eager(request: Request) -> Backend[EagerAllowed]:
```
dangotbanned, line 97: Hopefully soon this kind of thing will be available in Narwhals directly

- [ ] https://github.com/narwhals-dev/narwhals/issues/3653

packages/vgplot/widget/mosaic_widget/tests/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+        return self.value not in (_LAZY - {Impl.POLARS, "polars"})
+
+
+_BACKENDS: Final = (
+    Backend("polars"),
+    Backend("pyarrow"),
+    Backend("pandas"),
+    Backend("modin", "modin.pandas", Warn.COPY),
+    Backend("ibis", warn=Warn.ALL),
+    Backend("duckdb", warn=Warn.ALL),
```
dangotbanned, line 87: Does this seem odd to anyone else?

E.g., why do you need to materialize a `DuckDBPyRelation` to register it to duckdb?

packages/vgplot/widget/pyproject.toml
```diff
@@ -42,6 +42,7 @@ test = [
   "pyarrow-hotfix",
   "pytest-cov",
   "pytest",
+  "pytest-env",
```
dangotbanned, line 45: FYI, this is an "official" plugin (https://github.com/pytest-dev/pytest-env).


I don't feel *too* strongly about adding it.
But it is only 300 LOC and well-maintained (https://github.com/pytest-dev/pytest-env/releases)

packages/vgplot/widget/mosaic_widget/tests/conftest.py
```diff
@@ -0,0 +1,117 @@
+from __future__ import annotations
```
dangotbanned, line 1: ### What on earth is a `conftest`?

- https://docs.pytest.org/en/stable/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
- https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py

packages/vgplot/widget/mosaic_widget/tests/conftest.py
domoritz: Are you subscribed to the issue to clean things up here when this lands?

domoritz (approved):

packages/vgplot/widget/mosaic_widget/tests/conftest.py
dangotbanned: > Are you subscribed to the issue to clean things up here when this lands?

I sure hope so 😉 

image

domoritz:
Should we give you merge permissions and also make you a co owner of the python code?

dangotbanned:
> Should we give you merge permissions and also make you a co owner of the python code?

@domoritz I thought you'd make me work harder than this, but if it makes things easier for you then I'd be happy to 🙂

domoritz (approved):

packages/vgplot/widget/mosaic_widget/tests/conftest.py
domoritz: Not sure. Some of the way duckeb handles data frames is magic to me.

domoritz:
@dangotbanned you have merge permissions. I'll let you merge PRs once approved.

dangotbanned:
Thanks @domoritz!

PR #1095, opened 2026-07-24
chore(typing): Add `pyright` and fix errors

dangotbanned:
## Description
Main goal is to avoid red squiggles if a contributor opens a `.py` file in the workspace [^1].

**Where were the issues?**
See (a09f82b8e210afd1ef614a67f84395fd74e25c2f)

**How do we fix them?**
*I'm working on it*

[^1]: While using VSCode and [Pylance](https://github.com/microsoft/pylance-release)

## Possible `__all__` follow-up
I can fix this here, *but* the diff will be 98% related to it.
Did mention before (https://github.com/uwdata/mosaic/pull/1068#discussion_r3543675753) that fixing it is important:

<details><summary>Show <code>[reportUnsupportedDunderAll]</code></summary>
<p>

```py
mosaic/packages/vgplot/vgplot-python/vgplot/_generated/__init__.py:10:1
warning: Operation on "__all__" is not supported, so exported symbol list may be incorrect (reportUnsupportedDunderAll)
0 errors, 1 warning, 0 informations
```

</p>
</details>

There are many forms which are specified in [Typing spec - Library interface (public and private symbols)](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols).
Which forms are supported by each type checker[^2] **vary** and so a major issue in downstream projects is incomplete/invalid export definitions.

[^2]: aaaaaaand now there are [7 type checkers](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html) 🫠 

### Example

This code will fail at runtime:
```py
def check_exports_bad() -> None:
    import vgplot

    _not_really_there = vgplot.attributes.align

check_exports_bad()
```
```py
AttributeError: module 'vgplot' has no attribute 'attributes'
```

#### Pyright
Because it doesn't understand the `__all__` definition *fully*, it is inferring that `_generated.attributes.py` is available at the top level.
So it thinks that was valid and here you can see "Go to definition":

image

#### ty
However, with `ty` this is correctly reported as an error:

```rust
error[unresolved-attribute]: Module `vgplot` has no member `attributes`
   --> packages/vgplot/vgplot-python/tests/test_python_api.py:480:25
    |
480 |     _not_really_there = vgplot.attributes.align
    |                         ^^^^^^^^^^^^^^^^^
    |

Found 1 diagnostic
```

#### Others
- (https://github.com/facebook/pyrefly) matches `ty`
- (https://github.com/python/mypy) doesn't recognise anything in `__all__`
    - It reports `attributes` as an error, but also **316 others** that are in the runtime definition

## Related
- #1088
- https://github.com/microsoft/pyright
- https://github.com/vega/altair/pull/3482

### DIFF
--- a/packages/vgplot/vgplot-python/tests/test_python_api.py
+++ b/packages/vgplot/vgplot-python/tests/test_python_api.py
@@ -101,7 +101,6 @@ class TestDataFrames:
         assert d["plot"][1]["data"] == {"from": "athletes"}
         assert d["data"]["weather"] is weather
         assert d["data"]["athletes"] == {"type": "csv", "file": "athletes.csv"}
-        vg.params
 
 
 if TYPE_CHECKING:
--- a/packages/vgplot/vgplot-python/tests/test_python_api.py
+++ b/packages/vgplot/vgplot-python/tests/test_python_api.py
@@ -1,5 +1,7 @@
 # Unit tests for the vgplot Python API, covering behaviors that the
 # generated-spec round-trip suite does not exercise directly.
+from typing import TYPE_CHECKING
+
 import pytest
 
 import vgplot as vg
@@ -99,3 +101,18 @@ class TestDataFrames:
         assert d["plot"][1]["data"] == {"from": "athletes"}
         assert d["data"]["weather"] is weather
         assert d["data"]["athletes"] == {"type": "csv", "file": "athletes.csv"}
+        vg.params
+
+
+if TYPE_CHECKING:
+
+    def typing_dunder_all() -> None:
+        # Ok
+        vg.arrow
+        vg.errorbar_x
+        vg.mark
+
+        # TODO @dangotbanned: Get pyright to error on these
+        # Error (`_generated` modules)
+        vg.attributes  # ty: ignore[unresolved-attribute]
+        vg.marks  # ty: ignore[unresolved-attribute]
--- a/packages/vgplot/widget/mosaic_widget/__init__.py
+++ b/packages/vgplot/widget/mosaic_widget/__init__.py
@@ -103,10 +103,10 @@ class MosaicWidget(anywidget.AnyWidget):
                     f"spec must be a dict or have a to_dict() method, got {type(spec)}"
                 )
             try:
-                spec = to_dict(_context=caller_locals)
+                spec = to_dict(_context=caller_locals)  # pyright: ignore[reportAssignmentType]
             except TypeError:
-                spec = to_dict()
-        spec = _register_frame_data(spec, data)
+                spec = to_dict()  # pyright: ignore[reportAssignmentType]
+        spec = _register_frame_data(spec, data)  # pyright: ignore[reportArgumentType]
         if con is None:
             con = duckdb.connect()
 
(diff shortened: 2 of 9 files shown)


packages/vgplot/vgplot-python/tests/test_python_api.py
```diff
@@ -99,3 +101,18 @@ def test_frame_and_datadef_share_the_data_section(self):
...
+    def typing_dunder_all() -> None:
+        # Ok
+        vg.arrow
+        vg.errorbar_x
+        vg.mark
+
+        # TODO @dangotbanned: Get pyright to error on these
+        # Error (`_generated` modules)
+        vg.attributes  # ty: ignore[unresolved-attribute]
+        vg.marks  # ty: ignore[unresolved-attribute]
```
dangotbanned, line 117: [`vgplot/_generated/encodings.py`](https://github.com/uwdata/mosaic/blob/73178eaa3a668b4270dff68d9aee8a29f3e70af7/packages/vgplot/vgplot-python/vgplot/_generated/encodings.py) doesn't have the same issue because [`vgplot/encodings.py`](https://github.com/uwdata/mosaic/blob/73178eaa3a668b4270dff68d9aee8a29f3e70af7/packages/vgplot/vgplot-python/vgplot/encodings.py) "overrides" it.

I use quotes because in reality this is just inference failing 😅

packages/vgplot/widget/mosaic_widget/__init__.py
```diff
@@ -28,10 +28,15 @@
 
 
 class SupportsToDict(Protocol):
-    def to_dict(self) -> dict: ...
+    def to_dict(self, *, _context: dict[str, Any] | None = None) -> dict[str, Any]: ...
+
+
+def _has_to_dict(obj: Any) -> TypeIs[SupportsToDict]:
+    _sentinel = object()
+    return inspect.getattr_static(obj, "to_dict", _sentinel) is not _sentinel
```
dangotbanned, line 36: 1. `ty` is the only type checker that accepts narrowing with `{get,has}attr` (everything else requires this)
2. Doing that in the first place is unsafe https://docs.python.org/3/library/inspect.html#fetching-attributes-statically

dangotbanned:
Oh I forgot to mention, I **haven't** added `pyright` to CI or **added** any dev **dependencies**.

With this PR, the VSCode integration with the (https://github.com/microsoft/pylance-release) extension will work without errors.
The same is true for this command from the repo root:

```cmd
uv run --with=pyright pyright
```

## Why?
A large amount of the (existing) code is [gradually-typed](https://typing.python.org/en/latest/spec/concepts.html#gradual-types).
For me, closing those holes is a bigger priority than adding another type checker.

Addressing that can be done with a single type checker and will be less noisy for anyone less-familiar with python's type system along the way

domoritz (approved):

packages/vgplot/vgplot-python/tests/test_python_api.py
domoritz: Python type checking still has a ways to go.

PR #1099, opened 2026-07-27
fix: preserve visible table name in filter pushdown

### LINKED ISSUE #1074: filterPushdown produces unbindable SQL when columns are qualified by the base table name
Identified by claude as a potential issue when reviewing https://github.com/uwdata/mosaic/pull/1046

## Bug

`filterPushdown` rewrites *all* `TableRefNode`s that match the filtered table — including
column qualifiers — but string-based FROM entries are aliased back to the original table
name. The result is a query that references a name that no longer exists in scope.

Found while adding DuckDB parser/binder validation to the test suite (#1046).

### Repro

```js
import { column, filterPushdown, gt, Query } from '@uwdata/mosaic-sql';

const q = Query
  .select(column('num1', 't1'))  // column qualified by base table name
  .from('t1');
const f = filterPushdown(q, 't1', gt('num2', 2));
String(f);
```

Produces:

```sql
WITH "_t1" AS (SELECT * FROM "t1" WHERE ("num2" > 2))
SELECT "_t1"."num1" AS "num1" FROM "_t1" AS "t1"
```

DuckDB rejects this:

```
Binder Error: Referenced table "_t1" not found!
Candidate tables: "t1"
```

The FROM entry is aliased `AS "t1"` (because `Query.from('t1')` sets `alias = 't1'`), which
hides the CTE name — but the select-list qualifier was rewritten to `"_t1"`, so it dangles.

### Cause

The rewrite walk in `src/transforms/filter-query.ts` renames every matching table ref
regardless of position:

```js
walk(clone, (node) => {
  if (node.type === SCALAR_SUBQUERY) {
    return 1; // don't recurse
  } else if (isTableRef(node) && arrayEquals(node.table, tableRef.table)) {
    node.table = [filteredName];
  }
});
```

A matching `TableRefNode` can appear in positions with different requirements:

| position | after rewrite | correct behavior |
| --- | --- | --- |
| `FromClauseNode.expr` from `from('t1')` | `"_t1" AS "t1"` — visible name preserved | rename ✅ |
| bare join operand, e.g. `join('t1', 't2')` | `"_t1"` — visible name changes | rename ✅, and qualifiers must follow |
| column qualifier (`ColumnRefNode.table`) | `"_t1"."num1"` | must match the *visible* name of the source — renaming breaks the aliased case (this bug) |
| manually built `FromClauseNode` without alias | `"_t1"` — visible name changes | qualifier rewrite is then required |

Whether a qualifier should be renamed depends on whether the FROM entry it binds to keeps
its visible name — non-local information, so the per-node rewrite cannot be correct as is.

### Suggested fix

Invert the invariant: **every rewritten source keeps the original visible name, and column
qualifiers are never rewritten.** Using the `parent` argument of the `walk` callback:

- skip table refs whose parent is a `COLUMN_REF` / `COLUMN_PARAM` (qualifiers);
- when the parent is a `FromClauseNode` without an alias, set `alias` to the original
  table name;
- when the parent is a `JOIN_CLAUSE` with a bare `TableRefNode` operand, replace the
  operand with `new FromClauseNode(ref, originalName)`.

Then a qualifier like `"t1"` always binds, since the filtered source is visible as `"t1"`
everywhere.

Note: the naive fix (only skipping qualifier rewrites) would break the bare-join case,
which currently works precisely *because* qualifiers get renamed along with the unaliased
source.

### Workaround

Qualify columns by an explicit alias instead of the base table name
(e.g. `from({ t: 't1' })` + `column('num1', 't')`), or leave columns unqualified.

domoritz:
Closes https://github.com/uwdata/mosaic/pull/1089. Fixes #1074

### DIFF
--- a/packages/mosaic/sql/src/transforms/filter-query.ts
+++ b/packages/mosaic/sql/src/transforms/filter-query.ts
@@ -1,5 +1,7 @@
-import { SCALAR_SUBQUERY } from '../constants.js';
+import { COLUMN_PARAM, COLUMN_REF, SCALAR_SUBQUERY } from '../constants.js';
 import type { FilterExpr } from '../types.js';
+import { FromClauseNode } from '../ast/from.js';
+import { JoinNode } from '../ast/join.js';
 import { Query } from '../ast/query.js';
 import { isTableRef, type TableRefNode } from '../ast/table-ref.js';
 import { asTableRef } from '../util/ast.js';
@@ -45,13 +47,37 @@ export function filterPushdown(
     filteredName = `_${filteredName}`;
   }
 
-  // rewrite table references to use filtered data
-  walk(clone, (node) => {
+  // rewrite table references to use filtered data, keeping the visible name
+  const visibleName = tableRef.name;
+  walk(clone, (node, parent) => {
     if (node.type === SCALAR_SUBQUERY) {
       return 1; // don't recurse
-    } else if (isTableRef(node) && arrayEquals(node.table, tableRef.table)) {
-      // @ts-expect-error set read-only property
-      node.table = [filteredName];
+    }
+    if (!isTableRef(node) || !arrayEquals(node.table, tableRef.table)) {
+      return;
+    }
+    if (parent?.type === COLUMN_REF || parent?.type === COLUMN_PARAM) {
+      return; // qualifiers bind to the visible name, so leave them as-is
+    }
+
+    // @ts-expect-error set read-only property
+    node.table = [filteredName];
+
+    if (parent instanceof FromClauseNode) {
+      if (!parent.alias) {
+        // @ts-expect-error set read-only property
+        parent.alias = visibleName;
+      }
+    } else if (parent instanceof JoinNode) {
+      // bare join operands have no alias, so wrap to preserve the name
+      const wrapped = new FromClauseNode(node, visibleName);
+      if (parent.left === node) {
+        // @ts-expect-error set read-only property
+        parent.left = wrapped;
+      } else if (parent.right === node) {
+        // @ts-expect-error set read-only property
+        parent.right = wrapped;
+      }
     }
   });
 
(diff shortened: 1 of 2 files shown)


packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -47,37 +48,31 @@ export function filterPushdown(
...
-        // @ts-expect-error set read-only property
-        parent.left = wrapped;
-      } else if (parent.right === node) {
-        // @ts-expect-error set read-only property
-        parent.right = wrapped;
-      }
+      // a bare join operand can not carry an alias, so wrap it in one
+      const side = parent.left === node ? 'left' : 'right';
+      // @ts-expect-error set read-only property
+      parent[side] = new FromClauseNode(node, visibleName);
```
domoritz, line 75: Just slightly simpler compared to copying the left and right logic

SuMayaBee: That's much cleaner, I was basically duplicating the branch. 😅

packages/mosaic/sql/src/transforms/filter-query.ts
```diff
@@ -45,13 +48,31 @@ export function filterPushdown(
...
+      return; // column qualifiers keep binding to the visible name
+    }
+    // @ts-expect-error set read-only property
+    node.table = [filteredName];
+    if (parent instanceof FromClauseNode) {
+      if (!parent.alias) {
+        // @ts-expect-error set read-only property
+        parent.alias = visibleName;
+      }
+    } else if (parent instanceof JoinNode) {
```
SuMayaBee, line 71: ah the pivot case is a good catch, i'd only thought about FROM and JOIN positions

domoritz:
Thank you @SuMayaBee for proposing the fix.

PR #1103, opened 2026-07-28
ci(python): bump `uv.lock`

dangotbanned:
## Description
I noticed while working on (#1065) that `pyarrow` was locked at `pyarrow==24`.

The version above that resolved (https://github.com/narwhals-dev/narwhals/issues/3560), which will mean there's a path to adding typing for `pyarrow`.
For now that would be (https://github.com/zen-xu/pyarrow-stubs) but eventually (https://github.com/apache/arrow/issues/32609)


dangotbanned:
oof (https://github.com/uwdata/mosaic/actions/runs/30376725677/job/90334180342?pr=1103)
I forgot about the ruff issue (https://github.com/uwdata/mosaic/issues/1088#issuecomment-5078727400)


## Updated
- (https://github.com/uwdata/mosaic/pull/1103/commits/0ce346ef1011ef7a9144d3a68232fffc02110398)
	- I was planning to fix/introduce some of these - just not *right now* 🤦‍♂️
- (https://github.com/uwdata/mosaic/pull/1103/commits/48adbe1006c156b9f96c09f0796bcc9573df79c0)
	- I can try to figure out the config to prevent this getting ruff'd, but left it in for now

domoritz (approved):

PR #1104, opened 2026-07-28
fix(sql): AST walk skips join conditions

SuMayaBee:
## Description

I was poking around in mosaic-sql and noticed that when a query has a JOIN, anything inside the ON condition was being silently skipped. So if a column or a parameter lived inside the join condition, the query analysis just didn't see it. It shows no error, it was quietly dropped.

I reproduced it on the published npm package: I built a query with columns in the join condition, asked for the query's columns back, and the ones from the join condition simply weren't there.

### **Before (published @uwdata/mosaic-sql):** 
The join condition columns "a" and "b" are missing, only "id" from the SELECT shows up.

image

---

### **After (with the fix):** 
"a" and "b" now show up alongside "id".

image

---
I also added tests so it doesn't slip back.

### DIFF
--- a/packages/mosaic/sql/src/visit/recurse.ts
+++ b/packages/mosaic/sql/src/visit/recurse.ts
@@ -57,7 +57,7 @@ export const recurse: Record<string, string[]> = {
   [FROM_CLAUSE]: ['expr'],
   [FUNCTION]: ['args'],
   [IN_OPERATOR]: ['expr', 'values'],
-  [JOIN_CLAUSE]: ['left', 'right', 'on', 'using', 'sample'],
+  [JOIN_CLAUSE]: ['left', 'right', 'condition', 'using', 'sample'],
   [LIST]: ['values'],
   [LOGICAL_OPERATOR]: ['clauses'],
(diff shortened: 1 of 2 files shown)


packages/mosaic/sql/src/visit/recurse.ts
```diff
@@ -57,7 +57,7 @@ export const recurse: Record<string, string[]> = {
   [FROM_CLAUSE]: ['expr'],
   [FUNCTION]: ['args'],
   [IN_OPERATOR]: ['expr', 'values'],
-  [JOIN_CLAUSE]: ['left', 'right', 'on', 'using', 'sample'],
+  [JOIN_CLAUSE]: ['left', 'right', 'condition', 'using', 'sample'],
```
SuMayaBee, line 60: This one line was the whole bug. the map looked for `on`, but the property on `JoinNode` is actually `condition`, so `walk` never went into the ON clause and skipped whatever was in there.

domoritz (approved):
Goo catch. Thank you!

SuMayaBee:
> Goo catch. Thank you!

Thanks for the quick review and merge! 😊

Issue #958, opened 2026-01-21
duckdb-server-go: --function-blocklist flag has no effect and exec command bypasses schema validation


danielbodart:
## Summary

While investigating multi-tenant security for duckdb-server-go (following up on #817), I found two bugs and an additional security consideration:

1. **`--function-blocklist` flag is not wired up** - The flag is parsed and the validator exists and is tested, but it's never actually applied to queries
2. **`exec` command bypasses all validation** - Schema validation via `--schema-match-headers` only applies to `arrow` and `json` commands, not `exec`
3. **Shared preagg schema exposes cross-tenant data** - If multiple tenants share the default `mosaic` preagg schema, aggregated data leaks between tenants

## Issue 1: Function blocklist is dead code

The `--function-blocklist` flag is parsed in `main.go:40-42` and passed to options via `query.WithFunctionBlocklist()` at line 89:

```go
// main.go:40-42
var functionBlocklist []string
if *functionBlocklistStr != "" {
    functionBlocklist = strings.Split(*functionBlocklistStr, ",")
}

// main.go:89
query.WithFunctionBlocklist(functionBlocklist),
```

The option is stored in `Options.FunctionBlocklist` (`options.go:26`), and a working validator exists (`validator.go:208-259`).

**However**, `query.New()` never reads `FunctionBlocklist` from the options, and the validator is never instantiated in production code.

The tests pass because they create the validator directly:
```go
// validator_test.go:213-214
if len(tt.functionBlocklist) > 0 {
    validators = append(validators, newFunctionBlocklistValidator(tt.functionBlocklist))
}
```

**Impact**: Users who set `--function-blocklist "read_parquet,iceberg_scan"` expecting to block these functions are not protected.

## Issue 2: `exec` command bypasses schema validation

In `server.go:260-268`, schema validation is only applied to `arrow` and `json` commands:

```go
switch *params.Type {
case CommandExec:
    return nil, false, s.db.Exec(ctx, *params.SQL)  // No allowedSchemas!

case CommandArrow:
    return s.db.QueryArrow(ctx, *params.SQL, allowedSchemas, useCache)

case CommandJSON:
    return s.db.QueryJSON(ctx, *params.SQL, allowedSchemas, useCache)
}
```

And in `query.go:186-193`, `Exec()` runs queries directly without any validation:

```go
func (db *DB) Exec(ctx context.Context, query string) error {
    _, err := db.db.ExecContext(ctx, query)  // No ValidateSQL call
[...]

Issue #947, opened 2025-12-16
go: replace custom table parser


derekperkins:
Right now, there's custom introspection code in the Go server that pulls out table references for multi tenancy checking. The returned json isn't well typed, and isn't guaranteed to be stable. Since I wrote it, there are a couple extensions available that do everything needed for table and function inclusion/exclusion. This should be more future proof and reliable than the current hand rolled parser.

- https://duckdb.org/community_extensions/extensions/parser_tools
- https://duckdb.org/community_extensions/extensions/poached / https://github.com/sidequery/poached

I can make it so that whichever extension is chosen is required only if opting into multi-tenancy mode.

PR #1106, opened 2026-07-31
fix(go-server): enforce query policies

derekperkins:
## Stack

**1. #1106 — fix(go-server): enforce query policies**
2. #1117 — fix(go-server): make cached responses policy- and format-safe
3. #1118 — fix(go-server): classify query and request failures
4. #1119 — fix(go-server): harden schema and function authorization
5. #1120 — fix(go-server): harden SQL serializer validation
6. #1121 — docs(go-server): document query policy boundaries

This is **1 of 6**. Merge bottom-up.

## What

- retain and normalize the configured function blocklist
- validate JSON and Arrow queries against schema and function policies
- reject `exec` when schema or function validation is active

## Why

The function blocklist was parsed but not retained for query execution, and `exec` could bypass request-scoped schema policy. DuckDB’s serializer cannot currently authorize arbitrary DDL/DML, so restricted-mode `exec` now fails closed.

## Impact

Policy-enabled deployments enforce their configured controls. Unrestricted `exec` behavior is unchanged. Full restricted DDL/DML support remains follow-up work under #947.

## Verification

- `golangci-lint run`

Part of #958

### DIFF
--- a/packages/server/duckdb-server-go/internal/query/options.go
+++ b/packages/server/duckdb-server-go/internal/query/options.go
@@ -2,6 +2,7 @@ package query
 
 import (
 	"log/slog"
+	"strings"
 	"time"
 )
 
@@ -65,8 +66,12 @@ func WithLogger(logger *slog.Logger) OptionFunc {
 
 func WithFunctionBlocklist(blockedFunctions []string) OptionFunc {
 	return func(opts *Options) error {
-		if len(blockedFunctions) > 0 {
-			opts.FunctionBlocklist = blockedFunctions
+		opts.FunctionBlocklist = make([]string, 0, len(blockedFunctions))
+		for _, function := range blockedFunctions {
+			function = strings.ToLower(strings.TrimSpace(function))
+			if function != "" {
+				opts.FunctionBlocklist = append(opts.FunctionBlocklist, function)
+			}
 		}
 		return nil
 	}
--- a/packages/server/duckdb-server-go/internal/query/query.go
+++ b/packages/server/duckdb-server-go/internal/query/query.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"hash/maphash"
 	"io"
@@ -19,6 +20,8 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+var ErrExecWithValidation = errors.New("query: exec command is disabled when schema or function validation is active")
+
 type DB struct {
 	db *sql.DB
 
@@ -30,7 +33,8 @@ type DB struct {
 	cache     *otter.Cache[uint64, []byte]
 	cacheSeed maphash.Seed
 
-	logger *slog.Logger
+	functionBlocklist []string
+	logger            *slog.Logger
 }
 
 // New creates a new DB instance using the provided DuckDB connector, opening a sql.DB and arrow connection.
@@ -86,7 +90,8 @@ func New(ctx context.Context, connector *duckdb.Connector, opts ...OptionFunc) (
 		cache:     cache,
 		cacheSeed: maphash.MakeSeed(), // Initialize the cache seed for consistent hashing
 
-		logger: o.Logger,
+		functionBlocklist: append([]string(nil), o.FunctionBlocklist...),
+		logger:            o.Logger,
 	}, nil
 }
 
@@ -184,9 +189,33 @@ func (db *DB) Close() {
 }
 
 func (db *DB) Exec(ctx context.Context, query string) error {
+	if len(db.functionBlocklist) > 0 {
+		return ErrExecWithValidation
+	}
+
 	_, err := db.db.ExecContext(ctx, query)
 	if err != nil {
 		return fmt.Errorf("query: failed to execute query: %w", err)
+	}
+
+	return nil
+}
+
+func (db *DB) validateQuery(ctx context.Context, query string, allowedSchemas []string) error {
+	validators := make([]Validator, 0, 2)
+	if len(allowedSchemas) > 0 {
+		validators = append(validators, newBaseTableValidator(allowedSchemas))
+	}
+	if len(db.functionBlocklist) > 0 {
+		validators = append(validators, newFunctionBlocklistValidator(db.functionBlocklist))
+	}
+	if len(validators) == 0 {
+		return nil
+	}
+
+	err := db.ValidateSQL(ctx, query, validators...)
+	if err != nil {
+		return fmt.Errorf("query: validation failed for query: %w", err)
 	}
 
 	return nil
@@ -218,12 +247,9 @@ func (db *DB) QueryJSON(ctx context.Context, query string, allowedSchemas []stri
 }
 
 func (db *DB) WriteJSON(ctx context.Context, query string, allowedSchemas []string, w io.Writer) error {
-	if len(allowedSchemas) > 0 {
-		btValidator := newBaseTableValidator(allowedSchemas)
-		err := db.ValidateSQL(ctx, query, btValidator)
-		if err != nil {
-			return fmt.Errorf("query: validation failed for query: %w", err)
-		}
+	err := db.validateQuery(ctx, query, allowedSchemas)
+	if err != nil {
+		return err
 	}
 
 	arrow, err := db.getArrowConn(ctx)
@@ -299,12 +325,9 @@ func (db *DB) QueryArrow(ctx context.Context, query string, allowedSchemas []str
 }
 
 func (db *DB) WriteArrow(ctx context.Context, query string, allowedSchemas []string, w io.Writer) error {
-	if len(allowedSchemas) > 0 {
-		btValidator := newBaseTableValidator(allowedSchemas)
-		err := db.ValidateSQL(ctx, query, btValidator)
-		if err != nil {
-			return fmt.Errorf("query: validation failed for query: %w", err)
-		}
+	err := db.validateQuery(ctx, query, allowedSchemas)
+	if err != nil {
+		return err
 	}
 
 	arrow, err := db.getArrowConn(ctx)
--- a/packages/server/duckdb-server-go/internal/query/validator.go
+++ b/packages/server/duckdb-server-go/internal/query/validator.go
@@ -205,7 +205,7 @@ func (v *baseTableValidator) Validate() []error {
 	return errs
 }
 
-// functionBlocklistValidator validates that the SQL query only accesses schemas that match request headers
+// functionBlocklistValidator validates that the SQL query does not use blocked functions.
 type functionBlocklistValidator struct {
 	blockedFunctions []string
 	errs             []error
@@ -217,24 +217,12 @@ func newFunctionBlocklistValidator(blockedFunctions []string) Validator {
 	}
 }
 
-func (v *functionBlocklistValidator) CheckNode(node map[string]any, keyStack []string) {
-	if len(keyStack) > 0 && keyStack[len(keyStack)-1] != "function" {
-		return
-	}
-
+func (v *functionBlocklistValidator) CheckNode(node map[string]any, _ []string) {
 	class, exists := node["class"]
 	if !exists {
 		return
 	}
-	if class != "FUNCTION" {
-		return
-	}
-
-	typeVal, exists := node["type"]
-	if !exists {
-		return
-	}
-	if typeVal != "FUNCTION" {
+	if class != "FUNCTION" && class != "WINDOW" {
 		return
 	}
 
--- a/packages/server/duckdb-server-go/internal/server/server.go
+++ b/packages/server/duckdb-server-go/internal/server/server.go
@@ -259,6 +259,9 @@ func (s *Server) execCommand(ctx context.Context, params QueryParams, allowedSch
 
 	switch *params.Type {
 	case CommandExec:
+		if len(s.schemaMatchHeaders) > 0 {
+			return nil, false, query.ErrExecWithValidation
+		}
 		return nil, false, s.db.Exec(ctx, *params.SQL) // No data to return for exec command
 
 	case CommandArrow:
--- a/packages/server/duckdb-server-go/internal/query/query_test.go
+++ b/packages/server/duckdb-server-go/internal/query/query_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 // setupTestDB creates a new in-memory DuckDB instance for testing
-func setupTestDB(t *testing.T) *DB {
+func setupTestDB(t *testing.T, opts ...OptionFunc) *DB {
 	t.Helper()
 
 	ctx := context.Background()
@@ -27,7 +27,8 @@ func setupTestDB(t *testing.T) *DB {
 		Level: slog.LevelError, // Only show errors during tests
 	}))
 
-	db, err := New(ctx, connector, WithLogger(logger))
+	opts = append([]OptionFunc{WithLogger(logger)}, opts...)
+	db, err := New(ctx, connector, opts...)
 	require.NoError(t, err)
 
 	// Clean up when test completes
@@ -40,6 +41,62 @@ func setupTestDB(t *testing.T) *DB {
 	})
 
 	return db
+}
+
+func TestDB_FunctionBlocklist(t *testing.T) {
+	db := setupTestDB(t, WithFunctionBlocklist([]string{" RANGE ", "MD5", "SUM", "ROW_NUMBER"}))
+	ctx := context.Background()
+
+	tests := []struct {
+		name     string
+		function string
+		query    string
+		format   string
+	}{
+		{
+			name:     "JSON table function",
+			function: "range",
+			query:    "SELECT * FROM range(3)",
+			format:   "json",
+		},
+		{
+			name:     "JSON scalar function",
+			function: "md5",
+			query:    "SELECT md5('mosaic')",
+			format:   "json",
+		},
+		{
+			name:     "JSON window aggregate",
+			function: "sum",
+			query:    "SELECT sum(i) OVER () FROM (VALUES (1), (2), (3)) t(i)",
+			format:   "json",
+		},
+		{
+			name:     "JSON window function",
+			function: "row_number",
+			query:    "SELECT row_number() OVER ()",
+			format:   "json",
+		},
+		{
+			name:     "Arrow table function",
+			function: "range",
+			query:    "SELECT * FROM range(3)",
+			format:   "arrow",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var err error
+			if tt.format == "arrow" {
+				_, _, err = db.QueryArrow(ctx, tt.query, nil, false)
+			} else {
+				_, _, err = db.QueryJSON(ctx, tt.query, nil, false)
+			}
+
+			require.ErrorContains(t, err, "use of function '"+tt.function+"' is not allowed")
+		})
+	}
 }
 
 func TestDB_Exec(t *testing.T) {
@@ -63,6 +120,14 @@ func TestDB_Exec(t *testing.T) {
 		err := db.Exec(ctx, "INVALID SQL STATEMENT")
 		assert.Error(t, err)
 		assert.Contains(t, err.Error(), "query: failed to execute query")
+	})
+
+	t.Run("function validation rejects exec", func(t *testing.T) {
+		db := setupTestDB(t, WithFunctionBlocklist([]string{"range"}))
+
+		err := db.Exec(ctx, "SELECT 1")
+		require.ErrorIs(t, err, ErrExecWithValidation)
+		assert.EqualError(t, err, "query: exec command is disabled when schema or function validation is active")
 	})
 }
 
--- a/packages/server/duckdb-server-go/internal/server/server_test.go
+++ b/packages/server/duckdb-server-go/internal/server/server_test.go
@@ -1,0 +1,34 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/duckdb/duckdb-go/v2"
+	"github.com/stretchr/testify/require"
+
+	"github.com/uwdata/mosaic/packages/server/duckdb-server-go/internal/query"
+)
+
+func TestExecCommandHonorsSchemaPolicy(t *testing.T) {
+	connector, err := duckdb.NewConnector(":memory:", nil)
+	require.NoError(t, err)
+
+	db, err := query.New(t.Context(), connector)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		db.Close()
+		require.NoError(t, connector.Close())
+	})
+
+	command := CommandExec
+	sql := "SELECT 1"
+	params := QueryParams{Type: &command, SQL: &sql}
+
+	s := New(db, []string{"X-Tenant"}, nil)
+	_, _, err = s.execCommand(t.Context(), params, nil)
+	require.ErrorIs(t, err, query.ErrExecWithValidation)
+
+	s = New(db, nil, nil)
+	_, _, err = s.execCommand(t.Context(), params, nil)
+	require.NoError(t, err)
+}


packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -217,11 +226,7 @@ func newFunctionBlocklistValidator(blockedFunctions []string) Validator {
 	}
 }
 
-func (v *functionBlocklistValidator) CheckNode(node map[string]any, keyStack []string) {
-	if len(keyStack) > 0 && keyStack[len(keyStack)-1] != "function" {
-		return
-	}
-
+func (v *functionBlocklistValidator) CheckNode(node map[string]any, _ []string) {
```
derekperkins, line 220: **Blocklist bypass: window-form calls are never checked.**

`CheckNode` requires both `class == "FUNCTION"` and `type == "FUNCTION"`, but DuckDB uses a different class for `OVER (...)` calls. `sum(a) OVER ()` serializes to:

```json
{"class":"WINDOW","type":"WINDOW_AGGREGATE","function_name":"sum","children":[...]}
```

and `row_number() OVER ()` to `{"class":"WINDOW","type":"WINDOW_ROW_NUMBER","function_name":"row_number"}`. Both carry `function_name`, so they're trivially checkable — the early `return` just skips them.

Confirmed on this branch with `WithFunctionBlocklist([]string{"sum"})`:

```
plain    sum(i)          -> query: access denied: use of function 'sum' is not allowed
windowed sum(i) OVER ()  -> <nil>
```

So any blocklisted aggregate escapes by appending `OVER ()`. Suggested fix:

```go
class, _ := node["class"].(string)
if class != "FUNCTION" && class != "WINDOW" {
	return
}
```

The separate `type` check then becomes redundant, since `type` varies by window kind (`WINDOW_AGGREGATE`, `WINDOW_ROW_NUMBER`, `WINDOW_RANK`, ...) while `class == "FUNCTION"` always pairs with `type == "FUNCTION"`.

Worth a case in `TestDB_FunctionBlocklist`: `SELECT sum(i) OVER () FROM range(3) t(i)` with `sum` blocked.

packages/server/duckdb-server-go/README.md
```diff
@@ -62,6 +71,10 @@ multiple users / customers share the same DuckDB server instance while keeping t
    the server will allow access to any schemas that match the header values. If no headers are present, and `--schema-match-headers`
    is set, the server will return a 401 Unauthorized error.
 
+_Note:_ If either `--schema-match-headers` or `--function-blocklist` is configured, `exec` requests are rejected until
+full-statement authorization is supported. Mosaic pre-aggregation uses `exec` to create schemas and tables, so it is also
```
derekperkins: Two things to tighten in this note.

**(a) `--function-blocklist` is not a filesystem boundary.** DuckDB's replacement scan reads files with no function node in the AST, so blocking `read_csv`/`read_parquet` — the example in the flag's own help text (`main.go:32`) — doesn't stop file reads. Confirmed on this branch with `WithFunctionBlocklist([]string{"read_csv"})` and a real `probe.csv` on disk:

```
SELECT * FROM read_csv('probe.csv')  -> query: access denied: use of function 'read_csv' is not allowed
SELECT * FROM 'probe.csv'            -> <nil>   (file contents returned)
```

The bare-path form serializes to `{"type":"BASE_TABLE","table_name":"probe.csv"}` with no `schema_name`, so only `baseTableValidator` catches it, and only when `--schema-match-headers` is also set (its empty-schema branch rejects it). A blocklist-only deployment gets nothing. Either reject unqualified base tables while a blocklist is active, or say plainly here that `--function-blocklist` must be paired with `--schema-match-headers` to be a filesystem boundary.

**(b) The `exec` restriction is broader than preaggregation.** Every `Coordinator.exec()` call goes out as `type: "exec"` (`Coordinator.ts:157`), not just `PreAggregator` (`PreAggregator.ts:131`, `:215`). Any app- or spec-level `mc.exec(...)` — temp-table setup, `CREATE VIEW` data definitions, extension loading — also fails in this mode, and `preagg: { enabled: false }` won't help those. Suggest: "all `exec` requests are rejected, including Mosaic pre-aggregation and any `Coordinator.exec()` calls your application makes."

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -274,6 +310,11 @@ func (db *DB) WriteJSON(ctx context.Context, query string, allowedSchemas []stri
 }
 
 func (db *DB) QueryArrow(ctx context.Context, query string, allowedSchemas []string, useCache bool) ([]byte, bool, error) {
+	err := db.validateQuery(ctx, query, allowedSchemas)
+	if err != nil {
+		return nil, false, err
+	}
+
 	var key uint64
 	var data []byte
```
derekperkins, line 305: **Pre-existing, but sitting inside a function this PR rewrites:** line 322 just below passes `"j"` to `cacheGet`, identical to `QueryJSON` at line 235. The doc comment on `cacheGet` says the opposite — "the key has to be different based on the output data type, so we can cache arrow and json separately" — so this reads as a typo for `"a"`.

Effect: with `persist: true`, a JSON request and an Arrow request for identical SQL share one cache entry and whichever runs first wins. Confirmed on this branch:

```
json:  hit=false  data="[{\"answer\":42}\n]"
arrow: hit=true   firstBytes="[{\"answer\":42}\n]"     <- JSON bytes served as Arrow IPC
```

`handleHTTP` then ships those JSON bytes with `Content-Type: application/vnd.apache.arrow.stream` and `Cache-Status: mosaic-duckdb-go; hit`, so the client hits a decode failure on traffic that looks entirely healthy server-side.

Not introduced here, and `TestDB_CacheValidatesSchemas` can't catch it because each subtest builds its own `DB`. But that test is the new cache-behavior test, so a same-`DB` json-then-arrow assertion would be cheap, and the fix here is one character.

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -183,7 +189,11 @@ func (db *DB) Close() {
 	}
 }
 
-func (db *DB) Exec(ctx context.Context, query string) error {
+func (db *DB) Exec(ctx context.Context, query string, allowedSchemas []string) error {
+	if len(allowedSchemas) > 0 || len(db.functionBlocklist) > 0 {
```
derekperkins: This fail-closed condition mixes a per-request value (`allowedSchemas`) with a server-lifetime one (`db.functionBlocklist`). It's correct today only because `handleHTTP` and `handleWebSocket` 401 when `schemaMatchHeaders` is configured and nothing matched (`server.go:87`, `server.go:165`), which is what guarantees `allowedSchemas` is non-empty whenever the schema policy is on.

That's a non-local invariant carrying a security property, and nothing states it — not the `Exec` signature, not a doc comment. A future caller passing `nil` (background job, admin endpoint, bootstrap path, a new transport) silently gets unrestricted DDL/DML on a server started with `--schema-match-headers`.

Consider making the policy explicit on `DB` instead: plumb the configured header list into `query.New` (`WithSchemaPolicy(enabled bool)`, or reuse the options struct) and gate on `db.schemaPolicyEnabled || len(db.functionBlocklist) > 0`. Then `Exec` fails closed regardless of what the caller passes, and `allowedSchemas` stays purely an authorization input rather than doubling as a policy flag.

`TestExecCommandPassesAllowedSchemas` currently locks in the coupled behavior, so it would need updating alongside.

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -192,7 +202,32 @@ func (db *DB) Exec(ctx context.Context, query string) error {
...
+	err := db.ValidateSQL(ctx, query, validators...)
+	if err != nil {
+		return fmt.Errorf("query: validation failed for query: %w", err)
+	}
+
+	return nil
+}
+
 func (db *DB) QueryJSON(ctx context.Context, query string, allowedSchemas []string, useCache bool) (json.RawMessage, bool, error) {
+	err := db.validateQuery(ctx, query, allowedSchemas)
```
derekperkins: Moving validation ahead of the cache lookup is the right call for correctness, but note what it costs: cache hits no longer avoid a DuckDB round-trip. Every request now runs `json_serialize_sql` on a pooled connection and walks the resulting AST in Go, *then* checks the cache.

For the Mosaic workload that's the hot path — brushing replays the same handful of queries with `persist: true` at interaction rate — so this partly undoes what the cache is there for.

A verdict cache keeps both properties: key on a `maphash` of `query` + sorted `allowedSchemas` (+ a blocklist generation), value is just pass/fail. The schema set has to be in that key or you reintroduce exactly the bug this PR fixes.

Fine as a follow-up — flagging it because it's a latency-sensitive path and the regression is invisible in the test suite.

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -192,7 +202,32 @@ func (db *DB) Exec(ctx context.Context, query string) error {
...
+	if len(db.functionBlocklist) > 0 {
+		validators = append(validators, newFunctionBlocklistValidator(db.functionBlocklist))
+	}
+	if len(validators) == 0 {
+		return nil
+	}
+
+	err := db.ValidateSQL(ctx, query, validators...)
+	if err != nil {
+		return fmt.Errorf("query: validation failed for query: %w", err)
```
derekperkins, line 218: Nit: the wrapped errors are already package-prefixed, so this doubles up. `ErrorDetails.Error()` returns `"query: %s (%s) at %s: %s"` (`validator.go:39`) and `functionBlocklistValidator` errors begin `"query: access denied: ..."`, producing:

```
query: validation failed for query: query: access denied: use of function 'range' is not allowed
```

`fmt.Errorf("validation failed: %w", err)` reads better, and the existing `require.ErrorContains` assertions still pass since they match on the inner substring.

packages/server/duckdb-server-go/internal/server/server.go
```diff
@@ -259,7 +259,7 @@ func (s *Server) execCommand(ctx context.Context, params QueryParams, allowedSch
 
 	switch *params.Type {
 	case CommandExec:
-		return nil, false, s.db.Exec(ctx, *params.SQL) // No data to return for exec command
+		return nil, false, s.db.Exec(ctx, *params.SQL, allowedSchemas) // No data to return for exec command
```
derekperkins: `handleHTTP` maps every `execCommand` error to 500 (`server.go:206`), which now covers this PR's two main new failure modes: an authorization denial (`access denied: unauthorized access to schema`) and `ErrExecWithValidation`. Both surface to the client as server faults.

Pre-existing for the schema case, but this PR makes it the common path and adds a second class of it, so it's a good moment to map them:

- `403` for validation/authorization failures — the request was understood and refused.
- `400` or `501` for `ErrExecWithValidation` — a configuration-level "not supported in this mode", not a fault.

`errors.Is(err, query.ErrExecWithValidation)` makes the second trivial; the first needs a sentinel or typed error out of `validateQuery`.

Beyond correctness of the status line: a 500 means client retry logic will retry a query that can never succeed, and any alerting on 5xx rate treats a correctly-refused tenant query as an incident. Worth fixing while the new failure paths are being added.

packages/server/duckdb-server-go/internal/query/options.go
```diff
@@ -65,8 +66,12 @@ func WithLogger(logger *slog.Logger) OptionFunc {
 
 func WithFunctionBlocklist(blockedFunctions []string) OptionFunc {
 	return func(opts *Options) error {
-		if len(blockedFunctions) > 0 {
-			opts.FunctionBlocklist = blockedFunctions
+		opts.FunctionBlocklist = make([]string, 0, len(blockedFunctions))
+		for _, function := range blockedFunctions {
+			function = strings.ToLower(strings.TrimSpace(function))
```
derekperkins, line 71: Good that the blocklist gets trimmed and lowercased here.

`--schema-match-headers` still doesn't: `main.go:37` is a bare `strings.Split`, so `--schema-match-headers="X-Tenant-Id, X-Org-Id"` yields `" X-Org-Id"`, and `http.Header.Get` returns `""` for a key containing a space (`textproto.CanonicalMIMEHeaderKey` passes invalid keys through untouched). It fails closed — that tenant loses its schema or gets a 401 — but silently, and the flag takes a comma list where the whitespace slip is easy to make. Same `strings.TrimSpace` treatment would be worth it while you're in here.

Minor and separate: this now assigns unconditionally where it previously only assigned for a non-empty input, so a later `WithFunctionBlocklist(nil)` clears an earlier one. Harmless for `main.go`'s single call site — just noting the change in the option's composition semantics.

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -218,14 +253,15 @@ func (db *DB) QueryJSON(ctx context.Context, query string, allowedSchemas []stri
 }
 
 func (db *DB) WriteJSON(ctx context.Context, query string, allowedSchemas []string, w io.Writer) error {
```
derekperkins, line 249: `WriteJSON` and `WriteArrow` have no callers anywhere in the module — `QueryJSON`/`QueryArrow` now go to the unexported `writeJSON`/`writeArrow`, and nothing else references the exported pair. Worth either deleting them or keeping them deliberately as the documented streaming entry point.

The reason it matters more after this refactor: `writeJSON`/`writeArrow` are now *unvalidated* primitives, and the only thing keeping a future in-package caller off them is that nobody has reached for one yet. Validation used to be co-located with execution; now it isn't. A one-line comment on each — "callers must validate first; see `validateQuery`" — would make the invariant explicit at the place someone would otherwise get it wrong.

derekperkins (commented):
## Review: enforce query validation policies

Reviewed the full diff against the branch, ran `go test -tags=duckdb_arrow ./... -count=1` (passes) and `go vet` (clean), plus targeted probes to confirm the behavioral claims below.

### What this does

Four fixes to the Go server's experimental tenant controls, all real bugs:

1. **`WithFunctionBlocklist` is retained.** The option previously landed in `Options` and was then dropped on the floor — `DB` never held it, so `--function-blocklist` was inert. Now stored on `DB`, trimmed, and lowercased. The lowercasing is correct: DuckDB folds function names in the AST even for quoted identifiers, so `SELECT "MD5"('mosaic')` serializes as `function_name: "md5"` and there's no case-based bypass.
2. **The keyStack gate in `functionBlocklistValidator.CheckNode` is gone.** The old `keyStack[len-1] != "function"` guard meant scalar and aggregate calls in a select list were never visited, so even a retained blocklist wouldn't have fired. Second half of the same bug.
3. **Validation moved ahead of the cache lookup** — the clear win. Previously a `persist: true` result cached for tenant A was handed to tenant B with no check at all, because validation lived inside `WriteJSON`/`WriteArrow` on the miss path only.
4. **`exec` fails closed under policy** — correct given `json_serialize_sql` only handles `SELECT`. The alternative was unvalidated DDL/DML.

Plus the `stringField` change in `ValidateSQL`, which removes a real panic: `m["error_subtype"].(string)` was an unchecked assertion on `nil` for sparse serializer responses (`PRAGMA version` returns `error_type`/`error_message` but no `error_subtype`/`position`). `TestDB_FunctionBlocklistHandlesUnsupportedStatements` pins it.

### Assessment

Direction is right and the tests are good. `TestDB_CacheValidatesSchemas` tests the actual bug shape (miss → deny → hit) rather than a happy path, and feeding `" RANGE "` through the blocklist exercises the normalization. Failing closed on `exec` rather than half-validating is the correct trade-off.

Two things I'd fix before merge, both confirmed by running against this branch:

- **The blocklist still has a hole.** `sum(a) OVER ()` serializes with `class: "WINDOW"`, so `CheckNode` returns before looking at `function_name`. With `sum` blocked, the plain call is denied and the windowed call succeeds. Two-line fix — thread on `validator.go`.
- **`--function-blocklist` does not stop file reads.** DuckDB's replacement scan emits a `BASE_TABLE` node, not a function node: with `read_csv` blocked, `FROM read_csv('probe.csv')` is denied and `FROM 'probe.csv'` returns the file contents. The flag's help text advertises `read_parquet` as its example, so this needs either enforcement or an explicit caveat.

Separately I found a **pre-existing** cache bug in code this PR rewrites: `QueryArrow` passes `"j"` to `cacheGet`, same as `QueryJSON`, so Arrow and JSON responses for identical SQL collide. I verified an Arrow request serving cached JSON bytes under `Content-Type: application/vnd.apache.arrow.stream`. One-character fix, worth folding in.

The rest is hardening and polish: the `Exec` gate leans on a non-local invariant in the HTTP layer, validation now costs a DuckDB round-trip on every cache hit, and authorization denials return HTTP 500. Details inline.

### Notes not worth a thread

- **README accuracy** — `preagg: { schema }` and `preagg: { enabled }` both check out against `PreAggregator.ts:13-16`, and `coordinator(mc)` is the right registration call. Good.
- **`server_test.go`** — `query.New` without `WithLogger` emits debug output through `slog.Default()`; consider matching `setupTestDB`'s `slog.LevelError` handler.
- **`baseTableValidator.Validate`** (untouched here) appends two errors for an unauthorized empty-schema table: the `else` branch appends, then falls through to the `slices.Contains` check, which appends again. Cosmetic — wants a `continue`.

packages/server/duckdb-server-go/internal/query/validator.go
derekperkins: Confirmed and fixed in fdf7d63e. The validator now checks both FUNCTION and WINDOW AST classes without relying on the varying window `type`. Regression cases cover both `sum(...) OVER ()` and `row_number() OVER ()`.

packages/server/duckdb-server-go/README.md
derekperkins: Agreed on both points. 0c0b4b18 now documents that the blocklist covers explicit calls only, explains replacement scans and the shared-process/filesystem boundary, and clarifies that every `Coordinator.exec(...)` call is rejected. I chose the explicit caveat over rejecting all unqualified tables under a function blocklist, because that would broaden a named-function policy into a table policy and break ordinary unqualified table queries. The policy-mode coordinator example now also sets `preagg.enabled: false`.

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Confirmed and fixed in 2cf81f3c. Arrow now uses the `"a"` cache discriminator. A same-DB regression test proves the first JSON and Arrow requests miss independently and later requests hit their own format-specific bytes.

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Agreed on removing the `allowedSchemas` coupling; fixed in 2946906a with policy ownership kept at the server layer. `execCommand` now gates directly on the server-lifetime `schemaMatchHeaders`, while `DB.Exec` enforces its own function policy and no longer accepts `allowedSchemas`. This avoids duplicating the schema-policy flag across Server and DB, where the copies could drift. The test covers configured and unconfigured server policy.

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Agreed that validating cache hits adds a parser round-trip. I’m keeping the correctness-first ordering in this PR: a verdict cache adds a second cache with capacity/TTL, policy-key, and failure-caching decisions that should be benchmarked rather than folded into this security fix. The result cache still avoids executing and materializing the user query. I’m treating verdict caching as follow-up performance work.

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Fixed in 9428940c. The wrapper is now `validation failed: %w`, and access denials use a typed sentinel, so the duplicated `query: validation failed for query: query: ...` wording is gone.

packages/server/duckdb-server-go/internal/server/server.go
derekperkins: Fixed in 9428940c. Typed schema/function access denials now return HTTP 403, `ErrExecWithValidation` returns HTTP 400, and parser/database/internal failures remain 500. Handler-level regression tests cover function denial, schema denial, restricted exec, and missing schema headers.

packages/server/duckdb-server-go/internal/query/options.go
derekperkins: Fixed the header-name whitespace case in 90c90567 by trimming each configured name at lookup; the regression test covers spaces/tabs and verifies a whitespace-only configured entry still yields no schema (so policy remains fail-closed). The blocklist option’s clearing behavior is intentional last-option-wins option composition; the production path has one call.

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Kept `WriteJSON` and `WriteArrow` as the validated streaming entry points to avoid removing exported API. In 0c0b4b18 the private execution helpers now carry explicit SECURITY comments requiring successful validation for the same query and request-scoped allowed schemas before use.

packages/server/duckdb-server-go/internal/server/server.go
```diff
@@ -203,7 +203,14 @@ func (s *Server) handleHTTP(w http.ResponseWriter, r *http.Request) {
 
 	data, cacheHit, err := s.execCommand(r.Context(), params, allowedSchemas)
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		status := http.StatusInternalServerError
```
derekperkins: The mapping is right for the two cases I raised. Following on from it: this default 500 now catches a third class that is also client-caused. `json_serialize_sql` only handles `SELECT`, so under any policy every non-`SELECT` statement fails validation and lands here.

Measured on `0c0b4b18`, identical request, `"type":"json"`:

| SQL | unrestricted | with `--function-blocklist md5` |
|---|---|---|
| `SELECT 1` | 200 | 200 |
| `SHOW TABLES` | 200 | 200 |
| `DESCRIBE SELECT 1 AS a` | 200 | 200 |
| `PRAGMA version` | 200 | **500** |
| `SET threads=1` | 200 | **500** |

```
validation failed: query: not implemented () at : Only SELECT statements can be serialized to json!
```

Two things:

**1. Status class.** This is the same "not supported in this mode" condition as `ErrExecWithValidation`, which you map to 400 one case below. As a 500 it means client retry logic will retry a query that can never succeed, and 5xx alerting fires on correctly-refused traffic. Detecting it needs a sentinel, since today it arrives as a bare `ErrorDetails` carrying DuckDB's message — e.g. wrap `ErrorDetails` with an `ErrUnsupportedStatement` sentinel when `error_type == "not implemented"`, then map that to 400 (or 501) here.

**2. It's undocumented.** The README note says `exec` requests are rejected, which reads as "DDL/DML is off in this mode." It doesn't say that `json`/`arrow` requests are *also* narrowed to statements the serializer accepts. The boundary isn't guessable either: `SHOW` and `DESCRIBE` survive because they serialize to a `SELECT_NODE` with a `SHOW_REF` `from_table`, while `PRAGMA` and `SET` don't. One sentence in the same note would cover it.

Minor, same error path: `ErrorDetails.Error()` formats `Type (Subtype) at Position` unconditionally, which is where the empty `()` and trailing `at :` above come from — this response carries neither field. That's the sparse-response case `stringField` was added for, now visible in the message rather than as a panic. Skipping empty fields would tidy it.

packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -193,19 +205,19 @@ func (v *baseTableValidator) Validate() []error {
...
 				continue // empty schemas are allowed if they are CTEs
 			} else {
-				errs = append(errs, fmt.Errorf("access denied: unauthorized access to table '%v' with empty schema", baseTable.TableName))
+				errs = append(errs, fmt.Errorf("%w: unauthorized access to table '%v' with empty schema", ErrAccessDenied, baseTable.TableName))
 			}
 		}
 
 		if !slices.Contains(v.allowedSchemas, baseTable.SchemaName) {
-			errs = append(errs, fmt.Errorf("access denied: unauthorized access to schema '%v'", baseTable))
+			errs = append(errs, fmt.Errorf("%w: unauthorized access to schema '%v'", ErrAccessDenied, baseTable))
```
derekperkins: Now that denials return 403 with the error text as the response body, two formatting issues here became client-facing.

**`%v` on the struct.** `baseTable` is a `tableRef`, so the body renders Go struct syntax:

```
403 validation failed: query: access denied: unauthorized access to schema '{tenant_b secret false}'
```

The trailing `false` is `IsCTE`. Presumably wants `baseTable.SchemaName`, or `'%s.%s'` with schema and table.

**Duplicate error for unqualified tables.** The `SchemaName == ""` branch appends at line 208, then falls through to this `slices.Contains` check, which appends again — `""` is never in `allowedSchemas`. Both land in the joined body:

```
403 validation failed: query: access denied: unauthorized access to table 'unqualified_table' with empty schema
    query: access denied: unauthorized access to schema '{ unqualified_table false}'
```

A `continue` after the line 208 append fixes it and takes the empty-schema struct dump with it.

Both predate this PR. Flagging them because this is the error path a misconfigured tenant hits most often, and it's now the text they see rather than something in the server log.

derekperkins (commented):
## Re-review — 6 new commits (`db4c4600` → `0c0b4b18`)

Checked out `0c0b4b18` in a clean worktree and verified each reply against the code rather than the description. `go test -tags=duckdb_arrow ./... -count=1`, `go test -race -tags=duckdb_arrow ./...`, and `go vet` are all clean.

**Nothing unresolved.** All nine threads hold up. Several needed an actual run rather than a read, so here's what I confirmed:

- **Window-form blocklist bypass** (`fdf7d63e`) — blocking `sum`, `row_number`, `rank`, `lead`, `first_value`, or `quantile_cont` now returns 403 for each in `... OVER ()` form. Dropping the `type` check is safe: `class: "FUNCTION"` only ever pairs with `type: "FUNCTION"`, so the check was pure redundancy.
- **Arrow/JSON cache collision** (`2cf81f3c`) — same `DB`, `persist: true`: the json request returns JSON bytes and the arrow request returns Arrow IPC (`\xff\xff\xff\xff` magic), both stable across repeat hits. `TestDB_CacheSeparatesFormats` asserting `json.Valid(arrowData)` is `false` is a nice way to pin the format rather than just the byte count.
- **`exec` gated on a per-request value** (`2946906a`) — agreed this is better than what I suggested. Policy now sits with the config that owns it, and `DB` doesn't hold a second copy that could drift. `exec` → 400 under `--schema-match-headers` (server layer) and under `--function-blocklist` (DB layer); both HTTP and WebSocket route through `execCommand`, so neither transport is missed.
- **Denials as HTTP 500** (`9428940c`) — function denial 403, schema denial 403, restricted `exec` 400, missing header 401. `errors.Is` does traverse the `errors.Join` → `fmt.Errorf("%w")` chain, which was the part I couldn't tell by reading.
- **Header-name whitespace** (`90c90567`) — `--schema-match-headers=" X-Tenant-Id "` now resolves; a whitespace-only entry yields no schema, so still fail-closed.
- **Doubled `query:` prefix** (`9428940c`) — now reads `validation failed: query: access denied: ...`.
- **Replacement scans / `exec` breadth** (`0c0b4b18`) — confirmed the new README guidance actually works: under `--schema-match-headers`, `SELECT * FROM 'data.parquet'` is rejected 403 as an unqualified table reference. Agreed on choosing the caveat over rejecting all unqualified tables — the enforcement version would break ordinary unqualified queries, and the note now says plainly that the blocklist covers explicit calls only and isn't a sandbox.
- **`write*` invariant** (`0c0b4b18`) — the SECURITY comments are the version of this I was after.

**Verdict caching** (`query.go:226`) is the one item deliberately not done, and I agree with keeping it out of a security fix. Worth an issue so resolving the thread doesn't lose it.

### Two new findings

Both predate this PR, but the 403/400 mapping just promoted them from log noise to client-visible response bodies, so they're worth a look while you're in this code. Neither blocks merge — threads below.

- **Restricted mode turns non-`SELECT` statements into 500s.** `PRAGMA version` and `SET threads=1` return 200 unrestricted and 500 under any policy, via the same "not supported in this mode" condition you correctly map to 400 for `exec`.
- **Denial bodies leak Go struct formatting, and duplicate for unqualified tables.** `unauthorized access to schema '{tenant_b secret false}'` is now a 403 body.

packages/server/duckdb-server-go/internal/server/server.go
derekperkins: Agreed and fixed in ea1b4af7. `ErrorDetails` now classifies DuckDB `not implemented` serializer failures with `ErrUnsupportedStatement`; HTTP maps them to 400, sparse error rendering omits empty subtype/position fields, JSON and Arrow are covered, and the README documents the serializable-statement boundary (including the HTTP/WebSocket distinction). While testing the serializable `SHOW` boundary, I also found a direct schema-metadata bypass: `SHOW_REF` nodes were not validated. 5c9a0da6 now allows explicitly authorized `SHOW ... FROM schema`, rejects unauthorized or schema-less metadata enumeration, and leaves nested `DESCRIBE` queries to the existing base-table checks. Full-statement authorization remains follow-up work.

packages/server/duckdb-server-go/internal/query/validator.go
derekperkins: Agreed and fixed in f2de1e43. Schema denials now format only `baseTable.SchemaName`, and the empty-schema branch stops after its specific denial instead of falling through. Exact regression tests pin a single client-facing error and preserve `ErrAccessDenied`: `unauthorized access to schema 'tenant_b'` or `unauthorized access to table 'secret' with empty schema`.

derekperkins (commented):
## Re-review — 3 new commits (`0c0b4b18` → `5c9a0da6`)

Verified at `5c9a0da6` in a clean worktree. `go test -tags=duckdb_arrow ./... -count=1`, `-race`, and `go vet` all clean.

**Both open threads resolved.** Verified rather than read:

- **Unsupported statements** (`ea1b4af7`) — `PRAGMA version` and `SET threads=1` now return 400 on both `json` and `arrow`, and the message reads `query: not implemented: Only SELECT statements can be serialized to json!` with the empty `()` and trailing `at :` gone. The `ErrorDetails.Is` hook is a tidy way to do this without a wrapper type, and `errors.Is` does reach it through the `fmt.Errorf("%w")` chain. `SHOW`/`DESCRIBE`/`SELECT` still 200.
- **Denial formatting** (`f2de1e43`) — one line per denial, schema-only: `unauthorized access to schema 'tenant_b'` and `unauthorized access to table 'unqualified' with empty schema`.

`f2de1e43` also fixed something neither of us named: `Validate()` used to start from `var errs []error` and never returned `v.errs`, so every error appended by `handleBaseTable` — the `schema_name`/`table_name` type-assertion failures — was silently discarded, and a non-string `table_name` would have passed validation. `append([]error(nil), v.errs...)` closes that, and it's what makes `handleShowRef`'s errors reachable at all.

### The SHOW_REF find

Good catch, and it holds up. Under `--schema-match-headers` with `tenant_a` authorized:

| Query | Result |
|---|---|
| `SHOW TABLES` | 403 requires an explicit authorized schema |
| `SHOW ALL TABLES` | 403 |
| `SHOW TABLES FROM tenant_a` | 200 |
| `SHOW TABLES FROM tenant_b` | 403 |
| `DESCRIBE tenant_a.t` | 200 |
| `DESCRIBE tenant_b.secret` | 403 |
| `SUMMARIZE tenant_b.secret` | 403 |

The `query`-present early return is the right call — `SUMMARIZE` and `DESCRIBE <table>` both nest a `SELECT_NODE` whose `BASE_TABLE` carries the schema, so they fall through to the existing check instead of needing a second one.

### Two more gaps in the same boundary

Chasing that fix turned up two other places the schema policy doesn't hold, both the same shape: a node field the validator doesn't read. Threads below.

- **Metadata enumeration routes around the SHOW fix.** `duckdb_tables()`, `duckdb_schemas()`, `duckdb_columns()`, and `pragma_table_info('tenant_b.secret')` all return 200 and expose other tenants' schemas. They're `TABLE_FUNCTION` nodes, so there's no `BASE_TABLE` or `SHOW_REF` to check.
- **The policy isn't catalog-qualified.** With a second catalog attached, `tenant_a` reads `otherdb.tenant_a.secret` — row data, not metadata. `catalog_name` sits in the AST next to the fields that are read.

Neither is reachable by a tenant who can't already get `exec`, since `ATTACH` and DDL are both rejected under policy — so how much they matter depends on what the operator attaches and loads. Given #947 is where full-statement authorization lands, these may belong there. But this PR's SHOW fix sets the expectation that metadata enumeration is closed, and right now it's closed for one syntax out of several.

One thing left as-is and fine: a SQL syntax error under policy still returns 500 (`validation failed: query: parser (SYNTAX_ERROR) at 12: ...`). The unrestricted path 500s on the same input from the executor, so it's consistent — just noting it's the one client-caused error still in the 5xx bucket.

packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -157,6 +196,29 @@ func (v *baseTableValidator) CheckNode(node map[string]any, keyStack []string) {
...
 
+func (v *baseTableValidator) handleShowRef(showRef map[string]any) {
+	// DESCRIBE statements contain a nested query whose base tables are validated separately.
+	if _, exists := showRef["query"]; exists {
+		return
+	}
+
+	schemaName, exists := showRef["schema_name"]
+	if !exists {
+		v.errs = append(v.errs, fmt.Errorf("%w: SHOW statement requires an explicit authorized schema", ErrAccessDenied))
```
derekperkins: This closes the `SHOW` syntax, but the function equivalents reach the same catalog metadata and produce no `SHOW_REF` or `BASE_TABLE` node to check.

Measured at `5c9a0da6`, `--schema-match-headers=X-Tenant-Id`, header `tenant_a`, with `tenant_b.secret` existing:

| Query | Result |
|---|---|
| `SHOW TABLES` | 403 |
| `SELECT * FROM duckdb_tables()` | **200** — every table in every schema |
| `SELECT * FROM duckdb_schemas()` | **200** |
| `SELECT * FROM duckdb_columns()` | **200** — includes `tenant_b.secret.i` |
| `SELECT * FROM pragma_table_info('tenant_b.secret')` | **200** — names the table directly |

These serialize as `TABLE_FUNCTION` wrapping a `class: "FUNCTION"` node, so `baseTableValidator` never sees a table reference. `information_schema.tables` *is* caught, because that one is a genuine schema-qualified base table.

Scope: names and column types only — row data still needs a `BASE_TABLE` reference, which is checked. So this is the same disclosure class as the `SHOW TABLES` case you just closed, not an escalation past it. The inconsistency is the part worth flagging: after this PR a tenant hitting `SHOW TABLES` gets a 403 telling them enumeration is blocked, and `duckdb_tables()` hands them the list anyway.

Two ways to scope it:

- Fold the `duckdb_*` metadata functions and `pragma_table_info` into a default blocklist implied by `--schema-match-headers`. One list, but it's a denylist and DuckDB keeps adding functions.
- Leave it to #947 and note it in the README instead — "metadata enumeration via `duckdb_*` functions is not restricted; add them to `--function-blocklist` if that matters" — so operators can close it themselves today.

I'd lean to the second, since a real allowlist belongs with full-statement authorization rather than here.

packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -157,6 +196,29 @@ func (v *baseTableValidator) CheckNode(node map[string]any, keyStack []string) {
...
+	}
+	schemaNameStr = strings.TrimPrefix(schemaNameStr, "schema_name:")
+	if !slices.Contains(v.allowedSchemas, schemaNameStr) {
+		v.errs = append(v.errs, fmt.Errorf("%w: unauthorized access to schema '%s'", ErrAccessDenied, schemaNameStr))
+	}
+}
+
 func (v *baseTableValidator) handleBaseTable(baseTable map[string]any) {
 	var schemaNameStr string
 	schemaName, exists := baseTable["schema_name"]
```
derekperkins, line 162: This reads `schema_name` and ignores `catalog_name`, which is present on the same node. The authorization key ends up being a bare schema name, unqualified by catalog.

Reproduced at `5c9a0da6` — `--schema-match-headers=X-Tenant-Id`, header `tenant_a`, with a second catalog attached at startup:

```sql
ATTACH 'other.db' AS otherdb;
CREATE SCHEMA otherdb.tenant_a;
CREATE TABLE otherdb.tenant_a.secret (v VARCHAR);
INSERT INTO otherdb.tenant_a.secret VALUES ('other-catalog-data');
```

| Query | Result |
|---|---|
| `SELECT * FROM tenant_a.mine` | 200 |
| `SELECT * FROM otherdb.tenant_a.secret` | **200** — `[{"v":"other-catalog-data"}]` |
| `SHOW TABLES FROM otherdb.tenant_a` | **200** |

Unlike the metadata thread, this is row data. The field is right alongside the two you do read:

```json
{"type":"BASE_TABLE","schema_name":"tenant_a","table_name":"secret","catalog_name":"otherdb"}
```

`handleShowRef` has the same gap, which is the third row.

A tenant can't set this up themselves — `ATTACH` is a non-`SELECT` statement (400 under policy) and `exec` is rejected — so it needs the operator to have attached a catalog. But `--load-extensions` is a documented flag and `mysql_scanner` / `postgres_scanner` / `sqlite_scanner` exist to do exactly that, and the schema names inside those catalogs aren't under the operator's tenant-naming control. Two catalogs plus one schema-name collision is the whole requirement.

Cheapest fail-closed fix if you want it here: deny when `catalog_name` is present and isn't the server's own database. Otherwise it's worth an explicit README precondition — that the tenant boundary assumes a single catalog — so the assumption is stated rather than implied.

packages/server/duckdb-server-go/internal/query/validator.go
derekperkins: Agreed with the documentation/#947 scope. ac7121f6 now states that schema matching does not hide metadata returned by metadata functions, that deployments must add the exact exposed function names to --function-blocklist, that duckdb_* is not a supported wildcard, and that the list must be reviewed as DuckDB or its extensions change. I am not adding an implied default denylist here because it would be incomplete and version-dependent, which could overstate the isolation this policy provides.

packages/server/duckdb-server-go/internal/query/validator.go
derekperkins: Confirmed and fixed in e8b3abd7. The schema validator now fails closed on any explicit catalog qualifier in BASE_TABLE and SHOW_REF nodes, and I extended the same check to catalog-qualified function and window nodes found during the audit. Regression tests cover SELECT, SHOW, DESCRIBE, table-function, and window-function forms. I chose the stricter rule instead of special-casing the server catalog: resolving that catalog robustly would add state and still leave DuckDB two-part-name ambiguity. ac7121f6 therefore also states the policy assumes a single catalog and attached catalogs are outside its boundary.

derekperkins (commented):
## Re-review — 2 new commits (`5c9a0da6` → `ac7121f6`)

Verified at `ac7121f6` in a clean worktree. `go test -tags=duckdb_arrow ./... -count=1`, `-race`, and `go vet` all clean. **Both threads resolved.** Nothing further from me — details below.

### Catalog qualifier (`e8b3abd7`)

Fixed, and the strictness doesn't cost anything. With a second catalog attached and `tenant_a` authorized:

| Query | Before | Now |
|---|---|---|
| `SELECT * FROM otherdb.tenant_a.secret` | 200, returned row data | **403** |
| `SHOW TABLES FROM otherdb.tenant_a` | 200 | **403** |
| `DESCRIBE otherdb.tenant_a.secret` | — | **403** |
| `SELECT otherdb.main.md5('x')` | — | **403** |
| `SELECT * FROM otherdb.main.range(3)` | — | **403** |
| `SELECT otherdb.main.sum(i) OVER () ...` | — | **403** |
| `SELECT * FROM memory.tenant_a.t` | — | **403** |
| `SELECT * FROM tenant_a.t` | 200 | 200 |

The thing I wanted to check was false positives, since `rejectCatalogReference` fires on mere presence of the field and now runs against every `FUNCTION`/`WINDOW` node in the tree. **Zero false positives across 22 realistic shapes** — `count(*)`, `sum`/`avg`/`min`/`max`, `sum() OVER (ORDER BY)`, `row_number() OVER (PARTITION BY)`, `quantile_cont`, `histogram`, `string_agg`, `md5`/`upper`/`length`, `::` and `CAST`, `date_trunc`/`strftime`/`epoch_ms`/`extract`, list and struct literals, `list_transform` with a lambda, `unnest`, CTEs, self-joins, `UNION`, `IN (subquery)`, `main.md5(...)`, `DESCRIBE SELECT`, `SHOW TABLES FROM tenant_a`, `ORDER BY`/`LIMIT`.

That holds because DuckDB only emits `catalog` when the user writes an explicit qualifier: plain `md5(...)` and `count(*)` carry neither `catalog` nor `schema`, and `main.md5(...)` gets `schema: "main"` with no `catalog`. Worth knowing that the field is serialized *last* in the object — I misread it as absent on my first pass at a truncated AST, so anyone auditing this later should dump the whole node.

Agree with rejecting the server's own catalog rather than trying to permit it. Two-part names in DuckDB are genuinely ambiguous between `catalog.schema` and `schema.table`, so a permit-my-catalog rule would have to disambiguate against live catalog state at validation time — more state for no boundary gain, and the README now states the single-catalog assumption outright.

### Metadata functions (`ac7121f6`)

Documented, which is the option I recommended. I checked that the documented remedy actually works rather than taking the note at face value — with `duckdb_tables,duckdb_schemas,duckdb_columns,pragma_table_info` in `--function-blocklist`, all four return 403. And `duckdb_views()`, which I deliberately left out of that list, returns 200. That's exactly the sharp edge the README calls out: no wildcards, and the list has to be maintained. So the docs are accurate about both the mitigation and its limits, which is what makes this fine to leave to #947.

Agree with declining a default denylist. An incomplete built-in list would read as a guarantee, and that's worse than a documented gap an operator can close.

### Where this landed

Thirteen threads over four rounds, and the substantive ones were all real: the blocklist was inert, cached results skipped authorization, `exec` had no policy, window-form calls escaped the blocklist, Arrow and JSON collided in the cache, `SHOW_REF` was unvalidated, and the schema check ignored `catalog_name`. Two of those you found yourself while testing fixes for the others, which is the part that makes me trust the rest.

Residual items, all deliberate and documented rather than unknown:

- Verdict caching — deferred perf work; worth an issue so it isn't lost with the resolved thread.
- Full-statement authorization for `exec` and non-`SELECT` — #947.
- Metadata enumeration — operator-closable via `--function-blocklist`, documented.
- A SQL syntax error under policy still returns 500. The unrestricted path 500s on the same input from the executor, so it's consistent; only worth a line if you're already touching the status mapping.

Nothing blocking from my side.

packages/server/duckdb-server-go/internal/query/query.go
```diff
@@ -19,6 +20,9 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+// ErrExecWithValidation is returned for exec requests while schema or function validation is active.
```
domoritz: I usually avoid comments that just explain code that's already descriptive

domoritz (approved):

packages/server/duckdb-server-go/internal/query/query.go
derekperkins: Agreed. Removed the comments above ErrExecWithValidation, ErrAccessDenied, and ErrUnsupportedStatement in 17e16f9b; the error names and messages are descriptive on their own.

packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -250,7 +328,7 @@ func (v *functionBlocklistValidator) CheckNode(node map[string]any, keyStack []s
 	}
```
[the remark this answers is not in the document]
derekperkins: Fixed in 7453e7e7. The validator now lowercases function_name before the blocklist lookup, with regression coverage using an uppercase AST value. DuckDB currently emits lowercase names, but the authorization check no longer depends on serializer normalization.

packages/server/duckdb-server-go/internal/server/server.go
```diff
@@ -203,7 +203,14 @@ func (s *Server) handleHTTP(w http.ResponseWriter, r *http.Request) {
...
+		switch {
+		case errors.Is(err, query.ErrExecWithValidation), errors.Is(err, query.ErrUnsupportedStatement):
+			status = http.StatusBadRequest
+		case errors.Is(err, query.ErrAccessDenied):
+			status = http.StatusForbidden
+		}
+		http.Error(w, err.Error(), status)
```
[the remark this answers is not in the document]
derekperkins: Fixed in fb0434e0. handleHTTP now recognizes wrapped query.ErrorDetails values with errors.As and returns 400 Bad Request. The regression test exercises a syntax error while policy validation is active.

derekperkins:
@domoritz FYI, I wanted to keep the individual commits, so I converted this PR into a stacked PR, since this repo has a squash only policy for PRs

domoritz:
I think it still makes sense to squash into main since we have the individual commits here but I don't have strong feelings.

derekperkins:
I find it very helpful looking at when things changed and what things changed with them. This is a backlog of changes that have built up since I last came in and touched the go server that in a perfect world would've happened one by one over time, and would've naturally had their own commits.

PR #1119, opened 2026-08-03
fix(go-server): harden schema and function authorization

derekperkins:
## Stack

1. #1106 — fix(go-server): enforce query policies
2. #1117 — fix(go-server): make cached responses policy- and format-safe
3. #1118 — fix(go-server): classify query and request failures
**4. #1119 — fix(go-server): harden schema and function authorization**
5. #1120 — fix(go-server): harden SQL serializer validation
6. #1121 — docs(go-server): document query policy boundaries

This is **4 of 6**. Merge bottom-up.

## What

- validate schema references in `SHOW` and `DESCRIBE`
- reject explicitly catalog-qualified tables, functions, and metadata statements
- normalize extracted function names before blocklist matching
- trim configured schema-header names

## Why

These DuckDB AST forms are represented differently from ordinary base-table references and could otherwise fall outside the existing authorization checks.

## Impact

`SHOW ALL TABLES` is rejected under schema policy, `SHOW` requires an explicit authorized schema, and attached-catalog references fail closed.

## Verification

- `golangci-lint run`

Part of #958

### DIFF
--- a/packages/server/duckdb-server-go/internal/query/validator.go
+++ b/packages/server/duckdb-server-go/internal/query/validator.go
@@ -173,9 +173,18 @@ func newBaseTableValidator(allowedSchemas []string) Validator {
 }
 
 func (v *baseTableValidator) CheckNode(node map[string]any, keyStack []string) {
+	if class, exists := node["class"]; exists && (class == "FUNCTION" || class == "WINDOW") {
+		v.rejectCatalogReference(node, "catalog")
+	}
+
 	val, exists := node["type"]
-	if exists && val == "BASE_TABLE" {
-		v.handleBaseTable(node)
+	if exists {
+		switch val {
+		case "BASE_TABLE":
+			v.handleBaseTable(node)
+		case "SHOW_REF":
+			v.handleShowRef(node)
+		}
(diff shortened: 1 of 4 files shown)


packages/server/duckdb-server-go/internal/query/validator.go
```diff
@@ -189,7 +198,38 @@ func (v *baseTableValidator) CheckNode(node map[string]any, keyStack []string) {
...
+		v.errs = append(v.errs, fmt.Errorf("%w: SHOW statement requires an explicit authorized schema", ErrAccessDenied))
+		return
+	}
+
+	schemaNameStr, ok := schemaName.(string)
+	if !ok {
+		v.errs = append(v.errs, fmt.Errorf("invalid 'schema_name' in show reference, expected string: %v", schemaName))
+		return
+	}
+	schemaNameStr = strings.TrimPrefix(schemaNameStr, "schema_name:")
```
domoritz, line 222: This still feels brittle to me. Are we giving the wrong impression that the code is safe for production use cases?

domoritz (approved):

PR #1130, opened 2026-08-08
feat(go-server): extract reusable extension initialization

### LINKED ISSUE #1125: go: extract reusable DuckDB extension initialization
## Summary

Extract the DuckDB extension setup currently embedded in `main.go` into a small reusable package, tentatively `pkg/extension`.

The example binary should continue accepting `--load-extensions` with the same `name|repository,name2` syntax, while other projects can use structured extension specifications without depending on flags, string pointers, or the example logger.

## Why

The current private `extensionLoader` is useful outside the example binary, but it:

- accepts CLI-shaped input rather than a programmatic model;
- has no tests;
- logs and returns errors, which is undesirable library behavior;
- interpolates extension names directly into `INSTALL`/`LOAD` SQL;
- quotes custom repositories without escaping embedded quotes; and
- is used as a DuckDB connector initializer, so it runs for every newly created connection.

## Scope

- Add a structured extension specification and a connector-compatible initializer under `pkg/extension` (exact names can be chosen during implementation).
- Add a parser for the existing comma-separated CLI syntax, but keep the core loader independent of `flag`, pointers, and `slog`.
- Preserve the default `core` repository and the currently supported built-in repository names (`core`, `core_nightly`, `community`, `local_build_debug`, and `local_build_release`).
- Validate or safely quote extension identifiers and custom repository URLs/paths. Reject malformed, blank, or ambiguous specifications rather than constructing unsafe SQL.
- Return contextual errors that identify the extension, repository, and `INSTALL` or `LOAD` phase. Do not log inside the library.
- Explicitly define the per-connection lifecycle: whether `INSTALL` is intentionally repeated, how `LOAD` is applied to every connection, and what happens when a later connection cannot initialize.
- Rewire `main.go` to use the package without changing the `--load-extensions` contract.
- Document that extensions are trusted native code and that a requested extension failing to install/load fails startup or connection creation closed.

Keep flag parsing, TLS certificate discovery, process logging, and HTTP serving in the example binary. They are example-specific behavior, not reusable library APIs.

## Acceptance criteria

- [ ] Table tests cover empty input, whitespace, multiple extensions, default/core/community/nightly repositories, custom URL/path repositories, extra delimiters, blank names, and quote/injection inputs.
- [ ] Tests use a fake `driver.ExecerContext` to verify ordered `INSTALL` then `LOAD`, context propagation, and stopping at the first error without network access or extension downloads.
- [ ] Tests invoke the initializer more than once and lock in the documented multi-connection behavior.
- [ ] The library returns errors and performs no logging.
- [ ] The README includes a small programmatic example in addition to the unchanged CLI example.
- [ ] From `packages/server/duckdb-server-go`, `go test -tags=duckdb_arrow ./...` and `golangci-lint run` pass.

## Sequencing

Depends on #1123 and can proceed in parallel with #1124 and #1126. Coordinate any move of `query.GetExtensions`/`query.Extension` with #1124 rather than changing that surface independently in both branches.

derekperkins:
Closes #1125

## Summary

- add a reusable `pkg/extension` package with structured extension specifications and a connector-compatible initializer
- preserve the existing comma/pipe CLI grammar while also accepting a variadic list of strings
- validate extension identifiers, safely quote custom repositories and local paths, and return contextual `INSTALL`/`LOAD` errors without library logging
- support install-and-load, preinstalled load-only, local repository, and standalone extension-file workflows
- keep initialization stateless and repeat the applicable ordered plan for every new DuckDB connection
- rewire the example binary without changing `--load-extensions`, and document startup preflight, context lifetime, fail-closed behavior, and native-code trust

## Why

The existing private loader was coupled to flag pointers and the example logger, interpolated unvalidated identifiers, and did not safely escape custom repository values. It also left the per-connection lifecycle implicit. The extracted package gives embedding applications a validated programmatic API while retaining the example server's behavior.

## Validation

- `go test ./pkg/extension -count=1`
- `go test -race ./pkg/extension -count=1`
- `go build -tags=duckdb_arrow ./...`
- `golangci-lint run`
- `git diff --check`


packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+	commands := make([]command, 0, len(specs)*2)
+	seen := make(map[string]int, len(specs))
+
+	for index, spec := range specs {
+		compiled, key, err := compileSpec(spec)
+		if err != nil {
+			return nil, fmt.Errorf("extension specification %d: %w", index+1, err)
+		}
+
+		if previous, ok := seen[key]; ok {
```
derekperkins: **Exact duplicates are now a startup failure — a `--load-extensions` behavior change.**

`--load-extensions httpfs,httpfs` works today (both `INSTALL` and `LOAD` are idempotent). On this branch it refuses to start:

```
extension specification 2 duplicates specification 1 target "httpfs"
```

Rejecting *conflicting* duplicates is clearly right — `h3|core,h3|community` is genuinely ambiguous about which repository wins. Rejecting an identical repeat is stricter than it needs to be, and it is the case most likely to show up from templated or concatenated config (a Helm chart appending to a default list, an env var built from two sources).

Consider collapsing exact duplicates and erroring only when the mode or repository differs. If the strict behavior is intended, the flag help and README should call it out, since it is the one visible break for existing deployments and the PR description says the flag contract is preserved.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+	load.sql = "LOAD " + pathSQL
+	load.phase = "LOAD"
+
+	key := "path:" + spec.Path
+	if spec.Mode == ModeLoadOnly {
+		return []command{load}, key, nil
+	}
+
+	install := metadata
+	install.sql = "INSTALL " + pathSQL
```
derekperkins: **`InstallAndLoadFile` emits an `INSTALL` that the paired `LOAD` never consults.**

Both statements target the path:

```sql
INSTALL '/opt/custom_writer.duckdb_extension'
LOAD '/opt/custom_writer.duckdb_extension'
```

DuckDB's path-form `LOAD` reads that file directly rather than going through the extension directory, so this `INSTALL` does not feed the `LOAD` on the next line. Its only effect is the side effect of copying the binary into the extension directory under its derived name — which matters solely to some *other* connection doing `LOAD <name>`.

The canonical DuckDB pairing is `INSTALL '<path>'` followed by `LOAD <name>`, but `Spec` has no `Name` when `Path` is set, so the package cannot express it. Worth confirming against DuckDB's actual behavior, then either deriving the name from the filename for the `LOAD`, or documenting that `InstallAndLoadFile` differs from `LoadFile` only by persisting a copy. Right now the README lists the two side by side as if the difference were about how the extension gets loaded.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+	return specs, nil
+}
+
+// NewInitializer validates specs and returns an immutable, reentrant
+// connection initializer. ctx is used for every INSTALL and LOAD statement;
+// callers should therefore provide a context whose lifetime matches the
+// connector rather than an individual connection attempt.
+func NewInitializer(ctx context.Context, specs ...Spec) (Initializer, error) {
+	if ctx == nil {
+		return nil, errors.New("extension: nil context")
```
derekperkins: **No exported sentinels, which is the opposite of the sibling package's convention.**

`pkg/query` exports `ErrAccessDenied`, `ErrUnsupportedStatement`, and `ErrExecWithValidation` so callers can branch with `errors.Is`. This package returns only `errors.New`/`fmt.Errorf` values, so an embedding application cannot distinguish "the operator gave me a bad spec" from "DuckDB could not reach the repository" without string matching. That distinction has teeth: the first is fatal config, the second is worth retrying.

Consider exporting something like `ErrInvalidSpec` (wrapped by the `compile`/`compileSpec` paths) and `ErrInstall`/`ErrLoad` (wrapped by `command.wrap`).

Minor and related: the prefix style is inconsistent. This line uses the `query: `-style `extension: nil context`, but the `compile` errors render as `extension specification 1: ...` with no colon-prefixed package name.

packages/server/duckdb-server-go/pkg/extension/extension_test.go
```diff
@@ -0,0 +1,491 @@
...
+		{name: "single quoted repository", values: []string{"spatial|'core'"}},
+		{name: "double quoted repository", values: []string{"spatial|\"core\""}},
+		{name: "repository control character", values: []string{"spatial|https://example.test/\nrepo"}},
+		{name: "duplicate name", values: []string{"spatial,SPATIAL"}},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			if _, err := extension.Parse(test.values...); err == nil {
```
derekperkins: **Negative tables assert only `err != nil`, so a case can pass for the wrong reason.**

This table and the one in `TestNewInitializerRejectsInvalidSpecifications` never inspect the error, so a case can start failing on a different branch than its name claims and stay green. For example `{name: "blank path", specs: []extension.Spec{{Path: "  "}}}` is meant to exercise `compileFile`'s `path is blank` check, but it would pass just as happily if `validateValue` began rejecting it, or if `compileSpec`'s name/path exclusivity check changed shape.

`TestInitializerStopsAtFirstError` already handles this well with its `wantMessage` substrings. Extending the same pattern to these two tables — or `errors.Is` against exported sentinels, per the other thread — would pin down which validation actually fired.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+			(char >= '0' && char <= '9') ||
+			char == '_' {
+			continue
+		}
+		return false
+	}
+	return value != ""
+}
+
+func isBuiltIn(repository Repository) bool {
```
derekperkins: **Built-in repository names are matched case-sensitively.**

`x|Core` or `x|COMMUNITY` falls through to the custom-repository branch and gets quoted:

```sql
INSTALL x FROM 'Core'
```

which then fails at connection time with a DuckDB error that does not point at the capitalization.

The old loader had the same case sensitivity, so this is not a regression — but this PR adds the validation layer that makes catching it nearly free. Lowercasing before the switch (or erroring on a case-insensitive match to a built-in name) turns a confusing runtime failure into a startup one.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+					valueIndex+1,
+					entryIndex+1,
+				)
+			}
+
+			specs = append(specs, InstallAndLoad(name, Repository(repository)))
+		}
+	}
+
+	if _, err := compile(specs); err != nil {
```
derekperkins: **`Parse` reports errors in two different coordinate systems.**

Validating by compiling and discarding the result means `Parse` errors arrive in two vocabularies:

- its own checks: `extension input 2 entry 1 is blank`
- `compile`'s checks: `extension specification 3: name "sp-atial" is not a safe DuckDB identifier`

For the CLI path in `main.go` there is always exactly one input, so `input 1` is noise, and `specification 3` counts across all inputs rather than pointing at what the operator typed in the flag.

Since `compile` already returns positional errors, `Parse` could compile each entry as it appends and wrap the failure with its own `input i entry j` coordinates. That also drops the duplicate compile pass — `Parse` compiles here, then `NewInitializer` compiles the same specs again. The cost is negligible, but the two-pass shape is what forces the mismatched wording.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,408 @@
...
+		return []command{load}, key, nil
+	}
+
+	install := metadata
+	install.sql = "INSTALL " + pathSQL
+	install.phase = "INSTALL"
+	return []command{install, load}, key, nil
+}
+
+func validateValue(kind, value string) error {
```
derekperkins: **`Parse` trims surrounding whitespace; the programmatic API preserves it.**

`validateValue` rejects blank-after-trim but keeps surrounding whitespace in the value, so the two entry points disagree:

```
Parse("custom| /opt/repo ")                      -> INSTALL custom FROM '/opt/repo'
InstallAndLoad("x", Repository("  https://r  ")) -> INSTALL x FROM '  https://r  '
LoadFile("  /tmp/a.duckdb_extension  ")          -> LOAD '  /tmp/a.duckdb_extension  '
```

A caller building `Spec`s from YAML or env vars gets an opaque DuckDB error where `Parse` would have silently done the right thing.

Trailing whitespace is technically legal in a filesystem path, so trimming is not unambiguously correct — but the asymmetry should at least be a documented choice on the `Spec` fields rather than an emergent one.

packages/server/duckdb-server-go/main.go
```diff
@@ -13,6 +12,7 @@ import (
 
 	"github.com/duckdb/duckdb-go/v2"
 
+	"github.com/uwdata/mosaic/packages/server/duckdb-server-go/pkg/extension"
```
derekperkins: **This package name is shadowed later in the file.**

Line 137 declares `for _, extension := range extensions`, which shadows this import for the body of that loop. It compiles because the last package reference is on line 58, but it makes `extension.Name` on line 138 read like a package selector when it is a struct field.

The collision is introduced by this PR, so worth renaming the loop variable to `ext` here rather than leaving it for whoever next adds an `extension.` call below that point.

packages/server/duckdb-server-go/main.go
```diff
@@ -49,6 +49,18 @@ func main() {
 		Level: logLevel,
 	}))
 
+	extensionSpecs, err := extension.Parse(*extensionsStr)
+	if err != nil {
+		logger.Error("main: invalid load-extensions", "error", err, "load-extensions", *extensionsStr)
```
derekperkins: **Invalid config exits 0.**

Both new failure paths `return` from `main`, so a malformed `--load-extensions` produces a log line and a clean exit status. That matches the existing style in this file — every error path here does the same — so it is not a regression from this PR.

Flagging it because config validation at startup is exactly the case where an orchestrator wants a non-zero exit: systemd and Kubernetes both read a 0 as "did its job and stopped." If you would rather not convert every path in this file, a follow-up that routes `main` through a `run() error` helper would cover all of them at once.

derekperkins (commented):
## Review: extract reusable extension initialization

### What this does

Moves the private `extensionLoader` closure out of `main.go` into a new `pkg/extension` package with a structured `Spec` model, a CLI-grammar `Parse`, and a `NewInitializer` that precompiles an ordered `INSTALL`/`LOAD` plan into a `duckdb.NewConnector` initializer. `main.go` keeps flag parsing and logging; the library returns errors and logs nothing. The README gains a programmatic example plus notes on connector laziness, per-connection lifecycle, and native-code trust.

### Verification

Run against the PR head in a worktree:

| check | result |
| --- | --- |
| `go test ./pkg/extension -count=1 -race -cover` | pass, 96.5% of statements |
| `go build -tags=duckdb_arrow ./...` | pass |
| `golangci-lint run ./...` | 0 issues |
| `gofmt -l` | clean |

I also probed the generated SQL directly with a throwaway `driver.ExecerContext`. The injection handling holds up:

- `Repository("r'; ATTACH 'evil.db")` -> `INSTALL x FROM 'r''; ATTACH ''evil.db'` — one well-formed literal
- ``Repository(`C:\repo\'`)`` -> ``INSTALL x FROM 'C:\repo\'''`` — correct, since backslash is not an escape character in DuckDB's standard string literals
- `Repository("'core")` -> `INSTALL x FROM '''core'`

`validIdentifier` keeps names out of the quoted path entirely, `utf8.ValidString` runs before the `unicode.IsControl` range loop (so the loop can never see replacement runes), and `hasOuterQuoteWrapper` is a UX guard rather than the security boundary — `quote` is, and it is correct. This is a real improvement over the previous `TrimPrefix`/`TrimSuffix` quoting, which left embedded quotes unescaped.

### What's good

- Precompiling the plan in `NewInitializer` means spec errors surface at construction rather than on the first connection, and `TestInitializerStatements` pins down that mutating the caller's `[]Spec` afterwards cannot change the plan.
- The concurrency and repeat-invocation tests lock in the documented per-connection contract instead of only describing it.
- The README is unusually explicit about the things that actually bite people here: connector laziness vs. preflight, connector-lifetime context, and what happens to a *later* connection when initialization fails.

### Findings

Nothing blocking. Left as inline threads, roughly in priority order:

| # | Where | Issue |
| --- | --- | --- |
| 1 | `extension.go:230` | Exact duplicates now fail startup — a visible `--load-extensions` behavior change |
| 2 | `extension.go:341` | `InstallAndLoadFile`'s `INSTALL` does not feed the `LOAD` that follows it |
| 3 | `extension.go:170` | No exported sentinels; `pkg/query` sets the opposite precedent |
| 4 | `extension_test.go:168` | Negative tables assert only `err != nil` |
| 5 | `extension.go:381` | Built-in repository names are matched case-sensitively |
| 6 | `extension.go:157` | `Parse` reports errors in two different coordinate systems |
| 7 | `extension.go:346` | `Parse` trims whitespace, the programmatic API does not |
| 8 | `main.go:15` | Loop variable at line 137 shadows the newly-imported package |
| 9 | `main.go:54` | Invalid config exits 0 (matches existing style in this file) |

Against the issue's acceptance criteria: table tests, fake `ExecerContext`, repeated invocation, no library logging, README example, and `go test` / `golangci-lint` all check out. #1 is the one that touches "without changing `--load-extensions`".

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fixed in `901a01a6`. Exact semantic duplicates now collapse in the compiled plan, while same-target mode or repository conflicts still return `ErrInvalidSpec`. Comparison normalizes identifier case, default versus explicit `core`, and DuckDB extension aliases; coverage includes repeated CLI entries, empty/Core equivalence, direct-file repeats, and conflicting specs.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fixed in `901a01a6`. DuckDB repository aliases are case-sensitive, so I chose a construction-time canonical-case error instead of silently lowercasing and potentially reinterpreting a relative custom repository named `Core`. The diagnostic suggests `core`; `./Core` remains a quoted custom path. Parse and programmatic cases are covered.

packages/server/duckdb-server-go/main.go
derekperkins: Agreed on the process-level concern. These paths follow the existing return-after-log convention used by the other configuration and startup failures, and the server still fails closed. Making exit statuses consistently non-zero needs a broader `run() error` refactor; changing only the two new paths would be inconsistent and is outside issue #1125, so I am leaving that behavior unchanged here and recommend a focused follow-up.

packages/server/duckdb-server-go/pkg/extension/extension_test.go
derekperkins: Fixed in `901a01a6`. Both negative tables now assert `ErrInvalidSpec` plus a case-specific validation substring, so each row pins the intended branch. The execution-error table also asserts `ErrInstall` versus `ErrLoad` while retaining the underlying driver error.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fixed in `901a01a6`. Added exported `ErrInvalidSpec`, `ErrInstall`, and `ErrLoad` sentinels. Parse and compilation failures wrap `ErrInvalidSpec`; execution failures wrap the phase sentinel and original driver error, so `errors.Is` can classify both. Nil context and nil execer remain argument errors.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fixed in `901a01a6`. `Parse` now validates each entry at its source location and reports `input N entry M` for both validation and cross-entry conflicts. Multi-argument and multi-entry regression cases verify the current and prior coordinates; programmatic construction keeps the separate `specification N` vocabulary.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Verified against `901a01a6`.

`httpfs,httpfs` and `spatial,spatial|core` now both collapse to a single `INSTALL`/`LOAD` pair, while `h3|core,h3|community` and `s3|core,httpfs|community` still fail with `ErrInvalidSpec` and point back at `input 1 entry 1`. Folding alias-equivalence into the identity (so `s3,httpfs` collapses to one pair) is a better answer than what I asked for.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Verified — and I checked the DuckDB side directly rather than take the mechanism on trust, since my original comment asked for exactly that. Against DuckDB **v1.5.5**:

- `LOAD 'nosuchext_zzz'` resolves as a **name**, not a relative path — the error names `~/.duckdb/extensions/v1.5.5/osx_arm64/nosuchext_zzz.duckdb_extension`. So emitting the derived name as a quoted literal is safe; quoting does not push it onto the path branch.
- `INSTALL '/tmp/mysql.duckdb_extension'` fails with *"Installing extension **'mysql_scanner'** failed"* — DuckDB applies its alias mapping to the filename-derived name **at install time**. That is precisely what `canonicalExtensionName` reproduces, so the emitted `LOAD` names what `INSTALL` actually wrote.
- `SELECT extension_name, aliases FROM duckdb_extensions()` on v1.5.5 returns exactly your table, with nothing missing:

  ```
  httpfs            <- [http https s3]
  motherduck        <- [md]
  mysql_scanner     <- [mysql]
  odbc_scanner      <- [odbc]
  postgres_scanner  <- [postgres]
  sqlite_scanner    <- [sqlite sqlite3]
  unity_catalog     <- [uc_catalog]
  ```

Plan output matches: `INSTALL '/opt/MySQL.duckdb_extension'` → `LOAD 'mysql_scanner'`, covering basename, ASCII-casing, and alias in one path.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Verified: on an `INSTALL` failure `errors.Is(err, ErrInstall)` is true, `errors.Is(err, ErrLoad)` is false, and the driver error is still reachable via `errors.Is`. The double-`%w` in `wrap` carries both the phase sentinel and the cause, which is what makes the retryable-vs-fatal split usable from a caller.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension_test.go
derekperkins: Verified. Both tables now assert `ErrInvalidSpec` plus a branch-specific substring, and `TestInitializerStopsAtFirstError` separates `ErrInstall` from `ErrLoad`. The specific case I worried about — `{Path: "  "}` — now pins `path is blank` rather than passing for whichever validation happens to fire first.

Resolving.

One residual gap, non-blocking and not worth reopening: `command.wrap` sits at 40% statement coverage because only the `repository` branch is exercised. The `file` and bare-name branches format untested, so a failing `INSTALL`/`LOAD` on a file spec goes through a message path no test has seen.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Verified: `x|Core` now fails at construction with a message that names `core` and suggests `./Core`, and `x|./Core` still compiles to `INSTALL x FROM './Core'`.

Erroring is the better of the two options I offered — agreed that silently lowercasing would quietly reinterpret a relative directory genuinely named `Core`, which is worse than making the operator disambiguate.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Verified: `Parse("httpfs", "aws,sp-atial")` reports `input 2 entry 2: name "sp-atial" is not a safe DuckDB identifier`, and the conflict case correctly maps back to `input 2 entry 2 conflicts with input 1 entry 1`.

Keeping `specification N` for programmatic construction is right — different entry point, different coordinates, and conflating them would be worse than the original mismatch.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fair, and documenting was one of the two options I offered — trailing whitespace being legal path data is exactly why I did not push for trimming.

The `Spec` doc comment and the README now both state the split explicitly, and the tests pin exact preservation on the programmatic side while `Parse` stays grammar-trimming. Verified: `Parse("custom| /opt/repo ")` → `INSTALL custom FROM '/opt/repo'`, `InstallAndLoad("x", Repository("  https://r  "))` → `INSTALL x FROM '  https://r  '`.

Resolving.

packages/server/duckdb-server-go/main.go
derekperkins: Verified in `901a01a6` — loop binding renamed to `ext`, package reference unshadowed.

Resolving.

packages/server/duckdb-server-go/main.go
derekperkins: Agreed, and this matches how I framed it originally: pre-existing style, not a regression from this PR, and converting only the two new paths would leave the file internally inconsistent. Scoping a `run() error` refactor out of #1125 is the right call.

Resolving. Worth actually filing the follow-up rather than leaving it in a resolved thread, since that is where this kind of note usually goes to die.

derekperkins:
## Re-review of `901a01a6`

All nine threads verified and resolved — detail in each thread rather than repeated here.

### Verification on `901a01a6`

| check | result |
| --- | --- |
| `go test ./pkg/extension -count=1 -race -cover` | pass, 92.9% of statements |
| `go build -tags=duckdb_arrow ./...` | pass |
| `golangci-lint run ./...` | 0 issues |
| `gofmt -l` | clean |

I re-ran the SQL-generation probe against the rewritten compiler. Injection handling is unchanged and still correct, including on the new file-install path:

```
InstallAndLoad("x", Repository("r'; ATTACH 'evil.db"))  ->  INSTALL x FROM 'r''; ATTACH ''evil.db'
InstallAndLoadFile("/o'/x.duckdb_extension")            ->  INSTALL '/o''/x.duckdb_extension'  +  LOAD 'x'
```

The `InstallAndLoadFile` change is the one I would not have accepted on inspection alone, since my original comment asked for confirmation against real DuckDB. I ran that confirmation — v1.5.5 alias-resolves the filename-derived name at install time, `duckdb_extensions()` reports exactly the alias table in this PR, and a quoted argument to `LOAD` resolves as a name rather than a relative path. The mechanism holds; details are in that thread.

### Two non-blocking notes on the new code

Neither is worth its own thread, and neither should hold the PR.

1. **`canonicalExtensionName` is redundant in the emitted SQL.** It is genuinely load-bearing for dedupe identity, but DuckDB alias-resolves on `LOAD` as well — `LOAD 's3'` reports the `httpfs` path, and `LOAD 'mysql'` succeeds against an installed `mysql_scanner`. So `LOAD 'mysql'` would work as well as `LOAD 'mysql_scanner'`.

   The useful consequence is that alias-table drift is mostly benign: a *missing* alias costs you dedupe precision, not a broken `LOAD`. The one case that would actually break is DuckDB **removing or retargeting** an existing alias, which would make the derived `LOAD` name miss what `INSTALL` wrote. Low risk, and the comment already scopes the table to the pinned release.

2. **Coverage dipped 96.5% → 92.9%**, almost entirely in switch arms of the two lookup tables (`canonicalExtensionName` 50%, `caseInsensitiveBuiltIn` 57%) — low-value to chase for constant-mapping tables. The one with substance is `command.wrap` at 40%: only the `repository` branch is exercised, so an `INSTALL`/`LOAD` failure on a file spec formats through untested code. Noted in the test thread.

`targetConflictError.Error()` showing 0% is expected, not a gap — both call sites use `errors.As` and format their own message, so the method exists only to satisfy the interface.

### Assessment

The duplicate-collapse identity (normalizing case, default-vs-explicit `core`, and aliases together) is a better answer than the narrower fix I suggested. Nothing outstanding from my side.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,612 @@
...
+	case "postgres":
+		return "postgres_scanner"
+	case "sqlite", "sqlite3":
+		return "sqlite_scanner"
+	case "uc_catalog":
+		return "unity_catalog"
+	default:
+		return name
+	}
+}
```
derekperkins: how do we keep this up to date? what if it goes stale?

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,612 @@
...
+
+func asciiLower(value string) string {
+	result := []byte(value)
+	for index, char := range result {
+		if char >= 'A' && char <= 'Z' {
+			result[index] = char + ('a' - 'A')
+		}
+	}
+	return string(result)
+}
```
derekperkins: why does this exist and not strings.ToLower?

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Good question, and the honest answer is that today nothing keeps it up to date — the comment is the only thing pointing at the pinned release. Two parts to this: how bad drift actually is, and a way to make it self-policing.

### How bad is drift, really

Narrower than it looks, because DuckDB alias-resolves on **both** sides. Verified on v1.5.5:

- `LOAD 's3'` reports the `httpfs` path, and `LOAD 'mysql'` succeeds against an installed `mysql_scanner` — so DuckDB resolves aliases on `LOAD`
- `INSTALL '/tmp/mysql.duckdb_extension'` fails with *"Installing extension `'mysql_scanner'` failed"* — and on `INSTALL`

So the table is **redundant in the emitted SQL**: `LOAD 'mysql'` would work just as well as `LOAD 'mysql_scanner'`. It is only genuinely load-bearing for dedupe identity. That splits drift into:

| drift | effect |
| --- | --- |
| DuckDB **adds** an alias you lack | dedupe under-merges — two specs for one extension both emit. Harmless; `LOAD` still resolves. |
| DuckDB **removes or retargets** an alias | **breaks.** File installs emit a `LOAD` naming something `INSTALL` never wrote, and dedupe can raise a spurious `ErrInvalidSpec` conflict between two names DuckDB considers distinct. |

Only the second row is a real problem, and it is the rare one — aliases are backward-compat shims.

### Making it self-policing

`duckdb_extensions()` exposes the registry, and it needs no network and no installed extensions — the alias list is static and readable from `:memory:`. So the table can be pinned by a test rather than by vigilance.

That needs the switch to be enumerable, so as a map:

```go
// extensionAliases mirrors DuckDB's alias registry for the pinned DuckDB
// release so duplicate detection uses the same installed target names.
// TestExtensionAliasesMatchDuckDB pins this to duckdb_extensions().
var extensionAliases = map[string]string{
	"http": "httpfs", "https": "httpfs", "s3": "httpfs",
	"md":         "motherduck",
	"mysql":      "mysql_scanner",
	"odbc":       "odbc_scanner",
	"postgres":   "postgres_scanner",
	"sqlite":     "sqlite_scanner", "sqlite3": "sqlite_scanner",
	"uc_catalog": "unity_catalog",
}

func canonicalExtensionName(name string) string {
	name = asciiLower(name)
	if canonical, ok := extensionAliases[name]; ok {
		return canonical
	}
	return name
}
```

and then a `duckdb_arrow`-tagged test, so it rides along with the `go test -tags=duckdb_arrow ./...` you already run and the default pure-Go `go test ./pkg/extension` stays cgo-free:

```go
//go:build duckdb_arrow

package extension

import (
	"database/sql"
	"testing"

	_ "github.com/duckdb/duckdb-go/v2"
)

// TestExtensionAliasesMatchDuckDB pins extensionAliases to the alias registry of
// the DuckDB build this module links against, in both directions. A DuckDB bump
// that adds, removes, or retargets an alias fails here instead of silently
// drifting into a LOAD that names something INSTALL never wrote.
func TestExtensionAliasesMatchDuckDB(t *testing.T) {
	db, err := sql.Open("duckdb", ":memory:")
	if err != nil {
		t.Fatalf("open duckdb: %v", err)
	}
	defer db.Close()

	rows, err := db.Query(
		`SELECT unnest(aliases) AS alias, extension_name FROM duckdb_extensions()`,
	)
	if err != nil {
		t.Fatalf("query duckdb_extensions(): %v", err)
	}
	defer rows.Close()

	actual := make(map[string]string)
	for rows.Next() {
		var alias, canonical string
		if err := rows.Scan(&alias, &canonical); err != nil {
			t.Fatalf("scan: %v", err)
		}
		actual[alias] = canonical
	}
	if err := rows.Err(); err != nil {
		t.Fatalf("rows: %v", err)
	}
	if len(actual) == 0 {
		t.Fatal("duckdb_extensions() reported no aliases; query is wrong")
	}

	for alias, canonical := range actual {
		if got := canonicalExtensionName(alias); got != canonical {
			t.Errorf("DuckDB aliases %q -> %q, canonicalExtensionName gives %q",
				alias, canonical, got)
		}
	}
	for alias, canonical := range extensionAliases {
		got, ok := actual[alias]
		if !ok {
			t.Errorf("extensionAliases has %q -> %q but DuckDB has no such alias",
				alias, canonical)
			continue
		}
		if got != canonical {
			t.Errorf("extensionAliases maps %q -> %q, DuckDB maps it -> %q",
				alias, canonical, got)
		}
	}
}
```

I ran this against `901a01a6` — it passes, and I mutation-tested it against all three drift modes to make sure it is not vacuous:

```
alias removed from table  -> FAIL  DuckDB aliases "sqlite3" -> "sqlite_scanner", canonicalExtensionName gives "sqlite3"
alias retargeted          -> FAIL  extensionAliases maps "md" -> "markdown_scanner", DuckDB maps it -> "motherduck"
alias invented            -> FAIL  extensionAliases has "pg" -> "postgres_scanner" but DuckDB has no such alias
```

Both directions matter: the first catches DuckDB gaining an alias, the second and third catch the breaking case.

### The other option

If you would rather not carry the table at all, drop canonicalization from the **emitted SQL** and keep it only for dedupe keys — emit `LOAD '<raw derived name>'` and let DuckDB resolve. Then a stale table costs you dedupe precision and nothing else, and can never produce a broken statement. That is strictly less code, at the cost of the duplicate detector occasionally missing that `s3` and `httpfs` are the same thing.

My preference is the pinning test: you keep the better dedupe and the failure mode becomes a red CI run on a DuckDB bump, which is where you want to find out.

Happy either way — flagging that this is a genuine open question rather than something I think blocks the PR.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Because `strings.ToLower` would be **wrong here** — DuckDB lowercases extension names ASCII-only, and `strings.ToLower` does full Unicode case folding. I checked both sides rather than reason about it.

DuckDB v1.5.5, reading back which filename each `LOAD` resolves to:

```
LOAD 'HTTPFS'    => .../httpfs.duckdb_extension      <- ASCII IS lowercased
LOAD 'NoSuchExt' => .../nosuchext.duckdb_extension   <- ASCII IS lowercased
LOAD 'Ä'         => .../Ä.duckdb_extension           <- non-ASCII is NOT
LOAD 'K'         => .../K.duckdb_extension           <- U+212A KELVIN SIGN, NOT folded
LOAD 'İ'         => .../İ.duckdb_extension           <- U+0130, NOT folded
```

`asciiLower` reproduces that exactly. `strings.ToLower` does not:

```
"Ä"  -> "ä"       2 bytes -> 2   (DuckDB keeps Ä)
"K"  -> "k"       3 bytes -> 1   (U+212A folds into ASCII k)
"İ"  -> "i"       2 bytes -> 1   (U+0130 folds into ASCII i)
```

Two concrete failures if this were `strings.ToLower`:

1. **`LOAD` names a file `INSTALL` never wrote.** `InstallAndLoadFile("/opt/Ä.duckdb_extension")` would emit `INSTALL '/opt/Ä.duckdb_extension'` + `LOAD 'ä'`, but DuckDB installed it as `Ä`. That is precisely the bug the earlier `LOAD '<path>'` form had, reintroduced through case folding. Your `/opt/Ä.duckdb_extension` → `LOAD 'Ä'` test case is the regression guard for it.

2. **False duplicate collisions across the ASCII boundary.** `K` (U+212A) and `k` fold together under Unicode, so an extension named `K` and an unrelated one named `k` would either collapse into one plan entry or trip a spurious `ErrInvalidSpec` conflict — for two things DuckDB considers distinct. Same for `İ`/`i`.

So this is load-bearing, not a reinvention. Worth putting the reason in the comment though — right now `asciiLower` reads like a hand-rolled `strings.ToLower`, and the next person will make exactly this substitution. Something like:

```go
// asciiLower lowercases ASCII only, matching DuckDB's extension-name handling.
// strings.ToLower would fold non-ASCII into the ASCII namespace (U+212A -> "k",
// "Ä" -> "ä"), producing LOAD names DuckDB never installed and false duplicate
// collisions between distinct extensions.
```

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Addressed by the simplification in `3241f37d`: the alias table, canonicalization, duplicate identity, and conflict planner are gone. The initializer now emits the raw safely quoted derived name and lets DuckDB resolve aliases. There is no application-maintained alias registry left to drift.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Addressed by the simplification in `3241f37d`: `asciiLower` and all application-side case canonicalization were removed. Named extensions and file-derived basenames are emitted as safely quoted raw values, so DuckDB owns casing and alias resolution directly.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Confirmed — and deleting the mirror is the better of the two options I laid out, since there is now nothing left to drift.

I verified the round trip still holds without the table. `InstallAndLoadFile("/opt/MySQL.duckdb_extension")` now emits `INSTALL '/opt/MySQL.duckdb_extension'` + `LOAD 'MySQL'`, and DuckDB applies **the same** ASCII-lowercase-then-alias mapping on both sides: it installs that file as `mysql_scanner`, and `LOAD 'MySQL'` resolves through `mysql` to `mysql_scanner` as well. Both ends land on the same target, which is exactly why the application-side table was never load-bearing for correctness.

`extension_test.go:261` pinning `/opt/HTTPS.v2.duckdb_extension` → `LOAD 'HTTPS'` is a good regression guard for this: under the old code that would have been rewritten to `httpfs`, so the assertion fails loudly if canonicalization ever creeps back in.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Confirmed, and this resolves it more completely than documenting `asciiLower` would have — with canonicalization gone there is no case folding left to get wrong.

The property I was protecting still holds by construction: `installedExtensionName` returns the raw basename segment, so `/opt/Ä.duckdb_extension` yields `LOAD 'Ä'`, and I verified DuckDB does not fold non-ASCII (`LOAD 'Ä'` looks for `Ä.duckdb_extension`, `LOAD 'K'` with U+212A looks for `K.duckdb_extension`). Emitting the raw value and letting DuckDB own casing is the correct call, and it is now impossible to get the ASCII boundary wrong because nothing on this side touches it.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Reopening — not to relitigate the design, but because the basis on which we closed this no longer exists, and I would rather that be visible than buried in a resolved thread.

`901a01a6` fixed this with `caseInsensitiveBuiltIn`, failing at construction with `repository "Core" matches built-in repository "core" but must use canonical casing`. `3241f37d` removed that check, and `Core` is now accepted as a custom repository — `extension_test.go:136` pins the new behavior.

That is consistent with the new "DuckDB owns semantics" philosophy, and the resulting error is not bad:

```
INSTALL 'spatial' FROM 'Core'
=> IO Error: Failed to install local extension "spatial",
   no access to the file at PATH "Core/v1.5.5/osx_arm64/spatial.duckdb_extension"
```

It does at least surface `Core/` being treated as a directory. So I am **not** asking for the check back. I am asking that reversing a resolved thread be a stated decision rather than a side effect of the refactor. Happy for you to re-resolve with a one-line "intentional, DuckDB reports it adequately" — I just did not want to leave it silently closed on behavior that changed.

Broader trade-off in the new thread on `validateSpec`.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	hasName := spec.Name != ""
+	hasPath := spec.Path != ""
+	if hasName == hasPath {
+		return errors.New("exactly one of Name and Path must be set")
+	}
+
+	if hasName {
+		if strings.TrimSpace(spec.Name) == "" {
```
derekperkins: **Deferring name validation to DuckDB costs a clear startup diagnostic.**

Not blocking, and I like the direction of `3241f37d` — deleting the mirrored DuckDB rules is right, and it dissolved both of my open threads rather than patching them. But it is worth being explicit about what moved.

`validIdentifier` and `validateValue` are gone, so a name that is not a valid DuckDB identifier is now quoted and handed to DuckDB. **Safety is intact** — I re-verified that quoting neutralizes injection, including that `LOAD 'x''; ATTACH ''evil.db'' AS e; --'` attaches nothing (`duckdb_databases()` stays `[memory system temp]` across the attempt). What is lost is diagnostic quality, because DuckDB's message is written for a different mistake than the one the operator made:

```
--load-extensions "foo.bar"
  INSTALL 'foo.bar' FROM core
  => Invalid Input Error: Cannot pass both a repository and a full path url

--load-extensions "spatial|Core"
  INSTALL 'spatial' FROM 'Core'
  => IO Error: Failed to install local extension "spatial",
     no access to the file at PATH "Core/v1.5.5/osx_arm64/spatial.duckdb_extension"
```

The first is actively misleading: the operator typed a name and passed no path.

If you want the startup diagnostic back without reintroducing a mirror of DuckDB's identifier grammar, the cheap version is a **shape** check rather than a whitelist. `filePath`'s own comment already states the rule you depend on — DuckDB treats an operand containing `.`, `/`, or `\` as a path — so a *named* spec containing one of those can never be a name:

```go
if strings.ContainsAny(spec.Name, `./\`) {
    return fmt.Errorf(
        "name %q contains a dot or path separator, which DuckDB reads as a file path; "+
            "use LoadFile or InstallAndLoadFile instead", spec.Name)
}
```

Three lines, encodes a rule this file already relies on, and needs no tracking of DuckDB's naming rules as they evolve.

Your call whether the trade is worth it. Fail-closed-at-first-connection is a legitimate answer, particularly since `main.go`'s `GetExtensions` preflight forces that connection before the server accepts traffic — so an operator still finds out at startup, just with DuckDB's wording instead of yours.

derekperkins (commented):
Re-review of `3241f37d` — inline threads carry the detail.

derekperkins:
## Re-review of `3241f37d`

This is the right shape of fix for the two questions I had open. Rather than pinning the alias table to DuckDB, it deletes the mirror entirely and lets DuckDB own casing, aliases, and name resolution — which removes the drift risk instead of policing it. −375 net lines and the behavior is more correct, not less.

### Verification on `3241f37d`

| check | result |
| --- | --- |
| `go test ./pkg/extension -count=1 -race -cover` | pass, **94.9%** (up from 92.9%) |
| `go test -tags=duckdb_arrow ./... -count=1` | pass — `extension`, `query`, `server` all green |
| `go build -tags=duckdb_arrow ./...` | pass |
| `golangci-lint run ./...` | 0 issues |
| `gofmt -l` | clean |

### What I checked against real DuckDB rather than by inspection

The refactor changes what SQL is emitted for *every* named extension — names are now quoted — so the syntax assumption needed proving, not assuming:

```
INSTALL 'json' FROM core                            => OK
INSTALL 'httpfs' FROM core                          => OK
INSTALL 'inet' FROM 'https://extensions.duckdb.org' => OK
LOAD 'json'                                         => OK
```

Quoted operands resolve as extension **names**, so the primary path is sound.

**Injection remains fully neutralized, and now covers names too.** The payload is treated as one filename string, and I confirmed the side effect never fires:

```
LOAD 'x''; ATTACH ''evil.db'' AS e; --'
  => Permission Error: ... error while loading from 'x'; ATTACH 'evil.db' AS e; --

duckdb_databases() before: [memory system temp]
duckdb_databases() after:  [memory system temp]      <- no ATTACH occurred
```

`TestInitializerQuotesDuckDBValues` is a genuine improvement over what it replaced — it drives the nasty payloads through both `Parse` and the programmatic constructors and asserts the exact escaped SQL.

**The alias deletion is provably safe.** DuckDB applies ASCII-lowercasing and alias resolution on *both* `INSTALL` and `LOAD`, so emitting the raw basename round-trips: `/opt/MySQL.duckdb_extension` installs as `mysql_scanner`, and `LOAD 'MySQL'` resolves to `mysql_scanner` too. `extension_test.go:261` pinning `HTTPS.v2` → `LOAD 'HTTPS'` is a good guard — under the old code that would have been rewritten to `httpfs`.

`filePath()` is a real correctness gain independent of the cleanup: a bare relative filename like `myext` was previously ambiguous with an extension name, and `./myext` disambiguates it.

### Open items

- **One thread reopened** — `3241f37d` removed the canonical-casing check that closed the repository-casing thread in `901a01a6`. The new DuckDB error is adequate, so I am not asking for the check back, only that the reversal be an explicit decision rather than a silent one. Re-resolve with a one-liner and I am satisfied.
- **One new thread** on `validateSpec` about deferred name validation. Safety is intact; the cost is that `--load-extensions "foo.bar"` now surfaces *"Cannot pass both a repository and a full path url"*, which is misleading for someone who typed a name. I sketched a three-line shape check that recovers the startup diagnostic without reintroducing a mirror of DuckDB's grammar. Non-blocking.

One footnote not worth a thread: `installedExtensionName` can return `""` (for a path of `.` or `..`), which would emit `LOAD ''`. It is effectively unreachable, since the preceding `INSTALL '.'` fails first and short-circuits.

### Assessment

Net strongly positive. The simplification removed the two things I was most uneasy about — a hand-maintained mirror of DuckDB internals and application-side case folding — and replaced them with behavior I was able to verify end to end against DuckDB v1.5.5.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: I am keeping this deferred intentionally. The requested architecture is to lean on DuckDB for bad extensions and avoid maintaining a second grammar. The proposed shape check is small, but it would move semantic validation back into the package specifically for name/path discrimination. Values remain safely quoted, and the example server preflight forces the first connection before HTTP serving, so failure is still at startup. `Parse` keeps grammar errors, `NewInitializer` keeps only Go-model structural errors, and DuckDB owns the rest.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Intentional in `3241f37d`. The simplification deliberately removed the canonical-case precheck: exact built-ins remain tokens, while `Core` is passed as a custom repository path and DuckDB reports that interpretation. I do not want to silently lowercase a legitimate relative directory or add a special-case semantic validator when the package now delegates repository validity to DuckDB. The server preflight still surfaces the failure before serving.

derekperkins: That is exactly what I was after — an explicit decision rather than a silent reversal. Agreed on not lowercasing a legitimate relative directory just to catch a typo, and `Core` being reported as the path interpretation is a reasonable place to land.

Resolving.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: Fair, and the boundary you describe — `Parse` owns grammar, `NewInitializer` owns Go-model structure, DuckDB owns semantics — is a cleaner line than the one my shape check would have drawn. It was a judgment call and this is a reasoned answer to it.

I did check that the guarantee it rests on is documented rather than assumed. The README now states the delegation outright ("delegates extension names, aliases, repositories, file compatibility, and repeated or conflicting specifications to DuckDB ... DuckDB's own errors remain the source of truth"), and it still tells embedders that `NewConnector` is lazy and that they need their own `PingContext`/first-query preflight. That matters, because the `GetExtensions` preflight only covers the example server — a library consumer without one would meet these errors on first connection rather than at startup. Since that is spelled out, "failure is still at startup" holds for anyone who follows the documented pattern.

Resolving — nothing further from me.

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+func isBuiltIn(repository Repository) bool {
+	switch repository {
+	case Core, CoreNightly, Community, LocalBuildDebug, LocalBuildRelease:
+		return true
+	default:
+		return false
+	}
+}
```
derekperkins: do we need to know this?

packages/server/duckdb-server-go/pkg/extension/extension.go
```diff
@@ -0,0 +1,351 @@
...
+	Core Repository = "core"
+	// CoreNightly is DuckDB's nightly core extension repository.
+	CoreNightly Repository = "core_nightly"
+	// Community is DuckDB's community extension repository.
+	Community Repository = "community"
+	// LocalBuildDebug is DuckDB's local debug-build repository.
+	LocalBuildDebug Repository = "local_build_debug"
+	// LocalBuildRelease is DuckDB's local release-build repository.
+	LocalBuildRelease Repository = "local_build_release"
+)
```
derekperkins: do we need these?

derekperkins: No. Removed in 2f1c945d: Repository, Mode, Spec, all repository constants, and NewInitializer are gone. The package is now pkg/extensions; ParseAndInstall executes a string list directly, and callers needing control invoke the install/load helpers directly with ordinary repository strings.

packages/server/duckdb-server-go/pkg/extension/extension.go
derekperkins: No semantic repository knowledge remains after 2f1c945d. Empty repository strings omit FROM and use the DuckDB default; explicit identifier-shaped strings are passed through for DuckDB to recognize or reject, while paths and URLs are safely quoted. The only retained distinction is SQL syntax, because DuckDB parses repository aliases and repository paths as different token types and INSTALL does not accept parameters.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,230 @@
...
+// ParseAndInstall accepts the command-line grammar name|repository,name2 and
+// installs and loads each entry in order. Each argument may contain comma-
+// separated entries, so a []string can be expanded directly. An omitted
+// repository uses DuckDB's default.
+func ParseAndInstall(
+	ctx context.Context,
+	execer driver.ExecerContext,
+	values ...string,
+) error {
+	extensions, err := parse(values...)
```
derekperkins, line 40: **Grammar validation moved from startup into every connection.**

`ParseAndInstall` parses inside the connector callback, so `--load-extensions` grammar errors are now only discovered when a physical connection is made. I ran the built binary to see what an operator actually gets:

```
$ duckdb-server-go --load-extensions "spatial," --port 39311 --database ":memory:"
INFO  msg="DuckDB Server configuration" config="map[... load_extensions:spatial, ... port:39311]"
ERROR msg="main: error getting extensions" error="query: failed to get extensions: extensions: invalid input: input 1 entry 2: entry is blank"
$ echo $?
0
```

Three things degrade there:

1. The server prints a full, healthy-looking configuration banner before it notices the flag is malformed.
2. The headline is `main: error getting extensions`, wrapped through `query: failed to get extensions:` — so it blames the inventory query, not the flag. The previous code logged `main: invalid load-extensions` with the offending value as a structured field.
3. Exit status is `0`. That is the pre-existing issue from the other thread, but a pure config typo is exactly where it stings most.

I want to be explicit that this is a **different category** from the deferrals I already agreed with, so this does not read as me relitigating them. Those handed *DuckDB semantics* — does this extension exist, is this repository real — to DuckDB, which is right, because DuckDB is the authority and the package would otherwise maintain a second grammar. This defers **your own grammar**: `parse` is pure, needs no connection, no I/O, and no DuckDB, and it is the one thing in this package that genuinely is the package's job.

The minimal fix keeps the new shape and adds no types:

```go
// Validate reports whether values parse as the command-line grammar. It performs
// no I/O, so callers can reject a malformed flag before opening a connector.
func Validate(values ...string) error {
	_, err := parse(values...)
	return err
}
```

and in `main.go`, before `NewConnector`:

```go
if err := extensions.Validate(*extensionsStr); err != nil {
	logger.Error("main: invalid load-extensions", "error", err, "load-extensions", *extensionsStr)
	return
}
```

One exported function. No `Spec`, no initializer type, `ParseAndInstall` unchanged. It also stops re-splitting the same string on every physical connection, though that cost is negligible next to the diagnostic.

Not blocking — the server still fails closed before serving traffic. But this is the one deferral where the package is giving away something it already knows.

derekperkins (commented):
Re-review of `2f1c945d` — one thread, detail inline.

derekperkins:
## Re-review of `2f1c945d`

Second structural pass: `pkg/extension` → `pkg/extensions`, and `Spec`, `Repository`, `Mode`, the repository constants, and `NewInitializer` are all gone in favour of direct `(ctx, execer, …)` functions. Callers write the `driver.ExecerContext` closure themselves.

### Verification on `2f1c945d`

| check | result |
| --- | --- |
| `go test ./pkg/extensions -count=1 -race -cover` | pass, 94.0% of statements |
| `go test -tags=duckdb_arrow ./... -count=1` | pass — `extensions`, `query`, `server` all green |
| `go build -tags=duckdb_arrow ./...` | pass |
| `golangci-lint run ./...` | 0 issues |
| `gofmt -l` | clean |

### What this got right

**The repository token/string distinction is the best version of this we have had.** Emitting identifier-shaped repositories bare rather than quoted is what lets DuckDB produce a targeted diagnostic. The mis-cased case I reopened a thread about now yields:

```
INSTALL 'x' FROM Core
=> Binder Error: 'Core' is not a known repository name. Are you trying to query
   from a repository by path? Use single quotes: `FROM 'Core'`
```

That beats `3241f37d`'s opaque `no access to the file at PATH "Core/v1.5.5/..."`, and it beats the canonical-casing check we deleted — DuckDB states the problem more precisely than either version of our own code did. Good outcome for that thread.

**Dropping `FROM` when no repository is given** respects a configured `custom_extension_repository` rather than pinning `core`. Verified `INSTALL 'json'` with no `FROM` is valid, and the flag help was updated to match ("use DuckDB's default").

**Injection safety holds across the new unquoted path**, which is the thing worth checking hardest since bare emission is new. The bare branch is gated on `[A-Za-z_][A-Za-z0-9_]*`, so nothing reaching it can carry a quote, semicolon, or space. Verified:

```
InstallAndLoad(name: "x", repo: "core; ATTACH 'e.db'")
  -> INSTALL 'x' FROM 'core; ATTACH ''e.db'''      <- quoted, not a token
InstallAndLoad(name: "x'; DROP TABLE t; --", repo: "core")
  -> INSTALL 'x''; DROP TABLE t; --' FROM core     <- name always quoted
InstallAndLoadFile("/o'/x.duckdb_extension")
  -> INSTALL '/o''/x.duckdb_extension'  +  LOAD 'x'
```

### One thread

Grammar validation moved from startup into the per-connection path, so a malformed `--load-extensions` is now reported as `main: error getting extensions` *after* the config banner, wrapped through `query: failed to get extensions:`, with exit status 0. I ran the built binary to confirm rather than infer it. This is a different category from the deferrals I already agreed with — those handed DuckDB *semantics*, whereas `parse` is pure and is genuinely this package's own job. Suggested a one-function `Validate` that keeps the new shape. Detail in the thread; not blocking.

### Scope note

Issue #1125 asked for "a structured extension specification and a connector-compatible initializer" — neither survives this commit. The underlying goal, usable by other projects without flags, string pointers, or the example logger, is still met by the function API. You wrote the issue so this is your call to make; worth updating #1125 so its acceptance criteria match what actually shipped, since a later reader will otherwise check them against a `Spec` type that no longer exists.

Footnote, not worth a thread: `installedExtensionName` can still return `""` for a path of `.` or `..`, which would emit `LOAD ''`. Unreachable in practice — the preceding `INSTALL` fails first and short-circuits.

packages/server/duckdb-server-go/main.go
```diff
@@ -122,8 +124,8 @@ func main() {
 	fmt.Println("DuckDB Server Extensions:")
 	fmt.Printf("%-20s | %-8s | %-20s | %-20s\n", "name", "version", "repository", "install_mode")
 	fmt.Println("-------------------- | -------- | -------------------- | --------------------")
-	for _, extension := range extensions {
-		fmt.Printf("%-20s | %-8s | %-20s | %-20s\n", extension.Name, extension.Version, extension.Repository, extension.InstallMode)
+	for _, ext := range extensions {
+		fmt.Printf("%-20s | %-8s | %-20s | %-20s\n", ext.Name, ext.Version, ext.Repository, ext.InstallMode)
```
derekperkins: just leave the var named extension and there's no diff

derekperkins: Agreed on the loop variable — with the package now `extensions`, the singular `extension` collides with nothing, so reverting `ext` is correct and drops that hunk from the diff entirely. My original comment was scoped to the old singular package name and no longer applies.

Worth flagging that the rename **moved** the shadow rather than removing it, though. `main.go:116` declares a local slice also called `extensions`:

```go
extensions, err := db.GetExtensions(ctx)
```

which shadows the imported package for the rest of `main()`. It compiles today only because the package's sole use (`extensions.ParseAndInstall` at line 66) sits above it. I confirmed by adding a later reference:

```
./main.go:123:17: extensions.LoadInstalled undefined
    (type []query.Extension has no field or method LoadInstalled)
```

That is the same failure mode I originally described, just relocated one level up — and it is now on the *package-name* collision rather than a loop binding, so it will bite anyone who adds a second `extensions.` call in `main` later.

If you want both the zero-diff loop variable and no shadow, rename the slice instead:

```go
installed, err := db.GetExtensions(ctx)
...
for _, extension := range installed {
```

That restores `extension` as the loop binding exactly as it was, keeps the package reachable throughout `main`, and arguably reads better — `installed` says what the query returned, which is the currently-installed inventory rather than the requested list.

Your call; the current code is correct as written.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
derekperkins: Implemented in 7e7d3399. Added Validate(values...) as a pure wrapper over the same private parser, and main now calls it immediately after logger creation, before certificate probing, connector creation, or the configuration banner. A live malformed-flag probe now reports main: invalid load-extensions with the offending value as the first and only log line. Tests cover the default empty value, valid multi-value input, and ErrInvalidInput classification. ParseAndInstall still reparses per physical connection intentionally; eliminating that would reintroduce stored compiled state, which this API deliberately avoids.

packages/server/duckdb-server-go/main.go
derekperkins: Implemented in 7e7d3399. Restored for _, extension := range extensions and the original field references, so that loop has no PR diff. I kept the extensions slice name rather than adding a speculative rename: the package is used before that local enters scope and there is no later package reference in main.

derekperkins (commented):
Simplification review requested — one proposal, detail inline.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,236 @@
...
+	ErrLoad         = errors.New("extensions: load failed")
+)
+
+type extension struct {
+	name       string
+	repository string
+}
+
+// Validate reports command-line grammar errors before a connection is opened.
+func Validate(values ...string) error {
```
derekperkins, line 26: **Optional simplification: folding `Validate` into `Parse` would make eager validation structural rather than conventional.**

Raising this because you asked whether anything meaningfully simplifies — it is a design proposal, not a defect. `7e7d3399` is correct as it stands, and I verified the fix works:

```
$ duckdb-server-go --load-extensions "spatial," --port 39312
ERROR msg="main: invalid load-extensions" error="extensions: invalid input: input 1 entry 2: entry is blank" load-extensions=spatial,
```

The observation is that `Validate` exists *only* because `ParseAndInstall` defers parsing to connection time. If parsing returned something runnable instead, the second function is unnecessary:

```go
// Installer installs and loads a parsed extension list on one DuckDB connection.
type Installer func(context.Context, driver.ExecerContext) error

// Parse accepts the command-line grammar name|repository,name2 and returns an
// Installer that installs and loads each entry in order. Grammar errors are
// reported here, before any connection exists.
func Parse(values ...string) (Installer, error) {
	parsed, err := parse(values...)
	if err != nil {
		return nil, err
	}
	return func(ctx context.Context, execer driver.ExecerContext) error {
		for _, entry := range parsed {
			if err := InstallAndLoad(ctx, execer, entry.name, entry.repository); err != nil {
				return err
			}
		}
		return nil
	}, nil
}
```

`main.go` collapses to one call:

```go
installExtensions, err := extensions.Parse(*extensionsStr)
if err != nil {
	logger.Error("main: invalid load-extensions", "error", err, "load-extensions", *extensionsStr)
	return
}

connector, err := duckdb.NewConnector(*dbPath, func(execer driver.ExecerContext) error {
	return installExtensions(ctx, execer)
})
```

I prototyped this against `7e7d3399` — it builds clean, and end-to-end the malformed flag still errors eagerly before the configuration banner while `--load-extensions json` starts and preflights normally.

It is only −4 lines, so size is not the argument. Three things it buys:

1. **One exported function instead of two.** `Parse` subsumes `Validate` entirely.
2. **Validation stops being skippable.** Today a consumer can call `ParseAndInstall` without `Validate` and get grammar errors at connection time — exactly the failure `7e7d3399` just fixed. That fix currently lives in `main.go`'s discipline, not in the API's shape, so an embedding application has to know to repeat it. Worth noting the README example would need the same care.
3. **Parsing happens once** rather than on every physical connection, for a string that cannot change.

The honest costs: it adds an exported func type, and it is directionally back toward the `NewInitializer` you deliberately removed. It is much lighter than that was — no `Spec`, no `Mode`, no `Repository`, no repository constants — and unlike the old one it takes `ctx` as a parameter rather than capturing it, which is what forced the README paragraph about connector-lifetime contexts. Test migration is about 13 mechanical call sites.

Entirely your call, and "no" is a perfectly good answer — you have twice chosen to cut API surface on purpose, and one more exported type may not be a trade you want. I would not hold the PR for it.

---

For completeness, the other things I looked at and decided were **not** worth changing:

- **The six error-wrap sites** (`namedError` plus five inline `fmt.Errorf`) are real duplication, but collapsing them means either preformatting the subject — allocating on the happy path — or reintroducing a kind enum, which is what `command.wrap` was before you deleted it. Circular.
- **`repositoryIdentifier`'s ASCII loop** could be `regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")`. Reads better, pulls in `regexp` for one predicate. A wash.
- **The nil `ctx`/`execer` guards** in all four exec functions (~13 lines) return errors that wrap no sentinel, so a caller cannot act on them differently than on the panic they prevent, and `staticcheck` SA1012 already flags nil-context call sites. Marginal either way.

`filePath`, `installedExtensionName`, and the repository token-vs-string split all earn their place — non-obvious behavior with comments explaining why.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
derekperkins: Verified against `7e7d3399` by running the built binary, so this is confirmed rather than taken on the diff:

```
$ duckdb-server-go --load-extensions "spatial," --port 39312
ERROR msg="main: invalid load-extensions" error="extensions: invalid input: input 1 entry 2: entry is blank" load-extensions=spatial,
```

First and only log line, correct attribution, offending value as a field, and no configuration banner ahead of it. That is the whole of what I was asking for.

Noted on the per-connection reparse being deliberate. I have put the alternative in a separate thread since you asked about simplification generally — it avoids stored compiled state by returning a closure rather than a struct, but it is a design proposal and genuinely optional. Not reopening this one for it.

Resolving.

packages/server/duckdb-server-go/main.go
derekperkins: Agreed on both halves — the loop binding is back to zero diff, and declining a speculative rename is reasonable given the package is used only above the local's scope.

To be explicit about what I demonstrated: the compile error only materialises if someone later adds a second `extensions.` reference below line 116, which is a hazard rather than a defect. You have weighed it and the current code is correct, so that is a fine place to leave it.

Resolving.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
derekperkins: Thanks for prototyping this and laying out the tradeoffs. I'm going to keep `Validate` + `ParseAndInstall`.

Returning an `Installer` would retain the parsed list for the connector lifetime and move back toward the stored/composable initializer abstraction we deliberately removed. It also does not reduce total public surface: one exported function plus one exported type replaces two exported functions.

`ParseAndInstall` still validates the complete list before executing SQL on each connection. Omitting `Validate` changes when a grammar error appears, not whether validation occurs. Eager flag validation is intentionally a server concern handled explicitly in `main`; embedding callers can opt into it when startup timing matters. Re-parsing a short configuration list is negligible beside opening a DuckDB connection and running `INSTALL`/`LOAD`.

For callers wanting finer lifecycle control, the direct helpers are the intended path. I don't think the added runnable type is a net simplification here, so no code change planned.

derekperkins: Agreed, and two of those are corrections to my framing rather than differences of opinion.

**Public surface.** You are right that it is a wash. I wrote "one exported function instead of two" and quietly did not count the exported type that replaces it — two identifiers either way. That was the load-bearing claim in my first bullet and it does not hold.

**"Skippable" was the wrong word.** `ParseAndInstall` parses the entire list and returns before executing any SQL, so a malformed list never produces a partial install. The exposure is purely *when* an operator learns, not *whether* validation happens. "Unskippable validation" overstated it; "earlier reporting" is the accurate version, and that is a much smaller claim.

With those two corrected, what is left is the per-connection reparse, which I already flagged as the weakest of the three — and you are right that a `strings.Split` over a short list is invisible next to opening a connection and running `INSTALL`/`LOAD`, quite possibly over the network.

The boundary you are drawing — eager flag validation is a server concern, `main` owns it explicitly, embedders opt in when startup timing matters — is coherent and consistent with the rest of the package. No further argument from me.

Resolving. That clears my queue on this PR; nothing outstanding from my side.

packages/server/duckdb-server-go/README.md
```diff
@@ -44,6 +44,109 @@ mkcert -install # Install mkcert CA
 mkcert localhost # create localhost.pem and localhost-key.pem
 ```
 
+### Programmatic Extension Initialization
```
domoritz, line 47: This section could probably be more terse to be more readable.

packages/server/duckdb-server-go/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,236 @@
...
+// DuckDB distinguishes repository names from paths by SQL token type. A bare
+// relative repository path therefore needs an explicit ./ prefix.
+func repositorySQL(repository string) string {
+	if repositoryIdentifier(repository) {
+		return repository
+	}
+	return quote(repository)
+}
+
+func repositoryIdentifier(repository string) bool {
```
domoritz, line 220: This parsing logic feels brittle.  An we not use a native duckdb logic in some way?

domoritz (approved):

packages/server/duckdb-server-go/pkg/extensions/extensions.go
derekperkins: Good callout. The PR merged before I saw this (my bad for setting it to auto-merge), so I opened a follow up PR for both of these
https://github.com/uwdata/mosaic/pull/1136

packages/server/duckdb-server-go/README.md
derekperkins: https://github.com/uwdata/mosaic/pull/1136

PR #1133, opened 2026-08-09
chore: generate spec schema with ts-json-schema-generator v3 (native TypeScript 7)

domoritz:
Switches `@uwdata/mosaic-spec` schema generation to `ts-json-schema-generator@3.0.0-native.0` (npm dist-tag `native`).

## What v3 is

v3 is a native reimplementation of ts-json-schema-generator, written in Go on top of [typescript-go](https://github.com/microsoft/typescript-go) — the TypeScript 7 compiler. It ships as a prebuilt binary per platform rather than as JavaScript, and it is CLI-only: the same flags as 2.x, no Node API. Source lives at [vega/ts-json-schema-generator-go](https://github.com/vega/ts-json-schema-generator-go).

Because this repo only ever invokes the tool through its `schema` script, the CLI-only restriction costs us nothing — the command line in `packages/vgplot/spec/package.json` is unchanged:

```
ts-json-schema-generator -f tsconfig.json -p src/spec/Spec.ts -t Spec --no-type-check --no-ref-encode --functions hide > dist/mosaic-schema.json
```

## Why

- **TypeScript 7 readiness.** Schema generation now runs on the same type checker the rest of the ecosystem is migrating to, so the spec schema won't be the thing blocking a TS7 move.
- **Fewer transitive dependencies.** The JS implementation pulled in its own copy of `typescript`, `commander`, `glob`, and friends; the native package replaces those with a single platform binary. All six platform binaries (darwin/linux/win32 × arm64/x64) are recorded in the lockfile, so `--frozen-lockfile` installs work on CI and on contributor machines alike.

## Schema diff

The regenerated `dist/mosaic-schema.json` is byte-identical to the 2.9.0 output except for one definition. All 216 definitions are present, and no type, description, or enum ordering changed.

`CSSStyles` gains 27 properties and loses none. These come from TypeScript 7's newer `lib.dom`, which knows about CSS features the older lib predates — anchor positioning, scroll-driven animations, and similar:

```diff
 "all": { "type": "string" },
+"anchorName": { "type": "string" },
+"anchorScope": { "type": "string" },
 "animation": { "type": "string" },
```

The full set added: `anchorName`, `anchorScope`, `animationRange`, `animationRangeEnd`, `animationRangeStart`, `animationTimeline`, `dynamicRangeLimit`, `fieldSizing`, `fontLanguageOverride`, `fontVariantEmoji`, `mathShift`, `positionAnchor`, `positionArea`, `positionTry`, `positionTryFallbacks`, `positionTryOrder`, `positionVisibility`, `scrollTimeline`, `scrollTimelineAxis`, `scrollTimelineName`, `textAutospace`, `textJustify`, `timelineScope`, `viewTimeline`, `viewTimelineAxis`, `viewTimelineInset`, `viewTimelineName`.

Validation behavior is unchanged. `CSSStyles` carries `additionalProperties: {"type": "string"}` in both versions, so these names already validated against the index signature as strings; each new entry is likewise `{"type": "string"}`. The schema just names them explicitly now, which is what editor completion reads. The generated Python API and the example specs are unaffected — both CI "verify generated output does not change" gates produce an empty diff.

## Dependency placement

The generator moves from `dependencies` to `devDependencies`. Nothing under `src/` imports it; it is used only by the build-time `schema` script, so shipping it as a runtime dependency forced every consumer of `@uwdata/mosaic-spec` to install a compiler toolchain they never call. This matches how the repo treats other build-time tooling (`esbuild` and `anywidget` in `mosaic-widget`, `vite` in the examples).

## Testing

`pnpm run lint`, `pnpm run build`, and `pnpm run test` all pass (566 tests across 45 files, including the ajv validation of every example spec against the regenerated schema). `pnpm run generate:python-api` and `pnpm run docs:examples` both leave the tracked generated files unchanged.

Nothing else in the tree changes: `packages/vgplot/spec/dist` is gitignored, and the versioned copies under `docs/public/schema` are written at release time by `bin/publish-schema.js`, not by this PR.

### DIFF
--- a/packages/vgplot/spec/package.json
+++ b/packages/vgplot/spec/package.json
@@ -42,7 +42,9 @@
   "dependencies": {
     "@uwdata/mosaic-core": "workspace:^",
     "@uwdata/mosaic-sql": "workspace:^",
-    "@uwdata/vgplot": "workspace:^",
-    "ts-json-schema-generator": "^2.9.0"
+    "@uwdata/vgplot": "workspace:^"
+  },
+  "devDependencies": {
+    "ts-json-schema-generator": "3.0.0-native.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -222,9 +222,10 @@ importers:
       '@uwdata/vgplot':
         specifier: workspace:^
         version: link:../vgplot
+    devDependencies:
       ts-json-schema-generator:
-        specifier: ^2.9.0
-        version: 2.9.0
+        specifier: 3.0.0-native.0
+        version: 3.0.0-native.0
 
   packages/vgplot/vgplot:
     dependencies:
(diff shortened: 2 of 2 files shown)


domoritz:
Updating to ts7 is blocked by https://github.com/typescript-eslint/typescript-eslint/issues/10940 (at least, maybe more).

domoritz:
While the go port is experimental, I think we can move forward with this change. It's easy to revert and for now works fine (no major schema changes).

jheer (approved):

PR #1138, opened 2026-08-10
fix(python): Enable recommended `ty` config

dangotbanned:
## Description
This one snowballed quickly. 
The more things I filled in, the more issues popped up 😰

Most of the files touched were changes like (https://github.com/uwdata/mosaic/pull/1138/commits/253350112041ced3df91e3b0d3ed24a386847610).
But this PR adds typing for all of `packages/server/duckdb-server/` - which was *mostly* untyped

---

`ty` documents a [Recommended configuration](https://docs.astral.sh/ty/coming-from-mypy-or-pyright/#recommended-configuration) when [Coming from mypy or pyright](https://docs.astral.sh/ty/coming-from-mypy-or-pyright/).
By default it does not flag issues that will show up in other type checkers
The recommendation includes config for Ruff too, which is not yet enabled
`ty` is still in beta, and *for people like me* that means it is [missing features](https://github.com/narwhals-dev/narwhals/issues/3409) that (for now) make slower alternatives a better option

## Related
- Part of #1088
- https://docs.astral.sh/ty/coming-from-mypy-or-pyright/#recommended-configuration


dangotbanned:
RIP me (https://github.com/uwdata/mosaic/actions/runs/31443418608/job/93632614790?pr=1138)

**Updated**: fixed in (https://github.com/uwdata/mosaic/pull/1138/commits/14c7726ac84f598792c74c5784b62ff1be09c70d)

packages/vgplot/vgplot-python/vgplot/_types.py
```diff
@@ -22,3 +24,5 @@
 
 UNSET = sentinel("UNSET")
 """Sentinel for mark channels that were not passed (distinct from None)."""
+
+Mimebundle: TypeAlias = dict[str, str] | tuple[dict[str, Any], dict[str, Any]] | None
```
dangotbanned: Not sure if `MimeBundle` is better?

pyproject.toml
```diff
@@ -42,8 +44,14 @@ ignore = [
     "PLE0604",
     "RUF022", 
     "PLC0105",
+    # NOTE: https://docs.astral.sh/ty/coming-from-mypy-or-pyright/#recommended-configuration
+    "ANN401",
 ]
 
+[tool.ruff.lint.per-file-ignores]
+"**/*.ipynb" = ["ANN"]
+"dev/notebooks/*" = ["ANN"]
```
dangotbanned, line 53: I should've done the ignores this way in #1103, but in a rush and wasn't expecting it (https://github.com/uwdata/mosaic/pull/1103#issuecomment-5106676636) 🤦‍♂️

packages/server/duckdb-server/pkg/server.py
```diff
@@ -151,15 +152,19 @@ async def http_handler(res: Res, req: Req) -> None:
...
+            data = ujson.loads(message)
             handle_query(handler, con, cache, data)
         elif method == "POST":
-            data = (await res.get_json()) or {}
-            handle_query(handler, con, cache, data)
+            maybe_data: _QueryParams | None = await res.get_json()
+            if maybe_data:
+                handle_query(handler, con, cache, maybe_data)
+            else:
+                raise NotImplementedError
```
dangotbanned, line 167: https://github.com/cirospaciari/socketify.py/blob/62c4f1df00e819687d62bbb8c3f234ef8dcaa64a/src/socketify/socketify.py#L1598-L1603

Wasn't sure on how to handle the `None` case, so unconditionally raising seemed reasonable.
`handle_query` *can raise* on the last line here on main

https://github.com/uwdata/mosaic/blob/aff26ac56faa6d6a408209fd9b05c87c77f8d006/packages/server/duckdb-server/pkg/server.py#L82-L92

pyproject.toml
domoritz: well, learned something. Luckily, we can change things.

packages/vgplot/vgplot-python/vgplot/_types.py
domoritz: Yes, I'd vote for that

domoritz (approved):
Thank you!

dangotbanned:
Thanks for the speedy review @domoritz ❤️

PR #1140, opened 2026-08-11
perf: Avoid materializing `DuckDBPyRelation` for registration

### LINKED ISSUE #1105: widget: Investigate lazy `DuckDBPyRelation` registration
## Description
Follow-up to https://github.com/uwdata/mosaic/pull/1092#discussion_r3641479109

IIRC, `DuckDBPyRelation` is listed as an option in the error message for `register` (if you pass it something unsupported)
But currently `DuckDBPyRelation` is always collected to `pyarrow.Table` and then registered.

I'm thinking there *might* be a need for that if moving between connections? Perhaps not if we reuse the existing connection

dangotbanned:
Closes #1105

### DIFF
--- a/packages/vgplot/widget/mosaic_widget/frame_interop.py
+++ b/packages/vgplot/widget/mosaic_widget/frame_interop.py
@@ -29,7 +29,6 @@ def is_registrable_frame(obj: Any) -> TypeIs[IntoFrame]:
     )
 
 
-# TODO @dangotbanned: Avoid materializing `DuckDBPyRelation`
 def frame_to_duckdb_registrable(frame: IntoFrame) -> object:
     """Converts a native dataframe(-like) object to a DuckDB-registrable object.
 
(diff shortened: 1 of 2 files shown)


packages/vgplot/widget/mosaic_widget/frame_interop.py
```diff
@@ -47,19 +38,17 @@ def frame_to_duckdb_registrable(frame: IntoFrame) -> object:
 
     If the passed `frame` is a lazy frame, it is materialized.
     """
-    if _is_frame_native_to_duckdb(frame):
-        return frame
-
+    nw_frame = nw.from_native(frame)
+    if nw_frame.implementation in _DUCKDB_NATIVE:
+        return nw_frame.to_native()
     # If frame is not natively registrable to DuckDB, we convert it to an Arrow table via Narwhals.
```
dangotbanned, line 44: Are these comments/docs helpful for anyone?
It just looks like repeating what the code says 🤔

domoritz: I generally avoid comments that explain code since it's outdated too easily. But sometimes there is something surprising that's worth noting.

domoritz (approved):

PR #1141, opened 2026-08-11
chore(python): Enable & fix more ruff lints

dangotbanned:
## Description
Follow-up to (https://github.com/uwdata/mosaic/pull/1138#discussion_r3758263868) and (https://github.com/uwdata/mosaic/pull/1103#issuecomment-5106676636)

[`RUF10*`]: https://docs.astral.sh/ruff/rules/unused-noqa/

Addresses these points from #1088:

- [x] [Ruff-specific rules (RUF)](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf) like
    - [x] [`RUF10*`] rules which fix/flag unused/invalid suppression comments
- [x] [unsorted-imports](https://docs.astral.sh/ruff/rules/unsorted-imports/) ([Related](https://docs.astral.sh/ruff/settings/#lintisort))
    - [`required-imports = ["from __future__ import annotations"]`](https://docs.astral.sh/ruff/settings/#lint_isort_required-imports) combines with [this suggestion](https://github.com/uwdata/mosaic/pull/1067#discussion_r3530434157) to avoid circular dependencies, where the symbols are not used at runtime

And fixes all the new stuff that was enabled with https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set

## Related
- #1088


packages/vgplot/vgplot-python/tests/test_python_api.py
```diff
@@ -53,7 +53,7 @@ def test_date_param(self) -> None:
     def test_datetime_param(self) -> None:
         from datetime import datetime
 
-        assert vg.param(datetime(2013, 5, 13, 10, 30)).param_def() == {
+        assert vg.param(datetime(2013, 5, 13, 10, 30)).param_def() == {  # ruff: ignore[call-datetime-without-tzinfo]
```
dangotbanned, line 58: I thought this was worth discussing.

Would the JS side support both of these?

```py
"2013-05-13T10:30:00"
"2013-05-13T10:30:00+01:00"
```

aaaaaand does the answer change if the new fancy [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) thing is used?

## Related
- https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

packages/server/bench/bench.py
```diff
@@ -140,8 +144,9 @@ def http_post(host: str, port: int, payload: dict) -> bytes:
         )
         resp = conn.getresponse()
         data = resp.read()
-        if resp.status != 200:
-            raise RuntimeError(f"HTTP {resp.status}: {data[:200]}")
+        if resp.status != 200:  # ruff: ignore[magic-value-comparison]
```
dangotbanned, line 147: I've left these as individual ignores because this file is doing exactly as described in https://docs.astral.sh/ruff/rules/magic-value-comparison/

- `200` has a real meaning, (personally) I don't know which status it is though
- Lower down `3000` is used and would make more sense with a name
- `1024`, etc etc

packages/vgplot/vgplot-python/tests/test_python_api.py
domoritz: Temporal is not supported in safari yet. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse supports time zones.

packages/server/duckdb-server/pkg/__main__.py
```diff
@@ -11,7 +13,7 @@
 
 
 def serve() -> None:
-    db_path = sys.argv[1] if len(sys.argv) >= 2 else ":memory:"  # noqa: PLR2004
+    db_path = sys.argv[1] if len(sys.argv) >= 2 else ":memory:"  # ruff: ignore[magic-value-comparison]
```
domoritz, line 16: Should we just disable this check globally? Magic values are probably okay in most cases.

domoritz (approved):

packages/vgplot/vgplot-python/tests/test_full_round_trip.py
```diff
@@ -27,16 +29,18 @@ def run_spec_file_with_exec(path: Path) -> dict[str, Any]:
...
+        msg = f"`{path.name}` must define a top-level `view` variable"
+        raise AssertionError(msg)
     if not hasattr(view, "to_dict"):
-        raise AssertionError(f"`view` in `{path.name}` does not expose to_dict()")
+        msg = f"`view` in `{path.name}` does not expose to_dict()"
+        raise AssertionError(msg)
     # Execute to_dict() within the spec namespace so frame inspection finds the right variables
     namespace["__view__"] = view
-    exec("__result__ = __view__.to_dict()", namespace)
+    exec("__result__ = __view__.to_dict()", namespace)  # ruff: ignore[exec-builtin]
```
dangotbanned, line 43: Just highlighting this comment (https://github.com/uwdata/mosaic/pull/1141/commits/320fe1c9e00805dcd5f71b8da32ba6d9c3e3379e)

> IIRC, this means we can't use coverage,
> so I'd like to fix that eventually

packages/server/duckdb-server/pkg/__main__.py
dangotbanned: What does this code mean? 😄 

**Edit:** *more constructively* I'd find this case easier to understand if it used something like (https://docs.python.org/3/library/argparse.html)

Which I just learned was already being used here 😂 

https://github.com/uwdata/mosaic/blob/1649215d3ca5fa12c005661b0f5ad88471de891e/packages/server/bench/bench.py#L509-L514

packages/server/bench/bench.py
dangotbanned: Here's what fixing them would look like:

<details><summary>Show diff</summary>
<p>



```diff
diff --git a/packages/server/bench/bench.py b/packages/server/bench/bench.py
index 7a5eaca4..d75a7b7b 100755
--- a/packages/server/bench/bench.py
+++ b/packages/server/bench/bench.py
@@ -39,6 +39,7 @@ import statistics
 import subprocess
 import sys
 import time
+from http import HTTPStatus
 from pathlib import Path
 
 import websockets
@@ -144,7 +145,7 @@ def http_post(host: str, port: int, payload: dict) -> bytes:
         )
         resp = conn.getresponse()
         data = resp.read()
-        if resp.status != 200:  # ruff: ignore[magic-value-comparison]
+        if resp.status != HTTPStatus.OK:
             msg = f"HTTP {resp.status}: {data[:200]}"
             raise RuntimeError(msg)
         return data
@@ -223,13 +224,15 @@ async def bench_ws(
 # Reporting
 # ---------------------------------------------------------------------------
 
+KB = 1_024
+
 
 def fmt_bytes(n: int) -> str:
-    if n < 1024:  # ruff: ignore[magic-value-comparison]
+    if n < KB:
         return f"{n}B"
-    if n < 1024 * 1024:
-        return f"{n / 1024:.1f}K"
-    return f"{n / (1024 * 1024):.1f}M"
+    if n < KB * KB:
+        return f"{n / KB:.1f}K"
+    return f"{n / (KB * KB):.1f}M"
 
 
 HEADER_FMT = "  {:<40s} {:>6s} {:>8s} {:>8s} {:>8s} {:>8s}"
@@ -330,10 +333,15 @@ def build_and_start_go(port: int) -> subprocess.Popen | None:
     )
 
 
+SERVER_PORT = 3_000
+
+
 def build_and_start_python(port: int) -> subprocess.Popen | None:
     py_dir = SERVER_DIR / "duckdb-server"
-    if port != 3000:  # ruff: ignore[magic-value-comparison]
-        print("  WARNING: Python server does not support custom ports, using 3000")
+    if port != SERVER_PORT:
+        print(
+            f"  WARNING: Python server does not support custom ports, using {SERVER_PORT}"
+        )
     print("  Starting Python server ...")
     return subprocess.Popen(
         ["uv", "run", "duckdb-server"],
@@ -357,8 +365,10 @@ def build_and_start_node(port: int) -> subprocess.Popen | None:
         print(f"  Install failed:\n{result.stderr}")
         return None
 
-    if port != 3000:  # ruff: ignore[magic-value-comparison]
-        print("  WARNING: Node server does not support custom ports, using 3000")
+    if port != SERVER_PORT:
+        print(
+            f"  WARNING: Node server does not support custom ports, using {SERVER_PORT}"
+        )
     print("  Starting Node server ...")
     return subprocess.Popen(
         ["node", str(node_dir / "bin" / "run-server.js")],
@@ -453,10 +463,13 @@ def run_benchmarks(
     return results
 
 
+_COMPARABLE = 2
+
+
 def print_comparison(all_results: dict[str, ServerResults]) -> None:
     """Print a side-by-side comparison of all servers."""
     servers = list(all_results.keys())
-    if len(servers) < 2:  # ruff: ignore[magic-value-comparison]
+    if len(servers) < _COMPARABLE:
         return
 
     # Collect which transports were benchmarked
@@ -469,7 +482,7 @@ def print_comparison(all_results: dict[str, ServerResults]) -> None:
 
         # Only include servers that have results for this transport
         active = [s for s in servers if transport in all_results[s]]
-        if len(active) < 2:  # ruff: ignore[magic-value-comparison]
+        if len(active) < _COMPARABLE:
             continue
 
         labels = list(all_results[active[0]][transport].keys())
@@ -516,7 +529,7 @@ def print_comparison(all_results: dict[str, ServerResults]) -> None:
 
 def main() -> None:
     parser = argparse.ArgumentParser(description="Mosaic Server Benchmark")
-    parser.add_argument("-p", "--port", type=int, default=3000)
+    parser.add_argument("-p", "--port", type=int, default=SERVER_PORT)
     parser.add_argument("-n", "--iterations", type=int, default=100)
     parser.add_argument("-w", "--warmup", type=int, default=5)
     parser.add_argument(

```



</p>
</details> 

I wasn't expecting the `3000` to show up in another place, but now all 3 are derived from the same constant 🙂

PR #1143, opened 2026-08-12
fix(sql): add missing DuckDB aggregate names, guard against future drift

### LINKED ISSUE #1142: `aggregateNames` is missing 8 aggregate functions present in current DuckDB
`aggregateNames` in `packages/mosaic/sql/src/ast/aggregate.ts` has drifted from current DuckDB (verified against v1.5.5 via `duckdb_functions()`). Eight genuine aggregate functions are missing from the list:

`argmax`, `argmin`, `arg_max_nulls_last`, `arg_min_nulls_last`, `count_if`, `group_concat`, `listagg`, `sem`

When one of these appears in a verbatim SQL expression, `isAggregateExpression` fails to detect it, so downstream query construction (vgplot `markQuery`, core preaggregation, query consolidation) treats the aggregate as a grouping dimension and emits it into `GROUP BY`:

```js
vg.plot(
  vg.barY(vg.from("orders"), {
    x: "region",
    y: vg.count(),
    fill: vg.sql`argmax(product, price)`   // or group_concat / count_if / sem / listagg
  })
)
```

generates

```sql
SELECT argmax(product, price) AS "fill", count(*) AS "y", "region" AS "x"
FROM "orders" GROUP BY "fill", "x"
```

which DuckDB rejects with:

```
Binder Error: GROUP BY clause cannot contain aggregates!
```

This is a recurrence of the drift pattern previously fixed in #1051 (`mean` et al.). Since the list mirrors a moving target, a fix should also add a guard so the drift is caught mechanically the next time the DuckDB dependency is bumped, rather than by user bug reports.

Notes from the analysis: DuckDB also reports 13 window-only functions (`rank`, `row_number`, …) and the internal `sum_no_overflow` as `function_type = 'aggregate'` — those are deliberately not classification targets. In the other direction, `geomean`, `geometric_mean`, `weighted_avg`, and `histogram_values` are now macro-backed in DuckDB; the first three still expand to aggregates so keeping them listed remains correct.

*Found by an automated audit of the SQL layer run with Claude; the repro above is binder-verified against DuckDB 1.5.5.*

domoritz:
Fixes #1142.

> **Note:** This fix was created by Claude (an agent team ran a binder-error audit of the SQL layer; the fix went through automated implementation, simplification, and adversarial review passes).

## What changed

**Commit 1 — the fix.** Adds the 8 missing names to `aggregateNames` (alphabetical placement): `arg_max_nulls_last`, `arg_min_nulls_last`, `argmax`, `argmin`, `count_if`, `group_concat`, `listagg`, `sem`. Regression tests in `test/visitors.test.ts` ("Verbatim aggregate detection") build a markQuery-shaped query per name and assert via `toBeValidQuery` that no `GROUP BY` is emitted over the aggregate alias and the result binds in DuckDB.

**Commit 2 — the guard.** New `test/aggregate-names.test.ts` diffs `aggregateNames` in both directions against `duckdb_functions() WHERE function_type = 'aggregate'` on the live test-fixture DuckDB, so this drift class (#1051, and now #1142) is caught mechanically whenever the DuckDB dependency is bumped instead of by user bug reports. Two commented exclusion sets carry the intent: window-only functions that DuckDB reports as aggregates (usable only with `OVER`, which `isAggregateExpression` handles separately) plus internal `sum_no_overflow`, and macro-backed names (`geomean`, `geometric_mean`, `histogram_values`, `weighted_avg`) that still expand to aggregates. Failure messages print the exact ready-to-paste arrays. A minimal `queryFixture` helper was added to `test/util/validate.ts`.

Deliberately **not** added: the window-only pseudo-aggregates and `sum_no_overflow` (see exclusion rationale in the test).

(“F-FN-01” in the commit messages is the internal finding ID from the audit.)


packages/mosaic/sql/test/aggregate-names.test.ts
```diff
@@ -0,0 +1,64 @@
...
+// classifying them as aggregates remains correct.
+const macroBacked = new Set([
+  'geomean',
+  'geometric_mean',
+  'histogram_values',
+  'weighted_avg'
+]);
+
+describe('aggregateNames', () => {
+  it('matches the aggregates in the fixture DuckDB', async () => {
```
domoritz, line 38: This test will ensure that we have the right list of aggregates even in the future as duckdb changes.

dangotbanned: I haven't seen it mentioned anywhere in the PR, but FYI a lot of these names are aliases for functions you already have.

There's a breakdown here, which IIUC is based on the output of `duckdb_functions()`

- https://duckdb.org/docs/current/sql/functions/aggregates#general-aggregate-functions


One question I have is about what the current situation looks like for maintenance?
Is the loop intended to be this?

1. Everything is fine
2. DuckDB changes something and CI fails
3. Manually add things again
4. Go to step 1

If the effort is just adding names to a list, perhaps you could trigger that as job to update in CI on a schedule (by running the query again), instead of waiting for a test to fail?

domoritz: We specify a specific version that mosaic depends on. If we change that dependency and there is a new function, ci fails (the loop you describe). I don't think we need to regularly run the ci since this only matters once we raise the version we specify in the dependencies.

derekperkins:
When I was just working on the [function allowlist](https://github.com/uwdata/mosaic/pull/1150), `duckdb_functions()` wasn't quite enough for a full inventory, as several functions are classified as macros. I ended up creating a dedicated [AGENTS.md](https://github.com/uwdata/mosaic/blob/main/packages/server/duckdb-server-go/pkg/functionset/AGENTS.md) to assist with future updates.

That being said, this is probably sufficient here. I mentioned in my original allowlist PR, it might be useful to put function information in json format elsewhere so any package can look at the function introspection. Right now every core and core extension function is enumerated in the go server, split into compute-only and elevated permissions groups.

jheer (approved):

domoritz:
@derekperkins good points. Let's merge this for now and when we have a good inventory of functions, we can refer to that from a test similar to how we use `duckdb_functions()`.

PR #1145, opened 2026-08-12
fix(sql): make verbatim window detection whitespace-tolerant

### LINKED ISSUE #1144: Verbatim window expressions misclassified as aggregates unless `over` has single-space form
`isAggregateExpression` (`packages/mosaic/sql/src/visit/visitors.ts`) treats a verbatim expression as windowed — and therefore *not* an aggregate — only if it contains the exact lowercase text `') over '`, one space on each side. Users routinely write `OVER(PARTITION BY ...)` with no space, a newline before `OVER`, or extra whitespace. In those cases the expression is misclassified as an aggregate, so vgplot's `markQuery` omits the field from `GROUP BY` while grouping the other dimensions, and DuckDB's binder rejects the query (core's preaggregation is affected the same way):

```js
vg.plot(
  vg.lineY(vg.from("stocks"), {
    x: "date",
    y: vg.sql`avg(price) OVER(PARTITION BY symbol)`,  // no space before "("
    stroke: "symbol"
  })
)
```

generates

```sql
SELECT avg(price) OVER(PARTITION BY symbol) AS "y", "date" AS "x", symbol AS "stroke"
FROM "stocks" GROUP BY "x", "stroke"
```

```
Binder Error: column "price" must appear in the GROUP BY clause or must be part of an aggregate function.
```

The identical spec written `OVER (PARTITION BY symbol)` — single space — works, so this is purely whitespace sensitivity in the heuristic. It is distinct from #259 (true window + aggregate mixes), which fails for a different reason regardless of spacing.

*Found by an automated audit of the SQL layer run with Claude; repro binder-verified against DuckDB 1.5.5.*

domoritz:
Fixes #1144.

> **Note:** This fix was created by Claude (an agent team ran a binder-error audit of the SQL layer; the fix went through automated implementation, simplification, and adversarial review passes, plus review notes from @domoritz).

## What changed

The literal `s.includes(') over ')` check in `isAggregateExpression` is replaced by a small regex:

```ts
// regexp to match window function calls with inline or named definitions
const windowRegExp = /\)\s*over(\s*\(|\s+[\w"])/;
```

It tolerates arbitrary whitespace (including none) around `over` and also matches the named-window form (`... over win` / `... over "win"`). Requiring `(` or whitespace-plus-identifier after `over` avoids false positives on `OVERLAPS` / `overlay(...)` — which the old literal check also didn't match.

Regression tests in `test/visitors.test.ts`: the three whitespace variants plus a single-space control, each asserting `isAggregateExpression(...) === 0` and binder-validating the generated markQuery-shaped statement via `toBeValidQuery` (no `GROUP BY` emitted), plus a named-window case using a `WINDOW` clause.

Out of scope (unchanged behavior): a correctly-classified window expression combined with a real aggregate in the same mark still hits vgplot's `markQuery` limitation tracked in #259.

(“F-FN-02” in the commit message is the internal finding ID from the audit.)

### DIFF
--- a/packages/mosaic/sql/src/visit/visitors.ts
+++ b/packages/mosaic/sql/src/visit/visitors.ts
@@ -14,6 +14,9 @@ const aggrRegExp = new RegExp(`^(${aggregateNames.join('|')})$`);
 // function call tokens will have a pattern like "name(".
 const funcRegExp = /(\\'|\\"|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|\w+\()/g;
 
+// regexp to match window function calls with inline or named definitions
+const windowRegExp = /\)\s*over(\s*\(|\s+[\w"])/;
+
 function hasVerbatimAggregate(s: string) {
   return s
     .split(funcRegExp)
@@ -46,7 +49,7 @@ export function isAggregateExpression(root: SQLNode) {
         if (sub >= 0) s = s.slice(0, sub);
 
         // exit if expression includes windowing
-        if (s.includes(') over ')) return -1;
+        if (windowRegExp.test(s)) return -1;
         if (hasVerbatimAggregate(s)) {
           agg |= 2;
           return -1;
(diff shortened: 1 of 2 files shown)


dangotbanned:
I'm curious if this bug extends beyond `over`?

[Case insensitivity](https://duckdb.org/docs/current/sql/dialect/keywords_and_identifiers) and [insignificant whitespace](https://en.wikipedia.org/wiki/SQL#Syntax) show up in more places than that.
Might be worth looking into anywhere else regex and sql collide 😅

domoritz:
I've got some more issues/prs in the queue. Thanks for flagging this. I'll check the other regexes.

jheer (approved):

Issue #662, opened 2025-01-31
Issue Querying MotherDuck After v0.12.2 Upgrade


enlore:
# Summary

## Dependencies

Dependency versions:

```
  "dependencies": {
    "@motherduck/wasm-client": "^0.6.6",
    "@uwdata/vgplot": "^0.12.2",
    "apache-arrow": "^17.0.0",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"
  },
```

```
➜  debug-motherduck-mosaic-v_0_12_2 git:(main)  node --version
v21.0.0
```

## Errors

When attempting to build plots using data fetched from MotherDuck, one of two errors occurs.

ERROR 1: `getFieldInfo` throws:

```
field-info.js:75 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'column_type')
   at getFieldInfo (field-info.js:75:19)
   at async Promise.all (:5173/index 1)
   at async queryFieldInfo (field-info.js:50:13)
   at async Coordinator.initializeClient (Coordinator.js:251:24)
```

This happens when attempting to build a plot by querying data from MotherDuck.

ERROR 2: `arrowToColumns` throws:

```
hook.js:608 TypeError: data.toColumns is not a function
  at arrowToColumns (to-data-columns.js:35:35)
  at toDataColumns (to-data-columns.js:24:7)
  at ConnectedMark.queryResult (Mark.js:177:17)
  at Coordinator.js:207:24
```

While exploring this issue, I validated that I can execute `DESCRIBE` queries against the plot's source table directly using `vg.coordinator().query()`.

If I execute a `DESCRIBE` query with the same exact quotes as generated by Mosaic's query building logic before attempting to create the plot, this error is thrown instead of ERROR 1. See `src/App.tsx` line 28 for sample code.

# Repro

See reference repo here: https://github.com/enlore/debug-motherduck-mosaic-v_0_12_2

It should be good to go so you can clone, `npm i`, and `npm run dev`.

The repo includes a read only token so the app can connect to my personal MotherDuck free tier account for convenience.

* `src/App.tsx` contains the relevant logic
* Set `CONNECT_TO_MOTHERDUCK` to `true` to see ERROR 1
* Uncomment line 45 to see ERROR 2
* Set `CONNECT_TO_MOTHERDUCK` to `false` to validate that Mosaic can build a line plot from a browser local DuckDB Wasm instance without issue
* DuckDB version number is logged to the console for both code paths

[...]

PR #1146, opened 2026-08-13
fix: Issue Querying Motherduck

sittingthyme:
**1. Empty `DESCRIBE` result** — `packages/mosaic/core/src/util/field-info.ts`

`getFieldInfo` wrapped the describe query in a `try/catch`, which covers a *thrown*
query but not a *successful query that returns zero rows*. In that case
`[desc] = Array.from(...)` leaves `desc` undefined and the next line dereferences it:

```
TypeError: Cannot read properties of undefined (reading 'column_type')
  at getFieldInfo (field-info.js:75)
```

The dummy description is now a module constant applied via `desc ??= UNKNOWN_COLUMN`,
so it covers the empty result and the pre-existing window-aggregate failure path alike.

Related upstream: duckdb/duckdb#16017, where `DESCRIBE` discarded renamed columns.
Mosaic's field-info query always aliases `AS "column"`, so it sits directly on that
path — but Mosaic should not crash on an empty result regardless of the backend.

**2. apache-arrow tables** — `packages/mosaic/core/src/util/to-data-columns.ts`

`isArrowTable` duck-types on `getChild` alone, so an `apache-arrow` `Table` passes the
check, but `toColumns()` is Flechette-only:

```
TypeError: data.toColumns is not a function
  at arrowToColumns (to-data-columns.js:35)
```

`arrowToColumns` now calls `toColumns()` when present and otherwise reads columns via
`schema.fields` + `getChild(name).toArray()`, which both implementations support. A
local `ArrowLikeTable` structural type keeps `tsc --build` clean without scattering
casts. The loose `isArrowTable` check is deliberately unchanged — it is what routes
apache-arrow tables into the new branch.

Relates to #662

### DIFF
--- a/packages/mosaic/core/src/util/field-info.ts
+++ b/packages/mosaic/core/src/util/field-info.ts
@@ -11,6 +11,12 @@ export const Min = 'min';
 export const Distinct = 'distinct';
 export const Stats = { Count, Nulls, Max, Min, Distinct };
 
+const UNKNOWN_COLUMN: ColumnDescription = {
+  column_name: 'column',
+  column_type: 'DOUBLE',
+  null: 'YES'
+};
+
 const statMap: Record<Stat, (column: FieldRef) => AggregateNode> = {
   [Count]: count,
   [Distinct]: column => count(column).distinct(),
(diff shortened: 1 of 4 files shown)


domoritz:
Updated in #1148

PR #1150, opened 2026-08-13
feat(go-server): add function allowlist

derekperkins:
Actually enforcing an allowlist is easy, especially since we already had the blocklist implementation. The majority of this PR was building an inventory of function names of both core and core extensions, categorizing them as either compute only and thus safe / side-effect free, and elevated risk functions that might read across the network, have side-effects, etc. We're including all compute-only functions in the default allowlist and none of the elevated risk functions.

If there was interest for the other servers to implement something similar, we could move the inventory to a separate json file they could all consume. For upgrades, the worst case scenario of drift is that newly introduced functions aren't automatically included in the default, but can be trivially added by the user. I added a specific AGENTS.md file in that directory so future agents can update the list.

I'm also clarifying that the CLI here is not intended to support 100% of the custom options. It's built to provide a decent experience out of the box, but for production use cases, the expectation is users will provide their own main.go implementation, importing these libraries.

PR #1152, opened 2026-08-14
feat(go-server): reject remote URI literals

derekperkins:
Similar to #1150 AI did a deep dive to find all the functions in core + core extensions that are eligible to be fed into httpfs for remote reads. This then does a case insensitive comparison looking for `gcs://`, `https://`, etc., in those specific parameters. There are ways around it, some mentioned here, so this isn't a perfect way around remote reads, but as a best-effort I think it's decent.

This is orthogonal to the allowlist, so if you truly want to stop remote reads while still querying remote filesystems, you can attach a db/table and disallow all the file read functions.

## Summary

- add a source-audited, DuckDB 1.5.5-pinned inventory of 58 functions whose reviewed arguments can open caller-controlled paths
- add `query.WithRemoteURILiteralRejection()` to reject recognized remote URI literals case-insensitively in replacement scans and decoded literals within those path arguments
- preserve local paths, unrelated literals, and trusted pre-attached catalogs while documenting the best-effort boundary and disabling unvalidated `exec` commands

### DIFF
--- a/docs/server/index.md
+++ b/docs/server/index.md
@@ -10,32 +10,6 @@ This package provides a local DuckDB server. To instead use DuckDB-WASM in the b
 DuckDB can also connect to and query other databases, such as PostgreSQL and MySQL. See the [multi-database support page](/api/core/multi-database-support) for examples.
 :::
 
-## Go Server
-
-Mosaic also provides an embeddable [DuckDB Go server](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go).
-Applications that need local file readers but want to reject common caller-supplied remote locations can enable the strict
-query option:
-
-```go
-query.WithRemoteURILiteralRejection()
-```
-
-The option rejects recognized remote URI literals in DuckDB replacement scans and
-[reviewed function path arguments](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go/pkg/functionset/remoteread),
-including literal lists and literals within path expressions. Matching is case-insensitive and uses DuckDB 1.5.5's
-prefixes: `http://`, `https://`, `s3://`, `s3a://`, `s3n://`, `gcs://`, `gs://`, `r2://`, `hf://`, `azure://`, `az://`,
-and `abfss://`. Unrelated string literals remain unaffected. Ordinary local paths remain usable unless the path string
-itself contains a recognized marker. Trusted connector initialization can load extensions and attach remote catalogs
-before serving queries; subsequent queries against those catalogs remain usable.
-
-DuckDB's serialized AST cannot distinguish a replacement-scan string from a quoted table or CTE identifier, so URI-shaped
-identifiers are rejected too. This check is best-effort, not a network sandbox. Split or computed strings, macros, views,
-nested SQL, extension-defined schemes, GDAL virtual paths, remote locations reached through local Iceberg or Delta
-metadata, and SQL stored in SQLite views can evade it. Enabling the option also rejects `exec` commands, leaving only
-validated query forms. See the Go server's
-[security documentation](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go#remote-uri-literal-policy)
-for the full boundary.
-
 ## Usage
 
 The server package is available on [PyPi](https://pypi.org/project/duckdb-server/).
@@ -51,3 +25,9 @@ Alternatively, you can install the server with `pip install duckdb-server`. Then
 ## Developer Setup
 
 To run the server from the Mosaic repository and to run the server in development mode, follow the [instructions for the duckdb-server package](https://github.com/uwdata/mosaic/blob/main/packages/server/duckdb-server/README.md).
+
+## Go Server
+
+Mosaic also provides an embeddable [DuckDB Go server](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go).
+See its README for the Go API, query authorization options, and the best-effort
+[remote URI literal policy](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go#remote-uri-literal-policy).
--- a/docs/server/index.md
+++ b/docs/server/index.md
@@ -22,11 +22,11 @@ query.WithRemoteURILiteralRejection()
 
 The option rejects recognized remote URI literals in DuckDB replacement scans and
 [reviewed function path arguments](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go/pkg/functionset/remoteread),
-including literal lists and literals within path expressions. It uses DuckDB 1.5.5's prefixes: `http://`, `https://`,
-`s3://`, `s3a://`, `s3n://`, `gcs://`, `gs://`, `r2://`, `hf://`, `azure://`, `az://`, and `abfss://`. Unrelated string
-literals remain unaffected. Ordinary local paths remain usable unless the path string itself contains a recognized marker.
-Trusted connector initialization can load extensions and attach remote catalogs before serving queries; subsequent queries
-against those catalogs remain usable.
+including literal lists and literals within path expressions. Matching is case-insensitive and uses DuckDB 1.5.5's
+prefixes: `http://`, `https://`, `s3://`, `s3a://`, `s3n://`, `gcs://`, `gs://`, `r2://`, `hf://`, `azure://`, `az://`,
+and `abfss://`. Unrelated string literals remain unaffected. Ordinary local paths remain usable unless the path string
+itself contains a recognized marker. Trusted connector initialization can load extensions and attach remote catalogs
+before serving queries; subsequent queries against those catalogs remain usable.
 
 DuckDB's serialized AST cannot distinguish a replacement-scan string from a quoted table or CTE identifier, so URI-shaped
 identifiers are rejected too. This check is best-effort, not a network sandbox. Split or computed strings, macros, views,
--- a/docs/server/index.md
+++ b/docs/server/index.md
@@ -25,9 +25,3 @@ Alternatively, you can install the server with `pip install duckdb-server`. Then
 ## Developer Setup
 
 To run the server from the Mosaic repository and to run the server in development mode, follow the [instructions for the duckdb-server package](https://github.com/uwdata/mosaic/blob/main/packages/server/duckdb-server/README.md).
-
-## Go Server
-
-Mosaic also provides an embeddable [DuckDB Go server](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go).
-See its README for the Go API, query authorization options, and the best-effort
-[remote URI literal policy](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go#remote-uri-literal-policy).
--- a/docs/server/index.md
+++ b/docs/server/index.md
@@ -10,6 +10,32 @@ This package provides a local DuckDB server. To instead use DuckDB-WASM in the b
 DuckDB can also connect to and query other databases, such as PostgreSQL and MySQL. See the [multi-database support page](/api/core/multi-database-support) for examples.
 :::
 
+## Go Server
+
+Mosaic also provides an embeddable [DuckDB Go server](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go).
+Applications that need local file readers but want to reject common caller-supplied remote locations can enable the strict
+query option:
+
+```go
+query.WithRemoteURILiteralRejection()
+```
+
+The option rejects recognized remote URI literals in DuckDB replacement scans and
+[reviewed function path arguments](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go/pkg/functionset/remoteread),
+including literal lists and literals within path expressions. It uses DuckDB 1.5.5's prefixes: `http://`, `https://`,
+`s3://`, `s3a://`, `s3n://`, `gcs://`, `gs://`, `r2://`, `hf://`, `azure://`, `az://`, and `abfss://`. Unrelated string
+literals remain unaffected. Ordinary local paths remain usable unless the path string itself contains a recognized marker.
+Trusted connector initialization can load extensions and attach remote catalogs before serving queries; subsequent queries
+against those catalogs remain usable.
+
+DuckDB's serialized AST cannot distinguish a replacement-scan string from a quoted table or CTE identifier, so URI-shaped
+identifiers are rejected too. This check is best-effort, not a network sandbox. Split or computed strings, macros, views,
+nested SQL, extension-defined schemes, GDAL virtual paths, remote locations reached through local Iceberg or Delta
+metadata, and SQL stored in SQLite views can evade it. Enabling the option also rejects `exec` commands, leaving only
+validated query forms. See the Go server's
+[security documentation](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go#remote-uri-literal-policy)
+for the full boundary.
+
 ## Usage
 
 The server package is available on [PyPi](https://pypi.org/project/duckdb-server/).
--- a/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
+++ b/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
@@ -38,12 +38,13 @@ var inventoryGroups = [...]inventoryGroup{
 var inventory = mergeInventories(inventoryGroups[:])
 
 // Lookup returns the reviewed path arguments for a function name.
+// The returned slices alias package state and must not be mutated.
 func Lookup(name string) (PathArguments, bool) {
 	arguments, ok := inventory[strings.ToLower(strings.TrimSpace(name))]
 	if !ok {
 		return PathArguments{}, false
 	}
-	return cloneArguments(arguments), true
+	return arguments, true
 }
 
 // FunctionNames returns the sorted reviewed function names.
@@ -70,15 +71,8 @@ func mergeInventories(groups []inventoryGroup) map[string]PathArguments {
 			if _, ok := functions[name]; ok {
 				panic("duplicate remote-read function: " + name)
 			}
-			functions[name] = cloneArguments(arguments)
+			functions[name] = arguments
 		}
 	}
 	return functions
-}
-
-func cloneArguments(arguments PathArguments) PathArguments {
-	return PathArguments{
-		Positional: slices.Clone(arguments.Positional),
-		Named:      slices.Clone(arguments.Named),
-	}
 }
--- a/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
+++ b/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
@@ -1,0 +1,83 @@
+package remoteread
+
+import (
+	"slices"
+	"strings"
+)
+
+// PathArguments identifies arguments that can carry a URI or path.
+type PathArguments struct {
+	// Positional contains zero-based indexes among unnamed SQL arguments.
+	Positional []int
+	// Named contains lowercase DuckDB argument names.
+	Named []string
+}
+
+type inventoryGroup struct {
+	name      string
+	functions map[string]PathArguments
+}
+
+var inventoryGroups = [...]inventoryGroup{
+	{name: "avro", functions: avroFunctions},
+	{name: "core", functions: coreFunctions},
+	{name: "delta", functions: deltaFunctions},
+	{name: "ducklake", functions: duckLakeFunctions},
+	{name: "excel", functions: excelFunctions},
+	{name: "iceberg", functions: icebergFunctions},
+	{name: "json", functions: jsonFunctions},
+	{name: "lance", functions: lanceFunctions},
+	{name: "parquet", functions: parquetFunctions},
+	{name: "postgres", functions: postgresFunctions},
+	{name: "spatial", functions: spatialFunctions},
+	{name: "sqlite", functions: sqliteFunctions},
+	{name: "vortex", functions: vortexFunctions},
+}
+
+var inventory = mergeInventories(inventoryGroups[:])
+
+// Lookup returns the reviewed path arguments for a function name.
+func Lookup(name string) (PathArguments, bool) {
+	arguments, ok := inventory[strings.ToLower(strings.TrimSpace(name))]
+	if !ok {
+		return PathArguments{}, false
+	}
+	return cloneArguments(arguments), true
+}
+
+// FunctionNames returns the sorted reviewed function names.
+func FunctionNames() []string {
+	names := make([]string, 0, len(inventory))
+	for name := range inventory {
+		names = append(names, name)
+	}
+	slices.Sort(names)
+	return names
+}
+
+func positionalPath(index int, named ...string) PathArguments {
+	return PathArguments{
+		Positional: []int{index},
+		Named:      named,
+	}
+}
+
+func mergeInventories(groups []inventoryGroup) map[string]PathArguments {
+	functions := make(map[string]PathArguments)
+	for _, group := range groups {
+		for name, arguments := range group.functions {
+			if _, ok := functions[name]; ok {
+				panic("duplicate remote-read function: " + name)
+			}
+			functions[name] = cloneArguments(arguments)
+		}
+	}
+	return functions
+}
+
+func cloneArguments(arguments PathArguments) PathArguments {
+	return PathArguments{
+		Positional: slices.Clone(arguments.Positional),
+		Named:      slices.Clone(arguments.Named),
+	}
+}
--- a/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
+++ b/packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
@@ -19,6 +19,7 @@ type inventoryGroup struct {
 }
 
 var inventoryGroups = [...]inventoryGroup{
+	{name: "autocomplete", functions: autocompleteFunctions},
 	{name: "avro", functions: avroFunctions},
 	{name: "core", functions: coreFunctions},
 	{name: "delta", functions: deltaFunctions},
--- a/packages/server/duckdb-server-go/pkg/query/remote_uri.go
+++ b/packages/server/duckdb-server-go/pkg/query/remote_uri.go
@@ -9,7 +9,7 @@ import (
 	"github.com/uwdata/mosaic/packages/server/duckdb-server-go/pkg/functionset/remoteread"
 )
 
-// Source: https://github.com/duckdb/duckdb/blob/v1.5.5/src/include/duckdb/main/extension_entries.hpp#L1275-L1280
+// Sources are the DuckDB 1.5.5-pinned filesystem handlers listed in ../functionset/remoteread/README.md.
 var remoteURIPrefixes = [...]string{
 	"http://",
 	"https://",
@@ -22,7 +22,13 @@ var remoteURIPrefixes = [...]string{
 	"hf://",
 	"azure://",
 	"az://",
+	"abfs://",
 	"abfss://",
+}
+
+var nestedSQLExecutors = [...]string{
+	"json_execute_serialized_sql",
+	"query",
 }
 
 type remoteURILiteralValidator struct {
@@ -65,8 +71,13 @@ func (v *remoteURILiteralValidator) checkTableFunction(node map[string]any) {
 		v.errs = append(v.errs, fmt.Errorf("query: invalid remote URI table function node: expected string function_name: %v", function["function_name"]))
 		return
 	}
+	functionName = strings.ToLower(functionName)
+	if slices.Contains(nestedSQLExecutors[:], functionName) {
+		v.errs = append(v.errs, fmt.Errorf("%w: nested SQL executor '%s' is not allowed", ErrAccessDenied, functionName))
+		return
+	}
 
-	pathArguments, ok := remoteread.Lookup(strings.ToLower(functionName))
+	pathArguments, ok := remoteread.Lookup(functionName)
 	if !ok {
 		return
 	}
--- a/packages/server/duckdb-server-go/pkg/query/remote_uri.go
+++ b/packages/server/duckdb-server-go/pkg/query/remote_uri.go
@@ -26,10 +26,12 @@ var remoteURIPrefixes = [...]string{
 	"abfss://",
 }
 
-var nestedSQLExecutors = [...]string{
+var nestedSQLTableExecutors = [...]string{
 	"json_execute_serialized_sql",
 	"query",
 }
+
+const nestedSQLScalarExecutor = "json_serialize_plan"
 
 type remoteURILiteralValidator struct {
 	errs []error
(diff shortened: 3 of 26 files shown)


derekperkins (commented):
Reviewed the validator, the inventory, and the docs. The inventory work is careful — I spot-checked it against a local DuckDB v1.5.5 `d8cdaa33fd` (the exact revision pinned in `remoteread/README.md`) and the selectors I probed hold up, including the non-obvious `histogram(source := ...)` named form and the `parquet_bloom_probe` position-1 exclusion. Case-sensitive matching is also the right call: `HTTPS://` is not handled by httpfs even when it is explicitly loaded, so the test that pins that is correct.

Three findings on the boundary itself, one docs-structure comment, and one nit, in threads below. The `abfs://` gap and the `query()` bypass are the two I'd want resolved before merge; the fail-closed suggestion is a judgment call that is already partly documented.

All findings were reproduced against this branch, not inferred.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
```diff
@@ -0,0 +1,148 @@
...
+	"s3a://",
+	"s3n://",
+	"gcs://",
+	"gs://",
+	"r2://",
+	"hf://",
+	"azure://",
+	"az://",
+	"abfss://",
+}
```
derekperkins, line 27: **`abfs://` is missing from this list.**

The cited source is DuckDB's *autoload* prefix map, which only decides which prefixes trigger auto-installation of an extension. It is not the set of prefixes a loaded filesystem will handle. `duckdb-azure` registers both `abfss://` and `abfs://` on its DFS filesystem, but only `abfss://` appears in the autoload map — so `abfs://` never autoloads, and consequently never made it into this list.

That matters because the README documents trusted initialization loading filesystem extensions as a supported deployment. In exactly that deployment (`LOAD azure` at startup), on v1.5.5 `d8cdaa33fd`:

```
$ duckdb -c "LOAD azure; SELECT * FROM read_parquet('abfs://acct.dfs.core.windows.net/c/x.parquet')"
IO Error: AzureStorageFileSystem open file 'abfs://acct.dfs.core.windows.net/c/x.parquet' failed with code 'NoAuthenticationInformation' ...
```

and both of these pass validation on this branch today:

```
SELECT * FROM read_parquet('abfs://acct.dfs.core.windows.net/c/x.parquet')   -> err=<nil>
SELECT * FROM 'abfs://acct.dfs.core.windows.net/c/x.parquet'                 -> err=<nil>
```

I probed all twelve listed prefixes plus `abfs://`, `cos://`, `oss://`, `ftp://`, `file://`, and `ducklake://`. The twelve are all genuine handlers and `abfs://` is the only real one missing — so this is a one-line fix, not a broad re-audit.

Two process points that follow from the same root cause:

1. Deriving the list from `CanHandleFile` in each filesystem, rather than from `extension_entries.hpp`, would have caught this and will keep catching it.
2. `remoteread/AGENTS.md` tells the next upgrade to re-audit the inventory and replacement scans, but says nothing about this prefix list — which lives in a different package from the README that pins its DuckDB revision. Worth adding to that checklist so the two pins can't drift.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
```diff
@@ -0,0 +1,148 @@
...
+	functionName, ok := function["function_name"].(string)
+	if !ok {
+		v.errs = append(v.errs, fmt.Errorf("query: invalid remote URI table function node: expected string function_name: %v", function["function_name"]))
+		return
+	}
+
+	pathArguments, ok := remoteread.Lookup(strings.ToLower(functionName))
+	if !ok {
+		return
+	}
```
derekperkins: **Two core built-ins execute nested SQL and fall through this `Lookup` miss, which turns the whole option off.**

Neither needs an extension; both are available in a stock DuckDB. Against this branch:

```
SELECT * FROM query('SELECT * FROM read_csv(''https://example.com/x.csv'')')
  -> err=<nil>
SELECT * FROM json_execute_serialized_sql(json_serialize_sql('SELECT * FROM read_csv(''https://example.com/x.csv'')'))
  -> err=<nil>
```

Both really do execute the inner statement (verified reading a local CSV through each), and both serialize cleanly, so `ValidateSQL` sees only the harmless outer `SELECT`.

I recognize `remoteread/README.md` says the review "deliberately excludes nested SQL functions" and the Go README lists "nested SQL strings are not recursively inspected". Those read as caveats about *depth of inspection* — that a URI hidden inside a stored view body won't be found. They don't convey that a caller can disable the control outright with one core function call. And these two aren't path sinks, so they can't be fixed by adding them to a path-argument inventory; they need handling in the validator.

`functionset.DefaultFunctions()` already excludes `query`, `query_table`, and `json_execute_serialized_sql`, so an allowlisted deployment is covered. But this option is documented and tested as standalone — `TestDBRemoteURILiteralRejection` configures nothing else — and the docs pitch it at "applications that need local file readers", which is precisely the shape that opts out of the default allowlist.

Either would resolve this for me:

- reject `query` and `json_execute_serialized_sql` outright when the option is enabled (they take SQL, not a path, so there is nothing to inspect); or
- state on `WithRemoteURILiteralRejection`'s doc comment that it is only meaningful alongside a function policy denying nested-SQL execution, and pin that pairing with a test.

I'd prefer the first — the second leaves a standalone-safe-looking option that isn't.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		literal, ok := value["value"].(string)
+		if !ok {
+			return
+		}
+		if prefix, ok := findRemoteURIPrefix(literal); ok {
+			v.reject(prefix, fmt.Sprintf("path argument to function '%s'", strings.ToLower(functionName)))
+		}
+	})
+}
```
derekperkins, line 149: **Suggestion: also fail closed on non-constant nodes inside a reviewed path argument.**

The README does document `'gc' || 's://bucket/file.parquet'` as a known evasion, so this is a judgment call rather than an oversight. But it's worth being explicit that DuckDB constant-folds table-function arguments at bind time, so this isn't a theoretical gap — the read actually happens:

```
$ duckdb -c "SELECT * FROM read_parquet('htt' || 'ps://invalid.invalid/x.parquet')"
IO Error: Could not resolve hostname error for HTTP HEAD to 'https://invalid.invalid/x.parquet'
```

All three of these pass validation on this branch, and the folding step means each is a live remote fetch:

```
read_parquet('htt' || 'ps://example.com/x.parquet')
read_parquet(concat('htt', 'ps://example.com/x.parquet'))
read_parquet(format('{}://example.com/x.parquet', 'https'))
```

Since `walkValues` is already traversing this subtree, you can close all of them by *adding* a rejection alongside the existing constant scan: any node in a reviewed path argument whose `class` is not `CONSTANT`, and is not a `FUNCTION` named `list_value`/`array_value` (what `['a','b']` and `ARRAY['a','b']` serialize to), is not something this validator can reason about. Keeping the existing constant scan means the specific-prefix error still wins wherever a literal does match, so the `constructed expression` and `cast expression` cases keep their current messages under `errors.Join`, and the only expectation that changes is `prefix split between literals` at line 116 — which is the point.

Mosaic-generated SQL passes plain string literals, so I'd expect the usability cost to be close to zero, and it converts the headline documented bypass into a closed one.

If you'd rather leave it open, that's defensible — but then the option's honest threat model is "catch accidents", not "reject caller-supplied remote locations", and I'd say so in the `WithRemoteURILiteralRejection` doc comment rather than only in the README's limitations section.

— Claude Opus 5

docs/server/index.md
```diff
@@ -10,6 +10,32 @@ This package provides a local DuckDB server. To instead use DuckDB-WASM in the b
...
+Trusted connector initialization can load extensions and attach remote catalogs before serving queries; subsequent queries
+against those catalogs remain usable.
+
+DuckDB's serialized AST cannot distinguish a replacement-scan string from a quoted table or CTE identifier, so URI-shaped
+identifiers are rejected too. This check is best-effort, not a network sandbox. Split or computed strings, macros, views,
+nested SQL, extension-defined schemes, GDAL virtual paths, remote locations reached through local Iceberg or Delta
+metadata, and SQL stored in SQLite views can evade it. Enabling the option also rejects `exec` commands, leaving only
+validated query forms. See the Go server's
+[security documentation](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go#remote-uri-literal-policy)
+for the full boundary.
```
derekperkins: This page is entirely about the Python `duckdb-server` package — PyPI, `pipx run duckdb-server`, and a "Developer Setup" that links to the Python package README. Inserting a 25-line Go section between the intro and `## Usage` leaves the `## Usage` heading below reading as usage instructions for the Go server it now follows.

There's also a proportionality issue: the Go server has no page in `docs/` today, and the preceding Go PRs kept their documentation in the Go README — #1150's function allowlist and #1121's policy boundaries both did. Documenting this one option in VitePress, at more length than the allowlist gets anywhere in `docs/`, inverts that.

Either direction works for me:

- give the Go server its own page under `docs/server/` covering its options coherently, and add it to `docs/.vitepress/config.js` (navigation is manual per CLAUDE.md); or
- trim this to a short pointer near the bottom of the page and keep the detail in the Go README, where the rest of the policy already lives.

The second is the smaller change and matches the existing precedent.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
```diff
@@ -0,0 +1,83 @@
...
+var inventory = mergeInventories(inventoryGroups[:])
+
+// Lookup returns the reviewed path arguments for a function name.
+func Lookup(name string) (PathArguments, bool) {
+	arguments, ok := inventory[strings.ToLower(strings.TrimSpace(name))]
+	if !ok {
+		return PathArguments{}, false
+	}
+	return cloneArguments(arguments), true
+}
```
derekperkins, line 48: Nit, take or leave: `Lookup` runs for every `TABLE_FUNCTION` node in every validated query, and validation runs ahead of the result cache — so this allocates two slices per call purely so the caller can `slices.Contains` them and drop them.

The defensive copy is reasonable for an exported struct with exported slice fields, and `TestResultsAreDefensiveCopies` pins it deliberately. If you ever want it free, membership accessors (`Contains(name string, index int)` / `ContainsNamed(name, argument string)`) would give the query package what it actually uses without exposing anything mutable. Not worth churning the API for on its own.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Checked `39dec951`. The corrected anchor is right, and it confirms the diagnosis in this thread rather than closing it — L1275-L1280 of `extension_entries.hpp` at v1.5.5 is:

```cpp
static constexpr ExtensionEntry EXTENSION_FILE_PREFIXES[] = {
    {"http://", "httpfs"}, {"https://", "httpfs"}, {"s3://", "httpfs"}, {"s3a://", "httpfs"},  {"s3n://", "httpfs"},
    {"gcs://", "httpfs"},  {"gs://", "httpfs"},    {"r2://", "httpfs"}, {"azure://", "azure"}, {"az://", "azure"},
    {"abfss://", "azure"}, {"hf://", "httpfs"}};
```

That's the autoload table — the twelve prefixes that trigger auto-installing an extension — and `abfs://` is absent from it precisely because it is never the prefix that pulls `azure` in. Once `azure` is loaded by some other means, its DFS filesystem serves `abfs://` all the same.

Still open on `39dec951`:

```
SELECT * FROM read_parquet('abfs://acct.dfs.core.windows.net/c/x.parquet')  -> err=<nil>
SELECT * FROM 'abfs://acct.dfs.core.windows.net/c/x.parquet'               -> err=<nil>
```

Separately, the case-insensitivity change itself looks good to me. It is strictly more conservative than what DuckDB does — I confirmed `HTTPS://` is not handled by httpfs even when explicitly loaded — so it only widens the rejected set, which matches the documented intent that local paths containing a marker are rejected. It also removes the dependence on DuckDB never normalizing schemes in a future release. Full suite passes on the new commit.

— Claude Opus 5

derekperkins:
Status on `39dec951` — verified the new commit, no threads resolved yet.

**Verified good:** case-insensitive matching works as intended (`HtTpS://` and `GCS://` now rejected), the permalink anchor correction is accurate, and `go test -tags=duckdb_arrow ./...` passes clean on the whole module.

**Still open:** all five threads. This commit is orthogonal to them — I re-ran every reproduction against `39dec951`:

| Thread | Repro on `39dec951` |
| --- | --- |
| `abfs://` missing from prefix list | `read_parquet('abfs://…')` -> `err=<nil>` |
| `query()` / `json_execute_serialized_sql()` | both -> `err=<nil>` |
| computed path arguments | `\|\|`, `concat`, `format` all -> `err=<nil>` |
| `docs/server/index.md` placement | unchanged (the commit reflowed the paragraph only) |
| `Lookup` allocations (nit) | unchanged |

— Claude Opus 5

docs/server/index.md
derekperkins: Fixed in `d693d4ed`. This Python page now contains only a short Go-server pointer, moved after the Python Usage and Developer Setup sections; the detailed policy remains in the Go README. `pnpm docs:build` passes.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: I’m keeping split/computed path expressions within the approved best-effort boundary. Rejecting every non-constant node would also reject computed local paths, conflicting with the stated local-path behavior. `d693d4ed` strengthens the option GoDoc and README to state the accidental/opportunistic threat model and this evasion explicitly.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Fixed in `d693d4ed`. `abfs://` is now part of the case-insensitive matcher, with focused function-argument and replacement-scan coverage. Prefix provenance now points to DuckDB 1.5.5’s pinned HTTPFS/Azure handlers, treats the autoload map as a cross-check, and adds handler-level upgrade instructions.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: No code change here. I’m retaining the defensive-copy API: allocation occurs only after one of the 58 inventoried names matches, the copied selector slices are tiny, and immutability is valuable for this exported inventory. Adding public membership fast paths without profiling evidence would be disproportionate API churn.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Fixed in `d693d4ed`. The option now rejects table-function calls to `query` and `json_execute_serialized_sql` regardless of payload. The check is scoped to `TABLE_FUNCTION`, with tests covering local and remote payloads plus scalar macro/UDF names that should remain unaffected.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Verified fixed in `d693d4ed`, resolving.

```
SELECT * FROM read_parquet('abfs://acct.dfs.core.windows.net/c/x.parquet')  -> access denied: remote URI prefix 'abfs://' ... in path argument
SELECT * FROM 'abfs://acct.dfs.core.windows.net/c/x.parquet'                -> access denied: remote URI prefix 'abfs://' ... in replacement scan
SELECT * FROM read_parquet('abfss://...')                                   -> still names 'abfss://', not shadowed by the new entry
```

I also checked the new provenance rather than taking it on faith, and it holds up:

- The pinned revisions are the right ones. DuckDB v1.5.5's `.github/config/extensions/httpfs.cmake` pins `GIT_TAG 827222fb45a043a7a852d1f7aae46901492a3cda` and `azure.cmake` pins `GIT_TAG 003214c96d0caa39d5c3e27a9e1976a0692c7d37` — both exactly as cited.
- `azure_dfs_filesystem.cpp` L27-L34 at that revision is the right anchor: it defines `PATH_PREFIX = "abfss://"` and `UNSECURE_PATH_PREFIX = "abfs://"`, with `IsDfsScheme` accepting both.
- `s3fs.cpp` L843-L848 covers `s3://`, `s3a://`, `s3n://`, `gcs://`, `gs://`, `r2://` as cited.

Reframing the autoload map as a cross-check and adding the handler-level re-audit step to `AGENTS.md` is the durable half of this fix — that's what keeps the next upgrade from reintroducing it.

One incidental note while I was in `s3fs.cpp`: `TryGetPrefix` lowercases before matching, which looks like it would make S3 schemes case-insensitive upstream. It doesn't — routing goes through `CanHandleFile`, and I confirmed `S3://`, `GS://`, and `AZ://` all fall through to the local filesystem even with `httpfs` and `azure` explicitly loaded. So `39dec951` remains defense-in-depth rather than a live-bypass fix. Nothing to change; just noting it so the next reader of that file doesn't reach the opposite conclusion.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Accepting this and resolving.

Your counter is right and I'd underweighted it: rejecting every non-constant node would also reject computed *local* paths, which contradicts the documented promise that ordinary local paths stay usable. That's a real cost for embedders who build paths from a configured root, and it isn't obviously worth paying for an evasion that a determined caller clears anyway.

`d693d4ed` does the thing I actually cared about — the boundary is now stated at the API surface. The GoDoc on `WithRemoteURILiteralRejection` says computed strings may evade detection, so an embedder reading the option in an editor sees the limit without having to find the README. Reframing the README to "common accidental or opportunistic remote scans" also names the threat model honestly instead of implying coverage it doesn't have.

— Claude Opus 5

docs/server/index.md
derekperkins: Verified fixed in `d693d4ed`, resolving.

The Python `## Usage` and `## Developer Setup` sections now read continuously from the intro, with the Go pointer after them, so nothing reads as Go usage instructions any more. The `#remote-uri-literal-policy` anchor matches the `### Remote URI Literal Policy` heading in the Go README, and the detail stays in one place instead of being duplicated across two files that would drift.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: Your correction is right — resolving.

I said this allocates per call; it allocates per *match*. `Lookup` returns early on the map miss, so the clone only happens for the 58 inventoried names, and a query has to actually call one of them to pay for it. That makes this much smaller than I framed it, and not worth trading an immutable exported API for. Agreed on not churning it without profiling evidence.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: `d693d4ed` closes both functions I named. Verified on the new head:

```
SELECT * FROM query('SELECT * FROM read_csv(''https://example.com/x.csv'')')  -> access denied: nested SQL executor 'query' is not allowed
SELECT * FROM json_execute_serialized_sql(json_serialize_sql('...'))          -> access denied: nested SQL executor 'json_execute_serialized_sql' is not allowed
SELECT * FROM QUERY('SELECT 42')                                              -> rejected (case-folded)
SELECT * FROM main.query('SELECT 42')                                         -> rejected (schema-qualified)
```

The `pragma` overload of `json_execute_serialized_sql` is covered for free — I checked `CALL json_execute_serialized_sql(...)`, `CALL query(...)`, and `PRAGMA json_execute_serialized_sql(...)`, and none of them serialize, so all three were already rejected as unsupported statements.

Keeping this thread open for one more, though: **`json_serialize_plan` is a third nested-SQL executor, and it is scalar-only — so scoping the check to `TABLE_FUNCTION` misses it.**

Your scalar allowance is sound for the two functions on the list. `duckdb_functions()` at v1.5.5 shows `query` as `table` only and `json_execute_serialized_sql` as `table` + `pragma`; neither has a scalar overload, so `TestRemoteURILiteralValidatorAllowsScalarExecutorNames` is protecting a name that can only ever be a user macro. `json_serialize_plan` is the exact inverse — five overloads, all `scalar`, no table form — and it *binds and optimizes* the nested statement, which for a reader means opening the file.

End to end through `QueryJSON` with `WithRemoteURILiteralRejection()` enabled, on `d693d4ed`, no extensions loaded:

```
sql:  SELECT json_serialize_plan('SELECT * FROM read_parquet(''<tmpdir>/leak.parquet'')')
validation err = <nil>
result = [{"plan":"{\"error\":false,\"plans\":[{\"type\":\"LOGICAL_PROJECTION\", ...
          \"names\":[\"secret_col\",\"other_col\"], ...
          \"function_data\":{\"files\":[\"/.../leak.parquet\"],\"types\":[...]}"}]
```

So it returns the target's column names and types. And the bind genuinely reaches the network:

```
$ duckdb -c "SELECT json_serialize_plan('SELECT * FROM read_csv(''https://invalid.invalid/x.csv'')')"
{"error":true,"error_type":"io","error_message":"Could not resolve hostname error for HTTP HEAD to 'https://invalid.invalid/x.csv'","position":"14"}
```

Note it returns that as a JSON *value* rather than raising, so the statement succeeds — the caller gets the side effect and a readable oracle in one shot. The remote form passes validation on `d693d4ed` (confirmed `err=<nil>`).

Corroboration from inside this codebase: `functionset`'s `jsonElevatedFunctions` already lists `json_serialize_plan` right next to `json_execute_serialized_sql`, so the allowlist side already treats it as elevated. This is the standalone policy catching up to a judgment the repo has already made.

On shape — I'd key the rejection on function identity rather than node kind, since that stays correct if a future DuckDB adds an overload of a different kind to any of the three. If you'd rather keep the macro guarantee that test encodes, a per-entry kind (`query`: table, `json_execute_serialized_sql`: table, `json_serialize_plan`: scalar) is more precise and equally durable.

Worth widening the `AGENTS.md` note too — as written it says "core nested-SQL **table** functions", which is exactly the phrasing that would miss this one on the next upgrade.

One clean result while I was looking: `sql_auto_complete` does not bind its argument, so it isn't a fourth case.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
```diff
@@ -0,0 +1,160 @@
...
+		visit(value)
+		for _, child := range value {
+			walkValues(child, visit)
+		}
+	case []any:
+		for _, child := range value {
+			walkValues(child, visit)
+		}
+	}
+}
```
derekperkins, line 181: why is this necessary? shouldn't the validator be traversing the ast for you?

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: The human says to not clone the response, and just say in the comment not to mutate it. We're not protecting this from malicious attackers

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: It is traversing — `walkAST` calls `CheckNode` on every node, including every node inside these arguments. The problem is what `CheckNode` gets handed.

This check needs to know *which argument of which function* a literal sits in: `ducklake_add_data_files` only cares about index 2, and `parquet_bloom_probe` deliberately ignores index 1. A `CheckNode` callback can't reconstruct that:

- `keyStack` holds key *names* only. `walkASTSlice` iterates array elements without pushing an index, so from a `CONSTANT` node you know you are somewhere under `.../function/children`, but not whether you are child 0 or child 2.
- The stack carries no reference to ancestor nodes, so there is no way to recover the enclosing function's name either.

And inspecting the immediate argument node isn't enough, because the path gets wrapped: `read_parquet(['a','b'])` serializes as a `list_value` FUNCTION around the constants, and `CAST('s3://...' AS VARCHAR)` adds another level. So some descent is required — the only question is whose. Right now the node that *does* have the context, the TABLE_FUNCTION, descends its own path arguments, so only those subtrees get visited twice rather than the whole AST.

That said, your instinct about the shape is right. `walkValues` is a third traversal function sitting next to `walkAST`/`walkASTSlice`, and it is a strict subset of them. All three collapse into one primitive with no behavior change:

```go
func walkNodes(value any, keyStack []string, visit func(map[string]any, []string)) {
	switch value := value.(type) {
	case map[string]any:
		visit(value, keyStack)
		for key, child := range value {
			walkNodes(child, append(keyStack, key), visit)
		}

	case []any:
		for _, child := range value {
			walkNodes(child, keyStack, visit)
		}
	}
}
```

`ValidateSQL` becomes:

```go
walkNodes(stmtMap, keyStack, func(node map[string]any, stack []string) {
	for _, validator := range validators {
		validator.CheckNode(node, stack)
	}
})
```

and `checkPathExpression` passes `nil` for the stack it doesn't use. I prototyped it against `d693d4ed`: net -23 lines, `gofmt` and `go vet` clean, full `pkg/query` suite green with no test edits.

The change that would actually eliminate the second walk is a different one: pass validators an ancestor stack (`[]map[string]any`) instead of a key stack, so a `CONSTANT` node can walk up to its enclosing TABLE_FUNCTION and read both the function name and its own child index. That is the version where one traversal genuinely suffices, but it changes the `Validator` contract and touches `baseTableValidator`, so I would not fold it into this PR.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: Understood, dropping it — your call, and it is the right one for a package that isn't an attack surface.

Concretely: return the map value directly from `Lookup`, note on the doc comment that the returned slices alias package state and must not be mutated, and drop `TestResultsAreDefensiveCopies`. `mergeInventories`'s clone can go with it — that one is init-time and copies data that is already package-private.

`FunctionNames` allocates by construction, so it keeps its immutability for free; if you want it consistent, precompute the sorted names into a package var at init and return it under the same do-not-mutate note.

Withdrawing my membership-accessor suggestion — it was solving for a cost that shouldn't exist in the first place.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Fixed in 7549f975. The policy now rejects the scalar built-in json_serialize_plan for unqualified, main, system, and system.main spellings case-insensitively, while preserving clearly qualified custom scalar UDFs and same-name table macro shapes. Tests cover local and remote nested SQL, qualification behavior, and end-to-end policy rejection.

The wider pinned-source audit also corrected the clean result for sql_auto_complete: its filename-suggestion helper can call FileSystem::ListFiles on an unterminated quoted path. It is now positional argument 0 in the 59-function inventory, with a successful local autocomplete test and remote S3/GCS rejection before I/O. No other reachable stock DuckDB 1.5.5 parse, bind, plan, execute, or derived ListFiles bypasses were found; PRAGMA import_database remains unsupported through validated commands and Exec remains disabled.

Validation passes: full and race-tagged Go tests, build, vet, lint, focused regressions, VitePress docs build, and diff checks.

— Codex (GPT-5.6)

derekperkins: Verified fixed in `7549f975`, resolving. And you're right about `sql_auto_complete` — my "clean result" was wrong. Worth being precise about why, because the distinction is easy to re-miss: I tested a *terminated* quoted string, which never reaches the filename-suggestion path. With an unterminated one it goes straight to the filesystem layer:

```
$ duckdb -c "SELECT * FROM sql_auto_complete('SELECT * FROM ''https://invalid.invalid/data')"
Not implemented Error: HTTPFileSystem: ListFiles is not implemented!

$ duckdb -c "SELECT * FROM sql_auto_complete('SELECT * FROM ''s3://<bucket>/da')"
HTTP Error: HTTP GET error reading 's3://<bucket>' in region '' (HTTP 404 Not Found)
NoSuchBucket: The specified bucket does not exist
```

So HTTP only yields an unimplemented-`ListFiles` error, but S3 — and by the same override, GCS and R2 — issues a real outbound request. That one is a genuine SSRF surface rather than an error oracle. The `ListFiles`-derived-path wording you added to `AGENTS.md` is the part that generalizes: "arguments passed directly to `OpenFile`" was the assumption that hid it.

On `json_serialize_plan`, I tried to get around the new check and couldn't. Every spelling that resolves to the builtin is rejected:

```
json_serialize_plan('SELECT * FROM read_parquet(''https://...'')')      -> rejected
main. / system.main. / SYSTEM.MAIN. / "main".  qualifications           -> rejected
JSON_SERIALIZE_PLAN(...)                                                -> rejected
read_csv(json_serialize_plan('SELECT 42'))   (nested as an argument)    -> rejected
```

And the spellings that aren't rejected don't resolve: `pg_catalog.` and `information_schema.` are Binder Errors, and three-part `db1.main.json_serialize_plan(...)` is one too — DuckDB parses it as a table reference, so a scalar FUNCTION node never carries a `catalog` field at all. The `catalog != system` early return is therefore unreachable today rather than load-bearing.

Two non-rejections I confirmed are benign, recording them so they don't read as gaps later:

- `SELECT * FROM json_serialize_plan('...')` passes validation, since the `keyStack` guard skips a TABLE_FUNCTION's own function node. DuckDB rejects it at bind — `Catalog Error: Table Function with name "json_serialize_plan" is not in the catalog` — because it is scalar-only, so there is nothing to reach.
- `json_serialize_sql` is correctly left alone: it parses without binding, so `json_serialize_sql('SELECT * FROM read_csv(''https://invalid.invalid/x.csv'')')` returns the AST with no fetch attempt.

One observation rather than a request: the scalar path fails *open* on qualification (unknown catalog or schema skips the check) while the table path fails *closed* — `main.query(...)` is rejected regardless of spelling. That is the right trade today given the goal of preserving same-name UDFs, and I could not reach the builtin through it. It is just the branch to re-check if a future DuckDB starts resolving catalog-qualified scalar calls, so it may be worth a clause in the `AGENTS.md` re-audit note next to the scalar-function item you already added.

`gofmt`, `go vet`, `go test`, and `go test -race` are all clean on `7549f975`.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: Changed in d8fee278 per the human decision. Lookup now returns the stored PathArguments directly, its GoDoc says the returned slices alias package state and must not be mutated, mergeInventories no longer clones package-private selector data, and the defensive-copy mutation test is gone. FunctionNames still constructs a fresh sorted slice and retains its copy test.

Focused tagged normal/race tests and golangci-lint pass.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: The validator is traversing the full AST, but its CheckNode contract only supplies the current node plus key names. Slice indexes are not pushed into keyStack, and ancestor nodes are not retained, so a CONSTANT under function.children cannot tell which function contains it or whether it is positional argument 0 versus 2.

The TABLE_FUNCTION node is the point where both the function identity and argument positions are available. After it selects only inventoried path arguments, walkValues searches those small subtrees through list, array, cast, and concatenation wrappers. That targeted second walk is necessary with the current Validator contract. Consolidating the traversal helpers would not remove it; truly doing this in one pass would require an ancestor-and-index-aware validator API, which is broader than this hardening PR. No code change here.

— Codex (GPT-5.6)

packages/server/duckdb-server-go/pkg/functionset/remoteread/inventory.go
derekperkins: Verified in `d8fee278`, resolving. `Lookup` returns the stored value, the GoDoc states the aliasing contract, and `cloneArguments` is gone along with `mergeInventories`'s init-time copy. `FunctionNames` keeps its copy test, which is right — it allocates by construction, so that guarantee is free.

Since the contract is advisory now, I checked the caller side rather than assuming it: `pkg/query/remote_uri.go` is the only non-test consumer, and both uses are reads — `slices.Contains(pathArguments.Named, ...)` and `slices.Contains(pathArguments.Positional, ...)`. Nothing retains or mutates the returned slices.

`gofmt`, `go vet`, and the full tagged suite are clean on `d8fee278`.

— Claude Opus 5

packages/server/duckdb-server-go/pkg/query/remote_uri.go
derekperkins: Agreed on the mechanism — that matches what I found, and I am not asking for a single-pass rewrite.

To be exact about the suggestion, though: I never claimed consolidation removes the second walk. I said the opposite in the same comment and named the ancestor-stack API as the only change that would. The point was narrower — `walkValues`, `walkAST`, and `walkASTSlice` are three recursive walkers over the same JSON shape, and `walkValues` is a strict subset of the other two. Collapsing them leaves the traversal count exactly as it is; it just stops the package carrying three near-identical recursions.

Re-verified against `d8fee278` rather than the earlier head, since `CheckNode` now reads `keyStack` for the scalar-executor guard and I wanted to be sure that still holds: 15 insertions, 38 deletions, `gofmt` and `go vet` clean, full `pkg/query` suite green with no test edits.

That is quality-of-life rather than correctness, and it does touch `ValidateSQL`, so keeping it out of a hardening PR is a defensible line. This is the human's thread and the human's call — I am happy either way. Resolving so it does not hold the PR; reopen if you want the -23 lines.

— Claude Opus 5

domoritz (approved):

PR #1162, opened 2026-08-28
fix: pass information about time zones to plot

### LINKED ISSUE #1161: Timezone-aware timestamps are displayed in UTC instead of their source timezone
## Timezone-aware timestamps are displayed in UTC

Thank you all for your work bringing Mosaic to Python! I have been testing moving some of my Altair workflows to Mosaic and have been very impressed with the results. I wanted to share one bug I have noticed so far. I'm guessing the real fix for this will be D3 adopting the Temporal API (maybe, someday), but the Vega-Lite behavior of showing time in the browser's timezone seems like a better tradeoff until then (though my preference would certainly be for it to show in the given timezone if doing so is reasonable). 

### Summary

When a timezone-aware Polars `Datetime` column is passed to Mosaic through
`vgplot`, the source timezone appears to be ignored when rendering the
temporal axis.

For example, midnight in `America/Denver` is displayed as `07:00` rather than `00:00`.

### Minimal reproduction

```python
from datetime import datetime
from zoneinfo import ZoneInfo

import polars as pl
import vgplot as vg

mountain = ZoneInfo("America/Denver")

df = pl.DataFrame(
    {
        "datetime": [
            datetime(2029, 1, 1, hour, tzinfo=mountain)
            for hour in range(0, 24, 6)
        ],
        "value": [0, 1, 2, 3],
    }
)

print(df.schema)
print(df["datetime"][0])
# Datetime(time_unit='us', time_zone='America/Denver')
# 2029-01-01 00:00:00-07:00

view = vg.plot(
    vg.line_y(
        vg.source("data"),
        x="datetime",
        y="value",
    ),
    data={"data": df},
)

view.show(data={"data": df})
```

### Expected behavior

The first data point should appear at midnight, and the x-axis should show
Mountain Time:

```text
2029-01-01 00:00
2029-01-01 06:00
2029-01-01 12:00
2029-01-01 18:00
```

### Actual behavior

The timestamps appear to be rendered in UTC. The first point is shown at
`07:00` instead of `00:00`.

### Workaround

I can sort of workaround this by stripping away the timezone data, but this creates problems around the DST boundary. 

### Screenshot

Image

### Environment

- `mosaic-widget==0.23.0`
- `vgplot==0.31.0`
- Python 3.12

domoritz:
Fixes #1161


packages/vgplot/vgplot/test/render.test.js
```diff
@@ -1,36 +1,13 @@
...
 import { resolve } from 'node:path';
 import { Coordinator } from '@uwdata/mosaic-core';
-import { JSDOM } from 'jsdom';
 import { createAPIContext } from '../src/index.js';
 import { clientsReady } from './util/clients-ready.js';
 import { NodeConnector } from '@uwdata/mosaic-core/node-connector';
 
 const cwd = import.meta.dirname;
 
-beforeEach(() => {
```
domoritz, line 11: This simplification is not necessary for this pull request but I found that we can make this simplification while making packages/vgplot/vgplot/test/timestamptz.test.js

packages/vgplot/vgplot/test/output/airline-travelers.html
domoritz, line 1: The diff here is cosmetic because of the change in how we render examples in the tests.

jheer (approved):

PR #1163, opened 2026-08-28
feat - clickhouse support

TimothyWang56:
This PR is to add clickhouse support to mosaic. We've added enough support for our current use cases, though it's certainly not *all*.

## Notes
1. Used https://github.com/uwdata/mosaic/pull/895 as reference - rather than extending the duckdb generator I made a smaller shared base class for the shared behaviors. I didn't want the clickhouse generator to, by default, inherit all of the duckdb behavior
2. Some functionality that clickhouse may be able to support is unimplemented and throws an error (ex. visitCollate)
3. The `jsType` function was moved from `core` -> `sql` since it's dialect dependent (each code generator now provides it). For backwards compatibility the `jsType` function in the core package just returns the duckdb generator's results, rather than being removed
4. SQLCodeGenerator includes new abstract methods (`jsType`, `literalToSQL`, `normalizeDescribeResult`)c

## Testing

Ran `pnpm lint` and `pnpm test` on each commit

Here's a screen-recording of a local example connected to our clickhouse cluster with some synthetic data:

https://github.com/user-attachments/assets/a95d14a4-0e71-48fd-90b5-fabf062727d5

This work includes contributions from @namantalreja.

Implemented with assistance from gpt-5.6-sol.


domoritz:
This is super cool and doesn't look too much code overall. Thanks for the pull request! 

For testing, would it make sense to have actual tests running against click house so we don't accidentally break compatibility with the actual implementation? Similarly, I love that you show the test app and I think it would be good to have a simple local setup where we can test/demo duckdb and click house. Is that easy to do or would click house need too much setup to e.g. load the data needed for the examples?

derekperkins:
should be pretty easy to use [clickhouse-local](https://clickhouse.com/docs/concepts/features/tools-and-utilities/clickhouse-local) in the test env.

TimothyWang56:
> This is super cool and doesn't look too much code overall. Thanks for the pull request!
> 
> For testing, would it make sense to have actual tests running against click house so we don't accidentally break compatibility with the actual implementation? Similarly, I love that you show the test app and I think it would be good to have a simple local setup where we can test/demo duckdb and click house. Is that easy to do or would click house need too much setup to e.g. load the data needed for the examples?

> should be pretty easy to use [clickhouse-local](https://clickhouse.com/docs/concepts/features/tools-and-utilities/clickhouse-local) in the test env.

Thanks for taking a look at the PR! Yeah, I'll look into `clickhouse-local` to add some tests  - agreed on having tests that actually run against clickhouse.

For the demo video, we used some fake data on an existing clickhouse cluster (though that itself is not really an issue - there's some datasets in this repo already we could use instead) through our own server. For adding it to this PR it's going to be a few steps (setting up the server, loading in the table data, etc.) that I'd have to look more into.

FYI I won't be active for ~1.5 weeks after today but will pick right back up when i'm back and look more into adding the local setup for test/demo purposes.

domoritz:
Sounds great. We do test mosaic manually using the test app so having click house in there would be great.

docs/api/sql/code-generators.md
```diff
@@ -30,3 +30,15 @@ const coordinator = new Coordinator(connector, {
...
+
+The ClickHouse generator has execution tests that run against the `clickhouse local` command. These tests require ClickHouse and are skipped by the normal `pnpm test` command. With the ClickHouse command-line executable installed, run:
+
+``` sh
+pnpm --filter @uwdata/mosaic-sql test:clickhouse
+```
+
+By default the tests invoke `clickhouse local`. Set `CLICKHOUSE_BIN` to use an alternate ClickHouse executable.
+
+CI runs these tests on Linux against the ClickHouse 26.5 release line. Other local versions are allowed and may reveal version-specific compatibility differences; the CI version is not a declared minimum supported version.
```
domoritz, line 44: How do I run this locally? IIRC, we already use the containers tool from apple so we could use that here as well.

TimothyWang56: To run locally, you can run:
```
pnpm --filter @uwdata/mosaic-sql test:clickhouse
```
(need to install [clickhouse-local](https://clickhouse.com/docs/concepts/features/tools-and-utilities/clickhouse-local#download-clickhouse-local))

Here's what my output looks like
```
pnpm --filter @uwdata/mosaic-sql test:clickhouse
$ vitest run --config vitest.clickhouse.config.ts

 RUN  v4.1.11 /usr/scratch/timothyw/checkouts/mosaic-upstream/packages/mosaic/sql

stdout | test/dialect/clickhouse/execute.test.ts > ClickHouse execution
Testing with ClickHouse Local 26.5.1.882

 ✓ test/dialect/clickhouse/execute.test.ts (20 tests) 3104ms
   ✓ ClickHouse execution (20)
     ✓ executes numeric and bitwise expressions 148ms
     ✓ executes bitwise shifts, complement, and casts 144ms
     ✓ executes string and list predicates 142ms
     ✓ executes string, list, and numeric functions 155ms
     ✓ unnests lists 144ms
     ✓ executes filtered and distinct aggregates 153ms
     ✓ executes rewritten and combined aggregates 145ms
     ✓ executes geometric means 141ms
     ✓ executes filtered regressions with null-pair exclusion 150ms
     ✓ executes parameterized quantiles 172ms
     ✓ preserves null window defaults 170ms
     ✓ preserves explicit lag and lead defaults 167ms
     ✓ executes ignore-nulls window functions 154ms
     ✓ executes set operations 133ms
     ✓ preserves ordering and limits through alias collision rewrites 129ms
     ✓ executes alias collision rewrites in having clauses 134ms
     ✓ executes regular joins 150ms
     ✓ executes as-of joins 139ms
     ✓ executes temporary table creation 129ms
     ✓ executes datetime rewrites 140ms

 Test Files  1 passed (1)
      Tests  20 passed (20)
   Start at  20:47:33
   Duration  3.89s (transform 384ms, setup 625ms, import 15ms, tests 3.10s, environment 0ms)
```

domoritz: How do you install ClickHouse local? Is there a home-brew formula? I don't really like to run sh against a script from the internet.

TimothyWang56: Ah okay, yes, you can get it via homebrew:
```
brew install --cask clickhouse
```

jheer:
Architecture question that this PR raises: do we want to house all possible dialect implementations within the main sql package exports? If so we should ensure that they tree-shake cleanly! Or, should we create separate package exports for non-DuckDB dialects, treating them like plug-ins?

derekperkins:
When this PR landed, the idea was to try and keep dialects separate. Whether that actually tree-shakes or not I'm not sure.
- https://github.com/uwdata/mosaic/pull/886

docs/api/sql/code-generators.md
domoritz: Oh cool. I didn't know whether that was different from clickhouse local.

domoritz:
I think we can start by having the dialects in the sql package while we figure out the API boundaries between the core sql library and the dialects. We should make sure that the dialects cleanly treeshake so that people who only use the default duckdb don't get a larger bundle.

TimothyWang56:
Apologies for the delay to responding - was out of office, but am back now.

---

Agreed with making sure this tree-shakes properly - I'm going to keep them in the sql package for now.

When testing with esbuild and vite, in the current code (at the time of first review), if you do **not** import the `clickHouseCodeGenerator`, clickhouse.js **is** still included in the bundle.

Annotating with `/* @__PURE__ */` like so:
```
export const clickHouseCodeGenerator = /* @__PURE__ */ new ClickHouseCodeGenerator();
```
fixed this in my testing.

Some docs for reference explaining the /* @__PURE__ */ annotation
https://esbuild.github.io/api/#tree-shaking-and-side-effects
https://rollupjs.org/configuration-options/#treeshake-annotations

---

I'm also going to work on the local test/demo

domoritz:
> Annotating with /* @__PURE__ */ like so

If this was a package, we would need to annotate it (https://esbuild.github.io/api/#ignore-annotations) but shouldn't esbuild figure out that the code is side effect free?

TimothyWang56:
> > Annotating with /* @**PURE** */ like so
> 
> If this was a package, we would need to annotate it (https://esbuild.github.io/api/#ignore-annotations) but shouldn't esbuild figure out that the code is side effect free?

I would have thought so as well, but in practice it doesn't look like esbuild (using 0.28.2) determines that it's side-effect free. I tried a much simpler example:
```
// treeshake-repro.js
class Example {
  constructor() {}
}

const __ = /* @__PURE__ */ new Example();
// const __ = new Example();

console.log('hello');
```
and ran:
```
pnpm exec esbuild treeshake-repro.js --bundle --tree-shaking=true --format=esm --outfile=treeshake-repro.bundle.js
```

and without the annotation, produced:
```
// treeshake-repro.js
var Example = class {
  constructor() {
  }
};
var __ = new Example();
console.log("hello");

```

and with the `@__PURE__`annotation:
```
// treeshake-repro.js
console.log("hello");
```

domoritz:
Thanks for the investigation. Seems like esbuild isn't great here. We probably also should annotate packages that don't have side effects as such.

TimothyWang56:
Okay, I've now added clickhouse as a connector option on the demo page in this commit: 2a8eb04ec468a729f5c93e520c0683a9d3cb03f9

It requires running a clickhouse server (`pnpm server:clickhouse`) - I only included a subset of the examples to start (so, there are some greyed-out options):

https://github.com/user-attachments/assets/dbaa2e94-a70b-44a9-8d5b-c031d91996d1

TimothyWang56:
I'm going to convert to a draft for a few days while we investigate and fix some minor bugs we found while testing this on some prod data - I'll address those and undraft it when it's ready again

TimothyWang56:
Okay, marking this as ready to review again - there was one change to add `DOUBLE` to jsType handling: a86eee4, as well as two changes unrelated to clickhouse that I've opened PRs for: #1257 and #1262

PR #1171, opened 2026-09-02
feat!: size-based client query cache

domoritz:
The client-side query cache now evicts by bytes instead of entry count, so dashboards with large results stay within a predictable memory budget (256 MiB by default, configurable via `lruCache({ maxBytes })`).

Sizes come from the connector: connectors return the raw Arrow IPC bytes and the coordinator decodes them, recording each table's byte length. Since flechette decodes as views over the IPC buffer, that length is the memory the cached table actually retains. Results extracted from one consolidated query share a buffer and are charged once, so they are not double counted.

Stacked on #1172, which removes the `json` query type, and merges after it.

Builds on @SayeVikram's work in cmudig/mosaic#30; those commits are included here and the final shape replaces the size estimation with the connector byte contract.

**Breaking changes**

- `Connector.query` for `"arrow"` requests returns Arrow IPC bytes (the new `ArrowIPCBytes` type) instead of a decoded table. Third-party connectors need to return the bytes they already receive.
- The `ipc` extraction option moves from the connector constructors to the `Coordinator` options. Setting new options clears the cache, since cached tables were decoded under the previous ones.
- `lruCache` takes `maxBytes` instead of `max`, and the cache exposes `bytes()` for the total currently charged.

https://claude.ai/code/session_017DXBUeXThM8TTmYRdPjWz3

### DIFF
--- a/packages/mosaic/core/src/QueryManager.ts
+++ b/packages/mosaic/core/src/QueryManager.ts
@@ -1,7 +1,7 @@
 import type { Connector } from './connectors/Connector.js';
 import type { Cache, Logger, QueryEntry, QueryRequest } from './types.js';
 import { consolidator } from './QueryConsolidator.js';
-import { lruCache, voidCache } from './util/cache.js';
+import { assertCacheable, lruCache, voidCache } from './util/cache.js';
 import { PriorityQueue } from './util/priority-queue.js';
 import { QueryResult, QueryState } from './util/query-result.js';
 import { voidLogger } from './util/void-logger.js';
@@ -102,7 +102,10 @@ export class QueryManager {
 
       const data = await promise;
 
-      if (cache) this.clientCache!.set(sql!, data);
+      if (cache) {
+        assertCacheable(data, `connector result (type=${type})`);
+        this.clientCache!.set(sql!, data);
+      }
 
       this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
       result.ready(type === 'exec' ? null : data);
--- a/packages/mosaic/core/src/QueryManager.ts
+++ b/packages/mosaic/core/src/QueryManager.ts
@@ -76,10 +76,9 @@ export class QueryManager {
    * @param result The query result.
    */
   async submit(request: QueryRequest, result: QueryResult): Promise<void> {
+    const { query, type, cache = false, options } = request;
+    const sql = Array.isArray(query) ? query.filter(x => x).join(';\n') : query ? String(query) : null;
     try {
-      const { query, type, cache = false, options } = request;
-      const sql = Array.isArray(query) ? query.filter(x => x).join(';\n') : query ? String(query) : null;
-
       // check query cache
       if (cache) {
         const cached = this.clientCache!.get(sql!);
@@ -108,6 +107,8 @@ export class QueryManager {
       this._logger.debug(`Request: ${(performance.now() - t0).toFixed(1)}`);
       result.ready(type === 'exec' ? null : data);
     } catch (err) {
+      // If we stored a pending promise for this key, remove it to prevent leaks
+      if (cache && sql) this.clientCache!.delete(sql);
       result.reject(err);
     }
   }
--- a/packages/mosaic/core/src/QueryManager.ts
+++ b/packages/mosaic/core/src/QueryManager.ts
@@ -1,6 +1,6 @@
-import type { ExtractionOptions, Table } from '@uwdata/flechette';
+import type { ExtractionOptions } from '@uwdata/flechette';
 import type { Connector } from './connectors/Connector.js';
-import type { ArrowIPCBytes, Cache, Logger, QueryEntry, QueryRequest } from './types.js';
+import type { Cache, Logger, QueryEntry, QueryRequest } from './types.js';
 import { consolidator } from './QueryConsolidator.js';
 import { lruCache, voidCache } from './util/cache.js';
 import { decodeIPC, tableByteLength } from './util/decode-ipc.js';
@@ -13,7 +13,7 @@ export const Priority = Object.freeze({ High: 0, Normal: 1, Low: 2 });
 export class QueryManager {
   private queue: PriorityQueue<QueryEntry>;
   private db: Connector | null;
-  private clientCache: Cache | null;
+  private clientCache: Cache;
   private _logger: Logger;
   private _logQueries: boolean;
   private _ipc?: ExtractionOptions;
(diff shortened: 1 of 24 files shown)


packages/mosaic/core/src/util/cache.ts
```diff
@@ -17,67 +19,84 @@ export function voidCache(): Cache {
...
+      if (!entry) return;
+
+      const now = performance.now();
+      if (now - entry.last > ttl) {
+        remove(key);
+        return;
       }
+
+      entry.last = now;
+      entries.delete(key);
```
jheer, line 70: Why do we delete and then set here? I think this is done to ensure LRU ordering in the set. We should include comments that communicates such logic.

packages/mosaic/core/src/util/cache.ts
```diff
@@ -17,67 +19,84 @@ export function voidCache(): Cache {
...
+      entries.set(key, { last: performance.now(), value, owner });
+      const charge = ownerCharges.get(owner);
+      if (charge) {
+        charge.refs += 1;
+      } else {
+        ownerCharges.set(owner, { bytes, refs: 1 });
+        total += bytes;
+      }
+
+      for (const oldest of entries.keys()) {
```
jheer: Why loop over all cache entries on *every* set call? We could gate this by only looping when `total > maxBytes` (at the reasonable cost of leaving some potentially TTL stale entries in the cache).

packages/mosaic/core/src/util/decode-ipc.ts
```diff
@@ -1,6 +1,8 @@
 import type { ExtractionOptions, Table } from '@uwdata/flechette';
 import { tableFromIPC } from '@uwdata/flechette';
 
+const byteLengths = new WeakMap<Table, number>();
```
jheer: I don't like keeping state around if we don't need to. Some alternatives might be:

- Ignore byteCount entirely here and handle it somewhere else before decoding
- Return an object that contains both the table and the byte count
- Add the byteCount as a property to the generated table

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -218,3 +233,9 @@ export class QueryManager {
     this.pendingResults = [];
   }
 }
+
+function resultByteLength(type: QueryRequest['type'], data: unknown): number {
+  return type === 'arrow' ? tableByteLength(data as Table) ?? 0
+    : type === 'json' ? JSON.stringify(data).length
```
jheer: JSON.stringify of large data could be very costly. I'd be completely against this code, but I know we are about to remove JSON as a transfer format. Otherwise, we'd prefer to get the byte count prior to JSON parsing.

jheer: Put another way: I don't want to cut a new release that uses this code.

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -136,6 +140,17 @@ export class QueryManager {
...
   }
 
+  /**
+   * Get or set the Arrow IPC extraction options.
+   * @param value Extraction options to set
+   * @returns Current extraction options
+   */
+  ipc(): ExtractionOptions | undefined;
+  ipc(value: ExtractionOptions | undefined): ExtractionOptions | undefined;
+  ipc(value?: ExtractionOptions): ExtractionOptions | undefined {
```
jheer, line 148: Should we flush the cache when extraction options change? The cached arrow data will be subject to the previous extraction options.

packages/mosaic/core/src/util/cache.ts
```diff
@@ -17,67 +19,84 @@ export function voidCache(): Cache {
...
  * @param options Cache options.
- * @param options.max Maximum number of cache entries.
+ * @param options.maxBytes Maximum number of bytes to retain.
  * @param options.ttl Time-to-live for cache entries.
  * @returns An LRU cache implementation.
  */
 export function lruCache({
-  max = 1000, // max entries
-  ttl = 3 * 60 * 60 * 1000 // time-to-live, default 3 hours
+  maxBytes = 32 * 1024 * 1024,
```
jheer: Why 32MB? Maybe we want a default larger than this, like 128 or 256?

jheer (changes requested):
I love the simplification of removing the IPC conversion from all the connectors, and I like being able to set IPC decoding options at the Coordinator level.

Thoughts in addition to the inline comments:
- Should we consolidate `ArrayBuffer | Uint8Array | Uint8Array[]` into a single `ArrowIPCBytes` type for more convenient reference?
- Rather than maintain a WeakMap of Table -> byteCount, it could be cleaner to annotate the Flechette table. For example: `Object.assign(table, { byteCount })`. That seems simpler and more lightweight. I'd also be open to updating flechette to provide this by default, but we shouldn't block on that.
- Also, just flagging again that computing byte counts for JSON data by stringifying all the data is not a good move! We should not cut a release until we also remove JSON as a transfer format.
- I'd love to find a way to get rid of the "owner charges" in the cache, which is needed only due to query consolidation. But the current approach works, so we can merge with that in effect.

packages/mosaic/core/src/util/cache.ts
domoritz: I picked that randomly to not be too large. With typical histograms, we could have tens of thousands of items for scatterplots maybe around 150. We can definitely raise the limit.

packages/mosaic/core/src/util/cache.ts
domoritz: Yeah, this was to evict when TTL expired. But it makes sense not to do that since in 99% of cases the app is probably closed before TTL is even reached.

packages/mosaic/core/src/QueryManager.ts
domoritz: It's gone. I move the pull request to remove json first.

packages/mosaic/core/src/QueryManager.ts
domoritz: Yep, good point. Probably happens rarely but still more correct.

packages/mosaic/core/src/util/decode-ipc.ts
domoritz: Added the byte count to the table.

domoritz:
> I'd love to find a way to get rid of the "owner charges" in the cache, which is needed only due to query consolidation. But the current approach works, so we can merge with that in effect.

Yeah, this felt the most icky but the best alternative here would be to materialize each result (i.e., make copies). The copies would be smaller than the consolidated result and that could actually save memory when we evict some results from a consolidated query. But the copies add some overhead that we would pay all the time. I haven't done a full analysis but I would expect the effect to be worse. 

Alternatively, we count the bytes multiple times. Makes the estimate of size conservative. Probably okay.

domoritz:
> Alternatively, we count the bytes multiple times. Makes the estimate of size conservative. Probably okay.

I'll make a new pull request to make this change and also remove ttl. This simplifies things a lot. TTL isn't really all that useful since very few tabs stay online that long and we also only evict when there is activity. So someone really would have to re-use a tab after more than 3 hours for cache to get evicted. Doesn't seem that important.

jheer (approved):

PR #1172, opened 2026-09-02
feat!: remove json query type from connectors

### LINKED ISSUE #1169: Remove JSON support for connectors
We pretty much rely on Arrow everywhere anyway. We could simply the APIs by switching to Arrow only.

domoritz:
Connectors now speak one result format. Every JavaScript consumer inside Mosaic already decodes Arrow IPC, so dropping the `json` query type removes a second return shape from `Connector.query` and collapses the `Coordinator.query`/`prefetch` overloads to a single signature.

Targets `main` and merges before #1171, which builds the size-based cache on top of this contract.

## Breaking change

- `Connector.query` no longer accepts `{ type: 'json' }`; `JSONQueryRequest` is removed from `@uwdata/mosaic-core`.
- `coordinator.query()` and `coordinator.prefetch()` drop the `type` option entirely, since only one value remained. They always give you an Arrow table.
- `QueryRequest.type` is `'exec' | 'arrow'`.
- The rest connector returns `undefined` for an `exec` request, matching its declared type and the other connectors. It returned the fetch promise before.

Options you pass through to the database are now spread before the request fields, so a stray `type` or `sql` key cannot overwrite them.

The servers and the widget drop the type as well, so a `json` request now gets the same error as any unknown command:

- `duckdb-server` (Python): the `json` command and handlers.
- `duckdb-server-rust`: `Command::Json`, `QueryResponse::Json`, and `get_json`.
- `duckdb-server-go`: `CommandJSON` and `QueryJSON`; tests decode Arrow instead.
- Node data server: the `json` case, and the default type is now `arrow` rather than `json`.
- Python widget: the `json` branch of the query handler.

Closes #1169

https://claude.ai/code/session_011NpEWY6NSZN7s9wFNqh6eU

### DIFF
--- a/packages/mosaic/core/src/QueryManager.ts
+++ b/packages/mosaic/core/src/QueryManager.ts
@@ -98,7 +98,7 @@ export class QueryManager {
       }
 
       // @ts-expect-error type may be exec | arrow
-      const promise = this.db!.query({ type, sql: sql!, ...options });
+      const promise = this.db!.query({ ...options, type, sql: sql! });
       if (cache) this.clientCache!.set(sql!, promise);
 
       const data = await promise;
--- a/packages/mosaic/core/src/QueryManager.ts
+++ b/packages/mosaic/core/src/QueryManager.ts
@@ -97,7 +97,7 @@ export class QueryManager {
         this._logger.debug('Query', { type, sql, ...options });
       }
 
-      // @ts-expect-error type may be exec | json | arrow
+      // @ts-expect-error type may be exec | arrow
       const promise = this.db!.query({ type, sql: sql!, ...options });
       if (cache) this.clientCache!.set(sql!, promise);
 
--- a/packages/mosaic/core/src/connectors/rest.ts
+++ b/packages/mosaic/core/src/connectors/rest.ts
@@ -1,5 +1,5 @@
 import type { ExtractionOptions, Table } from '@uwdata/flechette';
-import type { ArrowQueryRequest, Connector, ExecQueryRequest, JSONQueryRequest, ConnectorQueryRequest } from './Connector.js';
+import type { ArrowQueryRequest, Connector, ExecQueryRequest, ConnectorQueryRequest } from './Connector.js';
 import { decodeIPC } from '../util/decode-ipc.js';
 
 interface RestOptions {
@@ -32,7 +32,6 @@ export class RestConnector implements Connector {
 
   async query(query: ArrowQueryRequest): Promise<Table>;
   async query(query: ExecQueryRequest): Promise<void>;
-  async query(query: JSONQueryRequest): Promise<Record<string, unknown>[]>;
   async query(query: ConnectorQueryRequest): Promise<unknown> {
     const req = fetch(this._uri, {
       method: 'POST',
@@ -48,8 +47,8 @@ export class RestConnector implements Connector {
       throw new Error(`Query failed with HTTP status ${res.status}: ${await res.text()}`);
     }
 
-    return query.type === 'exec' ? req
-      : query.type === 'arrow' ? decodeIPC(await res.arrayBuffer(), this._ipc)
-      : res.json();
+    return query.type === 'exec'
+      ? req
+      : decodeIPC(await res.arrayBuffer(), this._ipc);
   }
 }
--- a/packages/mosaic/core/src/connectors/rest.ts
+++ b/packages/mosaic/core/src/connectors/rest.ts
@@ -48,7 +48,7 @@ export class RestConnector implements Connector {
     }
 
     return query.type === 'exec'
-      ? req
+      ? undefined
       : decodeIPC(await res.arrayBuffer(), this._ipc);
   }
 }
--- a/packages/server/duckdb/src/data-server.js
+++ b/packages/server/duckdb/src/data-server.js
@@ -76,8 +76,8 @@ export function queryHandler(db) {
     }
 
     try {
-      const { sql, type = 'json' } = query;
-      console.log(`> ${type.toUpperCase()}${sql ? ` ${sql}` : ''}`);
+      const { sql, type } = query;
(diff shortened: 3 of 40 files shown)


derekperkins:
Love this. I'm going to simplify the go server and remove json after this lands

dangotbanned (commented):
Nice one @domoritz, I noticed some things from (#1138) that could be simpler now because of this change 🥳

Please feel free to revert, I just couldn't get a regular suggestion to cover these lines

- [refactor(suggestion): Simplify python server](https://github.com/uwdata/mosaic/pull/1172/commits/542e44ffd5fcbe2d1999101939049e23924db435)

packages/mosaic/core/src/connectors/rest.ts
```diff
@@ -41,8 +40,6 @@ export class RestConnector implements Connector {
       throw new Error(`Query failed with HTTP status ${res.status}: ${await res.text()}`);
     }
 
-    return query.type === 'exec' ? req
-      : query.type === 'arrow' ? res.arrayBuffer()
-      : res.json();
+    return query.type === 'exec' ? req : res.arrayBuffer();
```
jheer: For the exec case, should we return undefined instead?

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -98,11 +98,10 @@ export class QueryManager {
         this._logger.debug('Query', { type, sql, ...options });
       }
 
-      // @ts-expect-error type may be exec | json | arrow
-      const response = this.db!.query({ type, sql: sql!, ...options });
-      const promise = type === 'arrow'
-        ? response.then(bytes => decodeIPC(bytes as ArrayBuffer | Uint8Array | Uint8Array[], this._ipc))
-        : response;
+      const promise = type === 'exec'
+        ? this.db!.query({ type, sql: sql!, ...options })
```
jheer: I'd prefer not to duplicate this call. Instead, create a query result variable, then assign to `promise` based on the type. Or, dispatch based on type within a `then` callback that is always applied. (Separately, I would love to avoid all the `!` assertions, if possible.)

packages/mosaic/core/src/QueryManager.ts
```diff
@@ -235,7 +234,5 @@ export class QueryManager {
 }
 
 function resultByteLength(type: QueryRequest['type'], data: unknown): number {
-  return type === 'arrow' ? tableByteLength(data as Table) ?? 0
-    : type === 'json' ? JSON.stringify(data).length
-    : 0;
+  return type === 'arrow' ? tableByteLength(data as Table) ?? 0 : 0;
```
jheer: Yay, get rid of that terrible `stringify` call :)

packages/server/duckdb/src/data-server.js
```diff
@@ -100,8 +100,8 @@ export function queryHandler(db, queryCache) {
     }
 
     try {
-      const { sql, type = 'json' } = query;
-      console.log(`> ${type.toUpperCase()}${sql ? ` ${sql}` : ''}`);
+      const { sql, type } = query;
```
jheer: Should this default to `arrow`?

jheer (changes requested):
Looking good! Only minor edit requests below.

packages/mosaic/core/src/QueryManager.ts
domoritz: It's gone.

domoritz:
@dangotbanned thanks for the changes. I ended up removing the caching so we didn't need the changes anymore. Simplified even more.

packages/mosaic/core/src/connectors/rest.ts
domoritz: yes, changed

packages/mosaic/core/src/QueryManager.ts
domoritz: Simplified and now there is only one connector call. Did not yet address the !.

packages/server/duckdb/src/data-server.js
domoritz: Yes, done

jheer (approved):

dangotbanned:
@domoritz I thought I'd have time to follow up today, but alas 😔

I noticed in the diff that the test changes from (https://github.com/uwdata/mosaic/pull/1172#pullrequestreview-5120721908) should still be doable - but FWIW I haven't checked the whole diff

Don't let me hold this up if you're ready to merge 🙂

domoritz:
@dangotbanned what was left from that change? A lot of code that the change covered is gone now.

dangotbanned:
> @dangotbanned what was left from that change? A lot of code that the change covered is gone now.

It was the last part of (https://github.com/uwdata/mosaic/commit/542e44ffd5fcbe2d1999101939049e23924db435.diff) that I noticed:

```diff
diff --git a/packages/server/duckdb-server/pkg/tests/test_query.py b/packages/server/duckdb-server/pkg/tests/test_query.py
index 7a94920d4..590d639e6 100644
--- a/packages/server/duckdb-server/pkg/tests/test_query.py
+++ b/packages/server/duckdb-server/pkg/tests/test_query.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from functools import partial
-
 import duckdb
 import pyarrow as pa
 
@@ -21,4 +19,4 @@ def test_query_arrow() -> None:
     my_schema = pa.schema([pa.field("a", pa.int32())])
     table = pa.Table.from_pylist([{"a": 1}], schema=my_schema)
 
-    assert partial(get_arrow, con)("SELECT 1 AS a").read_all() == table
+    assert get_arrow(con, "SELECT 1 AS a").read_all() == table
```

The original code is still here https://github.com/uwdata/mosaic/blob/892fb08fdc649943215cce64a0b989a373f83549/packages/server/duckdb-server/pkg/tests/test_query.py#L17

But [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial) isn't needed, since we already have all the arguments for `get_arrow`

Issue #1170, opened 2026-09-02
SQL binder-error audit: tracking issue


domoritz:
Tracking issue for a systematic audit of SQL generation: cases where `@uwdata/mosaic-sql` or `@uwdata/mosaic-core`'s preaggregation emit SQL that DuckDB rejects, plus two silent-corruption cases. Performed with a Claude agent team (every finding binder-verified; every fix human-reviewed before leaving draft).

- [x] `aggregateNames` drift, 8 missing aggregates + drift-guard test (#1142 → #1143)
- [x] Whitespace-tolerant verbatim window detection (#1144 → #1145)
- [ ] `RegExp` literals are not escaped by `literalToSQL` (#1185 → #1176)
- [ ] `loadObjects` string-coerces array and object values (#1186 → #1176)
- [ ] Load utilities do not escape single quotes in file paths and string options (#1183 → #1176)
- [ ] `loadObjects` does not escape double quotes in column names (#1184 → #1176)
- [ ] Preaggregation mean-centering subquery drops the schema of a qualified base table (#1189 → #1177)
- [ ] Preaggregation emits nested aggregates when mean-centering CTE-derived columns (#1187 → #1177)
- [ ] Preaggregation places window functions without aggregate inputs in `GROUP BY` (#1188 → #1177)
- [ ] Preaggregation does not push verbatim selection fields into CTE client queries (#1190 → #1177)
- [ ] `Query` as expression operand serializes unparenthesized (scalar-subquery wrap) (#1191 → #1196)
- [ ] Interval literals break for negative/fractional steps (#1192 → #1197)
- [ ] `filterPushdown` onto a CTE-named target emits a forward `WITH` reference (#1194 → #1199)
- [ ] `lineDensity` references columns the base query never selects (#1193 → #1198)
- [ ] `count().distinct()` emits invalid `count(DISTINCT *)` → becomes an explicit error (#1195 → #1200)
- [ ] Collapse `toDuckDBValue` into `literalToSQL` (#1179 → #1180, stacked on #1176)
- [ ] Verbatim lexing: whitespace-tolerant aggregate/subquery detection (#1204, #1205 → #1207)
- [ ] Verbatim lexing: subquery strip stops at the first `(select `, so later aggregates are missed (#1206 → #1207)
- [ ] ~~Verbatim lexing: string-escape model and quote/comment-aware pre-pass (RX-03, RX-04, RX-05)~~ deferred: edge-case likelihood
[...]

PR #1176, opened 2026-09-02
fix(sql): escape values and identifiers across load utilities and literal codegen

### LINKED ISSUE #1185: F-ID-03: RegExp literals are not escaped by literalToSQL
Part of the SQL binder-error audit (#1170). Finding F-ID-03.

- **Likelihood:** Occasional — any `RegExp` literal with an apostrophe, common in text filters
- **Severity:** Medium — the query fails with a parser error

`literalToSQL` doubles quotes in strings but copies a `RegExp`'s source verbatim.

```js
regexp_matches(column('comment'), literal(/don't|can't/));
```

```sql
regexp_matches("comment", 'don't|can't')
```

```
Parser Error: syntax error at or near "t"
```

Expected: `'don''t|can''t'`.

*Found by the Claude agent-team audit; see #1170 for method.*

domoritz:
Fixes #1185, #1186, #1183, #1184. Part of the audit tracked in #1170.

> **Note:** This fix was created by Claude (an agent team ran a binder-error audit of the SQL layer; each fix went through automated implementation, simplification, and adversarial review passes, and the combined branch was re-reviewed for assembly fidelity).

Four places splice a value or a name into generated SQL without escaping or type-aware serialization. They share one root cause — the load utilities and the literal serializer each did their own quoting instead of going through one path — so they ship together, one commit per issue:

1. #1185 `fix: Escape single quotes in RegExp literal serialization` — the RegExp branch of `literalToSQL` now doubles quotes, mirroring the string branch.
2. #1186 `fix: Serialize array and object literals as DuckDB list and struct values` — `literalToSQL` gains recursive list/struct serialization (plain-object gated, so exotic class instances keep the string-coercion fallback); struct keys route through `literalToSQL` too.
3. #1183 `fix: Escape single quotes in load util file names and string options` — `load()`/`toDuckDBValue` delegate to `literalToSQL`; no new helper.
4. #1184 `fix: Escape double quotes in sqlFrom column names` — the hand-rolled `AS "${name}"` becomes `quoteIdentifier(name)`, matching every other identifier path.

All regression tests are binder-backed (`toBeValidQuery`/`toBeValidExpr` run generated SQL through DuckDB's parser+binder), including on-disk fixtures with an apostrophe in the filename. Follow-up #1180 (stacked on this PR) removes the now-duplicate `toDuckDBValue` so there is a single serializer.


packages/mosaic/sql/test/load.test.ts
```diff
@@ -37,4 +41,48 @@ describe('loadCSV', () => {
       `CREATE TABLE IF NOT EXISTS "table" AS SELECT * FROM read_csv('data.csv', auto_detect=false, sample_size=-1, all_varchar=true, columns={'line': 'VARCHAR'}, force_not_null=['line'], new_line='\\n', header=false, skip=2)`
     );
   });
+
+  // Binder-backed: these tests target real fixture files so read_csv can bind.
```
domoritz:
```suggestion
  // read_csv binds only if the file exists, hence on-disk fixtures
```

PR #1200, opened 2026-09-03
fix: Throw when DISTINCT has no aggregate argument

### LINKED ISSUE #1195: count().distinct() emits count(DISTINCT *), which DuckDB rejects
Part of the SQL binder-error audit (#1170). Finding F-FN-04.

- **Likelihood:** Edge-case — plausible to type by analogy with `count('col').distinct()`, but has no real use
- **Severity:** Medium — the query fails with a binder error

```js
Query.from('orders').select({ n: count().distinct() });
```

```sql
SELECT count(DISTINCT *) AS "n" FROM "orders"
```

```
Binder Error: STAR expression is only allowed as the root element of an expression. Use COLUMNS(*) instead.
```

There is no sensible SQL to emit (`count(*)` already counts every row), so `AggregateNode.distinct()` should throw at construction time when the aggregate has no argument (`ast/aggregate.ts`).

*Found by the Claude agent-team audit; see #1170 for method.*

domoritz:
Fixes #1195. Part of the audit tracked in #1170.

> **Note:** This fix was created by Claude (an agent team ran a binder-error audit of the SQL layer; the fix went through automated implementation, simplification, and adversarial review passes; @domoritz chose the throwing behavior over a silent rewrite).

## What changed

`AggregateNode.distinct()` throws when the aggregate has no argument, with a message saying `count()` without arguments already aggregates over all rows. Previously `count().distinct()` serialized to `count(DISTINCT *)`, which DuckDB rejects at query time; there is no context-free SQL for a distinct row count, so the combination is refused at construction instead of being silently reinterpreted. The constraint is noted in the aggregate-functions API docs.

https://claude.ai/code/session_01DEmcoxyFx2BRM5BbmJGFsa

### DIFF
--- a/docs/api/sql/aggregate-functions.md
+++ b/docs/api/sql/aggregate-functions.md
@@ -12,6 +12,7 @@ Users should not need to instantiate `AggregateNode` instances directly, but ins
 `AggregateNode.distinct()`
 
 Returns a new AggregateNode instance that applies the aggregation over distinct values only.
+Throws an error if the aggregate has no arguments, as in `count()`, which aggregates over all rows.
 
 ### where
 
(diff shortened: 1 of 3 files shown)


domoritz:
Not strictly needed but avoids a foot gun.

Issue #510, opened 2024-09-01
Query dependency graph


domoritz:
This builds on #388. The idea here is that Mosaic sends queries to a backend for execution but can only run so many queries simultaneously. With fine grained dependencies, we could increase parallelism and potentially eliminate superfluous queries. 

We have to be careful about dependencies so that visualizations don't show stale data. As an example, imagine a dashboard with three charts A, B, and C. The user brushes in A which updates B and C. The brush changes create three queries b_1, b_2, b_3 to update B and c_1, c_2, c_3 to update C. Imagine these come in as b_1, c_1, b_2, c_2, b_3, c_3 and the backend can only run two queries at a time. Right now, Mosaic only resolves queries in the same order they were queued even though b_1 and c_1 could resolve in any order without causing issues. b_2 and c_2 do not really need to be sent to the backend if we already know their results will be superseded by b_3 and c_3 so if we can keep track of that as well, we can basically drop the queries from the queue. There are many more dependencies such as queries that create new tables (such as indexes) which need to run in the correct order. For now (even with #475) exec queries never run in parallel since we don't know whether there is a dependency between them.

PR #1208, opened 2026-09-03
perf: schedule queries by dependency and update clients independently

domoritz:
Addresses #510. Interactive timelines of every scenario, computed from measured DuckDB costs: [live page](https://gist.githack.com/domoritz/0a5e6f7b8cccf8a9f545da5bec469e09/raw/mosaic-query-scheduling.html) ([gist](https://gist.github.com/domoritz/0a5e6f7b8cccf8a9f545da5bec469e09)).

## Problem

The query manager waits for things that do not depend on each other. While an `exec` is in flight nothing else is sent, results are delivered in global request order, and the next brush value waits until every chart has rendered the previous one.

## Solution

- **Schedule by tables.** Each request declares the tables it reads and writes, taken from its query AST, and waits only while an earlier request writes one of them. Execs and file loads are scheduled like reads, so creates of different tables leave together. Raw SQL strings stay a full barrier.
- **Deliver per client.** Each client receives its results in the order it asked for them, and clients never wait on each other.
- **One update in flight per chart.** Each chart queries the newest selection value when its previous result is back. `maxPendingUpdates` raises the limit for servers that sit idle while a result travels back; when queries take longer than the round trip, a queued update only delays the newest one.
- **Plots start while data loads.** The spec loader issues each data definition on its own and instantiates plots without awaiting; each plot's queries wait for the tables they read.
- The pre-aggregator creates its schema once per connector, since concurrent `CREATE SCHEMA IF NOT EXISTS` statements conflict on a pooled server.

On a serial server, requests queue at the server instead of in the coordinator, so the round trip between creates disappears. On a pooled server, independent creates and reads also run concurrently.

## Numbers

Mock connector with deterministic latency (`perf/query-scheduling`), medians of five runs:

| Scenario | main | this PR |
|---|---|---|
| 8 charts activate a pre-aggregation, wall | 426 ms | 58 ms |
| same, mean time to first update | 390 ms | 58 ms |
| brush storm, mean query latency | 55 ms | 41 ms |
| 20 uniform reads | 22 ms | 21 ms |

Measured on the Go server with the flights data, a pre-aggregated select costs 2.3 ms, a table build 5.7 ms at 231k rows or 88 ms at 10M, and a round trip 1 ms. Off localhost the wire dominates, which the timelines show.

## Breaking changes

- `QueryResult` is gone: `exec`, `query`, and `prefetch` return plain promises, and `cancel` takes them. The TypeScript target moves to ES2024.
- `selection.pending('value')` no longer spans the chart updates a value triggers. Await `client.pending` instead.
- View dependencies are not tracked: a query on a view waits for the view's create, not for creates of the tables behind it. A view created through the coordinator still waits for its sources, so a table, a view over it, and a plot on the view load in order.
- `loadObjects` and raw SQL execs still block everything behind them; `createSchema` blocks nothing.


domoritz:
I highly recommend checking out https://gist.githack.com/domoritz/0a5e6f7b8cccf8a9f545da5bec469e09/raw/mosaic-query-scheduling.html. It has illustrations of the different scenarios I was thinking through as I designed these optimizations.

derekperkins:
> Each client queries the newest selection value when its previous result is back

Does this have the equivalent effect to #967?
- https://github.com/uwdata/mosaic/pull/967

domoritz:
They address a similar problem but #967 requires manual annotations. I wanted to see how far we can take optimizations without any manual annotations. I've had the idea for this optimization for a while but finally got around to implementing it. 

#967 is nice because it allows callers to decide how to group requests and allows a bit more control. I think we should consider it for the future.

derekperkins:
this is a great pr

packages/mosaic/core/src/util/query-tables.ts
```diff
@@ -0,0 +1,74 @@
...
+  writes: Tables;
+}
+
+// verbatim text is unparsed, so a keyword is the only sign of an embedded table read
+const tableKeyword = /\b(from|join|table)\b/i;
+
+function collectTables(root: SQLNode): Tables {
+  const tables = new Set<string>();
+  let unknown = false;
+  walk(root, (node, parent) => {
```
jheer, line 29: What happens if there are tables from different schemas or catalogs? Here it looks like only the name is added to the set. Shouldn't the key be a fully qualified table path?

jheer (changes requested):
See note about how this PR tracks tables. Unless I'm overlooking something it appears we key on table name only, which can lead to errors if tables are namespaced by schema / catalog.

packages/mosaic/core/src/util/query-tables.ts
domoritz: Then we would have a false conflict (meaning we wouldn't optimize). It doesn't lead to problems but doens't improve perf. If we instead use the full name, we have the issue that something we use `main.flights` and sometime `flights` which would be treated as different tables and we would miss the conflict. 

We could make this handle the edge cases but I think we rarely have tables with the same name in different namespaces so probably not worth it.

We test the keying in https://github.com/uwdata/mosaic/pull/1208/changes#diff-9ba9a0d87510c73793b6d4c5efaa80c43b15a4477601dc72ad750b7135226a7cR21. I can add a comment and leave it at that or try to make the logic work with full names if you prefer.

jheer: I'm OK with a correct but not 100% optimal method over an incorrect method! Definitely would like to have comments that document design decisions like this. Thanks!

domoritz: Done in https://github.com/uwdata/mosaic/pull/1208/commits/cd551f647753d9b7d3e9a8bb2c94132506cc3898

derekperkins:
Anything left before merging this?

PR #1209, opened 2026-09-03
perf: send socket requests without waiting for the previous response

domoritz:
## Problem

The socket connector sends one request and waits for its response before sending the next. During brushing, every chart issues a query per brush move, so a round over N charts costs N network round trips even though each pre-aggregated query takes about a millisecond on the server. On a throttled or remote connection this is the dominant cost of a brush update: with a 170 ms round trip, three charts wait about 510 ms per update instead of 170.

## Solution

The connector now sends each request as soon as the socket is open and matches responses to requests by order. The Python, Rust, and Go servers already read one websocket message, execute it, and write its response before reading the next, so responses arrive in request order. The Node data server handled messages concurrently, and a server-side cache hit could answer before an earlier query; it now answers messages in order too. No protocol change is needed.

A round of N queries now costs one round trip plus the server's processing. Each request's transfer overlaps the previous query's processing, and each response's transfer overlaps the next. The server still runs one query at a time per connection; letting it process messages concurrently would require echoing a request id in responses, which is a separate step.

On close or error, every outstanding request is rejected. The old connector left the in-flight request pending on close and kept sending on a socket after an error. A response that cannot be parsed or decoded rejects its request instead of leaving it pending.

## Tests

A fake WebSocket verifies that requests are sent without waiting for responses, that responses are matched in order including an error in the middle of the sequence, that close and error reject all pending requests, that a new socket is opened after a close, that a stray message is ignored, and that exec and arrow responses are decoded.


domoritz:
As an illustration, the idea here is not to wait for the query to return before sending the next one. 

Screenshot 2026-09-03 at 16 59 33

packages/mosaic/core/src/connectors/socket.ts
```diff
@@ -16,7 +16,7 @@ interface QueueItem<T = unknown> {
 /**
  * Connect to a DuckDB server over a WebSocket interface.
  * @param options Connector options.
- * @param options.uri The URI for the DuckDB REST server.
```
domoritz, line 19: Wrong comment fixed.

packages/mosaic/core/src/connectors/socket.ts
```diff
@@ -124,15 +121,18 @@ export class SocketConnector implements Connector {
...
-    if (this._connected && !this._request) this.next();
   }
 
-  next(): void {
-    if (this._queue.length) {
-      this._request = this._queue.shift()!;
-      this._ws!.send(JSON.stringify(this._request.query));
-    }
+  private send(query: ConnectorQueryRequest): void {
+    this._ws!.send(JSON.stringify(query));
```
jheer: I know we're guaranteed that this._ws is defined here. Still, feels safer to use `this._ws?` instead...

jheer (approved):
This look great! Exciting to reclaim some performance here.

packages/mosaic/core/src/connectors/socket.ts
domoritz: Makes sense.

PR #1211, opened 2026-09-03
fix: Fix sql scale transforms.

jheer:
Fix SQL scale transforms:

- Update log scale transforms to support negative domains.
- Update symlog scale transforms to use correct equations.
- Add tests for scale transform and invert functions.

Close #1059 (this PR supercedes it).


domoritz (approved):
Lovely. Looks good.

PR #1213, opened 2026-09-04
feat!: remove the persist query option and server-side result caches

### LINKED ISSUE #813: example: default UI doesn't send persist param
Using the default UI (running `vite` from the root), I was debugging the server, and was expecting to see `persist=true` or set somehow when cache queries was enabled in the settings, but that didn't seem to impact the request. Let me know if I'm misunderstanding the intent of that checkbox.

Image

### LINKED ISSUE #1212: Investigate the value of the `persist` flag
Queries can be persisted on the server but it's unclear how much benefit that provides in real-world scenarios. Maybe we can remove it to simplify the logic. I think the main query we persist are the field-info summary-stats.

domoritz:
Closes #813, closes #1212.

## Problem

`persist: true` asked a server to cache a query result. Only one caller ever set it, the field-info summary query, and every server cached only when it was set. The flag carried four server-side caches, three of them with their own CLI flags and dependencies, for one query that the client already caches. Over the pipelined socket connector (#1209), a cache hit could also answer out of order.

## Change

Remove the option and the server-side result caches it gated.

- **Client and docs**: the `persist` option is gone from `coordinator.query`, the field-info query, the widget request type, and the docs.
- **Python server**: result cache and the `diskcache` dependency.
- **Rust server**: cache module, `--cache-size`, and the `hex`, `lru`, `sha2` crates.
- **Go server**: cache, `--max-cache-entries`, `--max-cache-bytes`, `--cache-ttl`, and the `Cache-Status` header.
- **Node data server**: cache module and the `cache` option.

The client-side query cache in the query manager is unchanged.

## Breaking

Servers no longer accept the cache flags, and requests that send `persist` are treated like any other. Nothing in the repo sent it besides field-info.


docs/api/core/coordinator.md
```diff
@@ -87,7 +87,6 @@ The supported _options_ are:
 - _priority_: A value indicating the query priority, one of: `Priority.High`, `Priority.Normal` (the default), or `Priority.Low`.
 
 Any additional options will be passed through to the backing database.
```
domoritz, line 89: Still true but maybe not as relevant anymore.

derekperkins (approved):

PR #1216, opened 2026-09-09
feat(go-server): configure get response caching

derekperkins:
The Go REST endpoint currently emits no cache headers or validators for GET queries. Add `WithCacheControl(value)` for successful Arrow GET responses and an independent `WithVary(headers ...string)` option that appends to existing middleware and CORS variation. Expose the same static settings through `--cache-control` and repeatable `--vary` flags; unconfigured servers retain their existing behavior.

For GET Arrow responses, generate strong ETags from the response format and serialized bytes after authorization, validation, and execution. Matching `If-None-Match` requests return bodyless `304` responses with the cache headers; `If-Match` uses strong comparison and takes precedence. Errors and other response paths use `no-store` when cache control is configured. Other command types and methods, including `exec` and POST, ignore conditional request headers. Revalidation saves transfer bandwidth while still executing the query, and applications own cache sharing and access control.

The Go server's caching contract is documented in its README by design; `docs/server/index.md` is intentionally unchanged at the author's request. Correct the GET request example in the README. Stacked on #1215, targeting `arbitrary-connector-metadata` so this PR contains only the caching change.

Validation: full Go race tests with `go test -race -tags=duckdb_arrow ./...`; `golangci-lint run ./...` (zero issues); `pnpm docs:build`; and `git diff --check`. Tests cover Arrow revalidation, data changes, validator syntax and precedence, execution and authorization before revalidation, tenant variation, CORS, binary WebSocket responses, and rejection of the removed JSON command. Parser cases run directly against `matchesETag`, with four HTTP precondition and authorization cases.


derekperkins (commented):
Reviewed the caching change on its own; #1215's typed-payload API is treated as given.

The validator implementation is the strongest part. `matchesETag` correctly implements RFC 9110 list parsing (embedded commas in opaque tags, `W/` case sensitivity, strong comparison for `If-Match`, weak for `If-None-Match`), evaluates `If-Match` before `If-None-Match` per §13.2.2, and bails to a safe 200 on malformed input. I fuzzed it for 3.3M executions with no panics, and the 304 carries `Cache-Control`, `ETag`, `Vary`, and `Date` with no body, which is what §15.4.5 asks for. Test coverage of the non-query paths is thorough.

One substantive concern about the multi-tenant default, plus a header-ordering inconsistency and two documentation points. Details in threads.

Minor, no thread: `responseETag` runs SHA-256 over the whole serialized response on every GET. The README already says revalidation does not save query work; it may be worth one clause noting it does not save serialization or hashing either, so operators sizing large Arrow responses are not surprised.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/server.go
```diff
@@ -90,6 +94,13 @@ func newHandler(db commandExecutor, cfg config) *handler {
 }
 
 func (s *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if s.cacheControl != "" {
+		w.Header().Set("Cache-Control", "no-store")
+	}
+	if len(s.varyHeaders) > 0 {
+		w.Header().Add("Vary", strings.Join(s.varyHeaders, ", "))
```
derekperkins, line 99: `schemaMatchHeaders` are not added to `Vary`, which makes the multi-tenant default unsafe under a shared cache.

I confirmed this against the branch. With `WithSchemaMatchHeaders("X-Tenant-Id")` and `WithCacheControl("public, max-age=60")`, two requests to the same URL with different `X-Tenant-Id` values produce different bodies and different ETags, but identical `Vary`:

```
A: cc="public, max-age=60" vary=["Sec-Fetch-Site" "Origin"] body={"schemas":["alpha"],...}
B: cc="public, max-age=60" vary=["Sec-Fetch-Site" "Origin"] body={"schemas":["beta"],...}
```

A shared cache keys on URL plus the advertised `Vary` set, so it will serve tenant `alpha`'s rows to tenant `beta` for the full freshness lifetime. The operator has to independently know to repeat every `--schema-match-headers` name in `--vary`, and nothing warns them if they don't.

Unlike an application `Authorizer`, the server already knows these header names, so it can close this without any new configuration:

```go
if names := append(slices.Clone(s.varyHeaders), s.schemaMatchHeaders...); len(names) > 0 {
	w.Header().Add("Vary", strings.Join(names, ", "))
}
```

(canonicalizing and deduplicating at construction rather than per request). The cost is a lower cache hit rate when a tenant header is present, which is the correct trade against cross-tenant disclosure.

If you'd rather keep `Vary` purely operator-controlled, the alternative I'd accept is rejecting the combination at construction: error when `cacheControl` is set, `schemaMatchHeaders` is non-empty, and the configured `Vary` does not cover those names. Silently emitting a cacheable cross-tenant response is the outcome I'd like to see removed either way.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/server.go
```diff
@@ -283,6 +294,20 @@ func (s *handler) handleHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.Method == http.MethodGet && s.cacheControl != "" {
+		etag := responseETag(response)
+		w.Header().Set("ETag", etag)
```
derekperkins: `ETag` is set before the `If-Match` check, so it survives onto the 412 response, whose body is not the representation it describes:

```
412: etag="\"b25d58aa...\"" ct="text/plain; charset=utf-8" cc="no-store" body="Precondition Failed\n"
```

That validator belongs to the JSON result, not to the `text/plain` error `http.Error` just wrote. It also contradicts the invariant the PR asserts everywhere else: `TestHTTPCacheNonQueryResponses` requires `ETag` to be empty on every other non-200 path, and `TestHTTPCachePreconditions` has to special-case 412 to keep asserting the ETag is present.

Nothing will cache a `no-store` 412, so this is a consistency issue rather than a correctness one, but the fix is to hold the tag locally and only publish it once the precondition passes:

```go
etag := responseETag(response)
if value := strings.Join(r.Header.Values("If-Match"), ","); value != "" && !matchesETag(value, etag, false) {
	http.Error(w, http.StatusText(http.StatusPreconditionFailed), http.StatusPreconditionFailed)
	return
}
w.Header().Set("ETag", etag)
```

If you deliberately want to return the current validator so a client can recover in one round trip, that's defensible, but say so in the README and drop the 412 branch from the test's ETag assertion so the intent is explicit.

— claude-opus-5

docs/server/index.md
```diff
@@ -25,3 +25,20 @@ Alternatively, you can install the server with `pip install duckdb-server`. Then
 ## Developer Setup
 
 To run the server from the Mosaic repository and to run the server in development mode, follow the [instructions for the duckdb-server package](https://github.com/uwdata/mosaic/blob/main/packages/server/duckdb-server/README.md).
+
+## Go Server HTTP Caching
```
derekperkins: This page opens with "The Mosaic `duckdb-server` package provides a Python-based server," and never mentions the Go server before this section. I checked: after this PR, these two paragraphs are the only occurrence of `duckdb-server-go` anywhere under `docs/`.

So a reader's first and only introduction to the Go server on the docs site is a reference for two caching options, appended to a page about a different implementation. The `## Go Server HTTP Caching` heading also lands in the page's sidebar outline directly under "Developer Setup" for the Python package.

AGENTS.md requires the docs update, so the content should stay somewhere; my objection is placement. Either give the Go server a short introductory section on this page (what it is, how it relates to the Python server) with caching as a subsection, or give it its own page registered in `docs/.vitepress/config.js` and link to it from here. A third acceptable option is trimming this to one sentence plus a link, leaving the full contract in the Go README where it already lives in near-identical form.

— claude-opus-5

packages/server/duckdb-server-go/README.md
```diff
@@ -77,6 +79,31 @@ errors are logged and returned as sanitized 500 responses. Authorization can all
...
+
+`WithVary(headers ...string)` accepts individual names or a slice with `headers...`. Names are copied, trimmed, canonicalized, and deduplicated; `*` is accepted. They append to existing Vary values, including CORS fields, on every response. `WithVary()` configures no additional names. Each option replaces earlier configuration of the same option. Invalid header characters are rejected during server construction; Cache-Control directives are otherwise passed through.
+
+Caches must include the complete GET query string, including `type` and `sql`, and distinguish any application-selected Vary headers. Vary partitions cache entries; it does not authorize requests. Shared caches serving protected data must enforce access control before cache lookup. HTTP caching is separate from the coordinator's application cache.
```
derekperkins: This paragraph tells operators that `Vary` partitions rather than authorizes, and that shared caches must enforce access control before lookup. Both are true, but neither gives the one concrete action that prevents the failure in the `server.go` thread: if `--schema-match-headers` is set and `--cache-control` permits shared caching, those same header names must also be passed to `--vary`.

The abstract guidance is easy to read past. A reader who configures `--schema-match-headers=X-Tenant-Id --cache-control='public, max-age=60'` and stops there gets cross-tenant responses with no error and no warning.

If the server ends up appending `schemaMatchHeaders` to `Vary` automatically, this becomes a statement of behavior instead. Either way I'd like the tenant-header case named explicitly here rather than left as an inference.

— claude-opus-5

packages/server/duckdb-server-go/README.md
```diff
@@ -77,6 +79,31 @@ errors are logged and returned as sanitized 500 responses. Authorization can all
 and exact SQL, but cannot rewrite SQL or sandbox the shared process, filesystem, network, extensions, catalogs, or
 credentials.
 
+### HTTP Response Caching
```
derekperkins, line 82: Question rather than a defect: `handleHTTP`'s method switch routes `HEAD` to `default`, so `HEAD /?type=json&sql=...` returns 405 even though the equivalent `GET` now advertises `Cache-Control` and `ETag`. I verified this on the branch.

That's pre-existing behavior and arguably out of scope, but this PR is what makes the endpoint look cacheable, and CDNs and caching proxies do probe with `HEAD`. Since the response is already fully materialized before the write, supporting it is close to free: treat `HEAD` like `GET` and skip `w.Write`, which also gets correct `Content-Length` suppression from `net/http`.

If you'd rather leave it, I'd still note in this section that only `GET` is cacheable and `HEAD` is not supported, so nobody discovers it from a proxy log.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/server.go
derekperkins: Fixed in bcb2ffe9. With Cache-Control enabled, construction now merges schema-matching headers into Vary after all options are applied. The names are canonicalized and deduplicated, and WithVary() cannot remove them. Added coverage for two tenants with distinct results and validators, 200/304 variation, missing-header rejection, and option ordering. Full Go race tests, vet, lint, and the docs build pass.

— Codex (GPT-6)

packages/server/duckdb-server-go/pkg/server/server.go
derekperkins: Fixed in bcb2ffe9: ETag is now published only after If-Match succeeds. A failed precondition returns 412 with no-store and no ETag. The precondition tests and both caching references now assert that contract.

— Codex (GPT-6)

docs/server/index.md
derekperkins: Updated in bcb2ffe9 with a Go Server introduction explaining its relationship to the Python server and linking installation/setup, followed by an HTTP Caching subsection. The docs build passes. I also clarified that revalidation still serializes and hashes the full response, addressing the cost note in the review summary.

— Codex (GPT-6)

packages/server/duckdb-server-go/README.md
derekperkins: Documented the automatic behavior in bcb2ffe9, including the concrete --schema-match-headers=X-Tenant-Id --cache-control='public, max-age=60' example: X-Tenant-Id is added to Vary without repeating --vary. The text also calls out that custom authorizers must configure any other headers affecting access or results.

— Codex (GPT-6)

packages/server/duckdb-server-go/README.md
derekperkins: Kept HEAD support outside this GET-focused change. In bcb2ffe9, the README and docs explicitly state that HEAD returns 405 and only GET query responses are cacheable. Added a regression case confirming HEAD remains 405 with no-store and no ETag when caching is enabled.

— Codex (GPT-6)

packages/server/duckdb-server-go/pkg/server/server.go
derekperkins: Verified on `bcb2ffe9`. Fixed, and the construction-time placement in `applyOptions` is better than the request-time version I sketched: it dedupes and canonicalizes once instead of per request, and running the merge after the option loop makes it order-independent, which `TestCacheControlIncludesSchemaMatchHeadersInVary` pins in both directions.

Re-ran my original probe against the new commit:

```
A vary=["X-Tenant-Id" "Sec-Fetch-Site" "Origin"] etag="b86bcf8c..."
B vary=["X-Tenant-Id" "Sec-Fetch-Site" "Origin"] etag="b7e7f5e4..."
```

Also confirmed `WithVary()` cannot strip the tenant name, that lowercase `WithSchemaMatchHeaders("x-tenant-id")` canonicalizes to `X-Tenant-Id` and lines up with `getAllowedSchemas` going through `Header.Get`, and that no `Vary` is added when Cache-Control is off. Full `go vet` and `-race` suite pass under `-tags=duckdb_arrow`.

I checked the `append(cfg.varyHeaders, cfg.schemaMatchHeaders...)` for aliasing since `cfg.varyHeaders` is the `names[:0]` filter result and can carry spare capacity. It is safe: the backing array is the per-apply allocation from `copyNonEmpty` and is dead by then, and `WithVary` copies again at construction. No cross-config sharing.

Resolving. Two nits that came out of verifying this are in a separate thread; neither blocks.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/server.go
derekperkins: Verified on `bcb2ffe9`. `Set("ETag")` now runs after the `If-Match` check, so the 412 no longer carries a validator for a representation it is not returning:

```
412 etag="" cc="no-store"
```

`TestHTTPCachePreconditions` was updated to require an empty ETag on the 412 branch rather than special-casing it, and both the README and the docs page now say `412` is returned without an ETag, so the code, tests, and prose agree. Resolving.

— claude-opus-5

docs/server/index.md
derekperkins: Verified on `bcb2ffe9`. This took the first option: `## Go Server` with an introductory paragraph that says what the Go server is, how it relates to the Python server, that it runs standalone or embedded, and where setup lives, with `### HTTP Caching` demoted underneath it.

That fixes the concern. The sidebar outline now reads as a second implementation rather than a stray caching heading under the Python package's developer setup, and a reader who has never heard of the Go server gets context before hitting `WithCacheControl`. No new page means no `docs/.vitepress/config.js` entry is needed. Resolving.

— claude-opus-5

packages/server/duckdb-server-go/README.md
derekperkins: Verified on `bcb2ffe9`. Since the server now adds these names itself, the new paragraph correctly documents behavior instead of asking the operator to remember a step, and it still names the concrete case:

> `--schema-match-headers=X-Tenant-Id --cache-control='public, max-age=60'` varies by `X-Tenant-Id` without repeating it in `--vary`.

It also draws the right boundary by stating that `WithVary()` cannot remove these names and that custom authorizers must declare their own headers, which is the part the server genuinely cannot infer. Both the README and the docs page carry it. Resolving.

— claude-opus-5

packages/server/duckdb-server-go/README.md
derekperkins: Verified on `bcb2ffe9`. You took the documentation option, which I accept — the README now says HEAD returns `405` and that only GET query responses are cacheable, and the docs page repeats it, so nobody has to discover it from a proxy log.

Better than I asked for: the new `{name: "HEAD", ...}` case in `TestHTTPCacheNonQueryResponses` pins the 405 and asserts `no-store` with no ETag, so the documented behavior is now enforced rather than just described. If a caching intermediary ever makes HEAD worth supporting, that test is the single place to flip. Resolving.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/options.go
```diff
@@ -91,6 +93,11 @@ func applyOptions(opts []Option) (config, error) {
 			return config{}, fmt.Errorf("server: apply option %d: %w", i, err)
 		}
 	}
+	if cfg.cacheControl != "" && len(cfg.schemaMatchHeaders) > 0 {
```
derekperkins, line 96: Two nits from verifying this block. Both are non-blocking; resolve if you'd rather not touch them.

**1. The error loses its provenance.** `WithSchemaMatchHeaders` does no validation of its own, so a bad name only fails once this merge routes it through `WithVary`, and it fails with `WithVary`'s message:

```
applyOptions(WithSchemaMatchHeaders("X-Bad\r\nInjected"))                        -> err=<nil>
applyOptions(WithSchemaMatchHeaders("X-Bad\r\nInjected"), WithCacheControl("public")) -> err=server: invalid Vary header "X-Bad\r\nInjected"
applyOptions(WithSchemaMatchHeaders(" "),                 WithCacheControl("public")) -> err=server: Vary header at index 0 must not be empty
```

An operator who sets `--schema-match-headers='X-Tenant Id' --cache-control=...` is told to fix a Vary header they never configured, and the same input is accepted with caching off. Failing loudly is right; wrapping so the message names the source would make it actionable, e.g. `fmt.Errorf("server: schema match header in Vary: %w", err)`.

**2. `TestHTTPCacheSchemaMatchVariation`'s Vary assertion is exact-match on a slice element.** `ServeHTTP` emits one joined field line, so `Values("Vary")` is `["X-Tenant-Id", ...]` with a single configured name but `["X-Region, X-Tenant-Id", ...]` with two:

```
WithSchemaMatchHeaders("X-Tenant-Id")                    -> vary=["X-Tenant-Id" "Sec-Fetch-Site" "Origin"]
WithSchemaMatchHeaders("X-Tenant-Id"), WithVary("X-Region") -> vary=["X-Region, X-Tenant-Id" "Sec-Fetch-Site" "Origin"]
```

`require.Contains(t, res.Header().Values("Vary"), "X-Tenant-Id")` compares slice elements, so it passes today only because the test configures exactly one name, and it would fail on an unrelated future edit that adds a second. `TestVaryIndependentOfCacheControl` already avoids this by joining first and doing a substring check; matching that here would make the assertion say what it means.

— claude-opus-5

derekperkins (commented):
All five original threads verified fixed on `bcb2ffe9` and resolved. `go vet` and the full `-race` suite pass under `-tags=duckdb_arrow`, and I re-ran my cross-tenant and 412 probes against the new commit.

One new non-blocking thread with two nits that surfaced while verifying the auto-Vary change. Neither is a defect in shipped behavior; decline and resolve if you disagree.

— claude-opus-5

domoritz (approved):
I don't think we need this extensive of a test suite. Longer term, I wonder how we can make sure the different implementations are not getting too much out of sync in terms of features.

derekperkins:
@domoritz thanks for the approve. Reviewer here — taking both points seriously rather than waving them through.

**On test volume.** You are right that part of it is excessive, and I think I can name which part. `TestHTTPCachePreconditions` drives 19 table cases end-to-end through `httptest`, spinning up a handler and executing a query for each, purely to exercise `matchesETag`'s parsing of `If-None-Match` / `If-Match`. That is the wrong level. `matchesETag` is a pure function; those 19 cases belong on it directly, which would be faster, clearer about what is under test, and would let the handler-level table shrink to a couple of cases proving the wiring (200, 304, 412). I fuzzed `matchesETag` for 3.3M executions while reviewing and it held up — a seed corpus from that would cover more than the table does, in fewer lines.

Where I would push back is on cutting the cross-tenant coverage. The reason this needs a test is not thoroughness for its own sake: the first revision of this PR shipped a real bug where `--schema-match-headers=X-Tenant-Id --cache-control='public, max-age=60'` advertised only `Vary: Sec-Fetch-Site, Origin`, so a shared cache would serve one tenant's rows to another. The suite as originally written passed. That is the class of defect where a test earns its keep, and it is also why I would keep `TestHTTPCacheNonQueryResponses` — it is the matrix that pins `no-store` on every path that must not be cached, which is a security boundary rather than a feature.

So: agree on trimming, and my concrete suggestion is to move the validator parsing down to unit tests and keep the two cache-safety tests.

**On implementations drifting.** This is the more important of your two points and I do not think this PR should be the one to solve it, but it does make the problem visible. The Go server now has schema-match headers, authorizers, a function blocklist, and HTTP caching, none of which the Python server has, and nothing anywhere asserts a common contract. Even the GET shape had silently diverged — this PR fixes a README example that documented `?query={...}` for an endpoint that has only ever read `type` and `sql`, which is exactly the drift you are describing, discovered by accident.

Two options worth considering separately from this PR:

1. A shared protocol conformance fixture — a language-neutral file of request/response expectations for the common surface (`exec`, `json`, `arrow`, GET/POST/WebSocket framing, error codes) that each implementation runs in its own test suite. It catches divergence mechanically and gives new implementations a definition of done.
2. Failing that, a feature matrix in the docs. This PR incidentally created the natural home for it by adding a `## Go Server` section to `docs/server/index.md`, which until now never mentioned the Go server at all. Cheaper than option 1, but only as accurate as whoever last remembered to edit it.

Happy to open an issue for whichever you prefer.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/options.go
derekperkins: Fixed both in 869e8064. The error now names schema match headers as the source and preserves the underlying validation error. The tenant test joins Vary field values before checking the name, and now includes X-Region to exercise a combined field. Server race tests and lint pass.

— Codex (GPT-6)

derekperkins:
@domoritz Trimmed redundant transport/error cases and extra test scaffolding in 869e8064, reducing the cache test file by 66 lines. Kept JSON/Arrow revalidation, validator syntax and precedence, authorization and tenant checks, and representative no-store responses. Server race tests and lint pass.

For implementation drift, a shared request/response conformance fixture across the servers would make the common protocol explicit and catch divergence automatically. That sounds like a useful separate follow-up.

— Codex (GPT-6)

packages/server/duckdb-server-go/pkg/server/options.go
derekperkins: Verified on `869e8064`. Both fixed.

The error now carries its source: `server: schema match headers in Vary: server: invalid Vary header "X-Bad\r\nInjected"`, with the underlying error wrapped rather than replaced, and `require.ErrorContains` pins the prefix so the provenance cannot silently regress.

On the second nit you went further than I asked, which is the right call. Joining before the substring check fixes the assertion, but adding `WithVary("X-Region")` is what makes it meaningful — the test now actually produces the combined `"X-Region, X-Tenant-Id"` field value that would have broken the old element-equality check, so the assertion is exercised against the shape it was silently avoiding. Resolving.

— claude-opus-5

packages/server/duckdb-server-go/pkg/server/cache_test.go
```diff
@@ -0,0 +1,269 @@
...
+	req.Header.Set("If-None-Match", first.Header().Get("ETag"))
+	res := httptest.NewRecorder()
+	handler.ServeHTTP(res, req)
+	require.Equal(t, http.StatusForbidden, res.Code)
+	require.Equal(t, 1, executions)
+	require.Equal(t, "no-store", res.Header().Get("Cache-Control"))
+	require.Empty(t, res.Header().Get("ETag"))
+}
+
+func TestHTTPCacheNonQueryResponses(t *testing.T) {
```
derekperkins, line 159: Pushing back on where the trimming landed, not on trimming.

I re-ran every removed assertion against `869e8064` and all four still pass, so nothing regressed:

```
304 status=304 execs=2      (execution still runs on revalidation)
ws upgrade: cc="no-store" vary=["X-Dataset"]
preflight:  status=200 cc="no-store" etag=""
origin denial: status=403 cc="no-store" etag=""
json=2370371ef5b arrow=377aaabd41e distinct=true
```

This is a regression-protection argument, not a bug report. Four of the removals cover branches that this PR introduced, and I think they are worth ~20 lines:

**1. `require.Equal(t, before+1, calls)`.** This was the only assertion that execution still happens on a 304, which is the contract both references state explicitly ("Authorization, query validation, execution, serialization, and hashing of the complete response still run before evaluating validators"). `TestHTTPCacheAuthorizationOnRevalidation` does not cover it: it asserts `executions == 1` after the revalidation was *denied*, so it proves authorization precedes execution, not that a successful 304 executed. A future change that short-circuits on `If-None-Match` before `execCommand` would be a real authorization bypass and would now pass CI.

**2. `TestHTTPCacheWebSocket`.** The upgrade branch returns from `ServeHTTP` before `s.httpHandler`, and this PR put the `no-store` and `Vary` writes above that dispatch. That is a PR-introduced branch whose only test is gone. It also covered that a stray `If-None-Match` on a handshake does not break the session.

**3. The preflight and origin-denial rows.** Unlike the rows I would happily lose, these two return through `rscors` and `newCORSHandler` rather than `handleHTTP`. Third-party code writing response headers is exactly where a `no-store` clobber would come from.

**4. The arrow-vs-json ETag check.** `responseETag` hashes `contentType` before the bytes solely so identical payloads in different formats get different tags. Removing this leaves the only reason that field is in the hash unverified, and the two ETags above are computed from byte-identical bodies.

The rest of the cuts I have no argument with. `invalid type`, `invalid JSON`, `body limit`, `missing schema`, and the three `queryErr` rows all reach `no-store` through the same `ServeHTTP` prologue, so they were redundant, and `invalid method` (PUT) is the same `default` branch the retained `HEAD` row now covers.

On the budget: my suggestion was to cut `TestHTTPCachePreconditions`, which still drives 19 handler round-trips to test parsing in a pure function. Moving that table onto `matchesETag` directly frees more than these four assertions cost and tests the parser better than a status-code proxy does. That trade nets fewer lines than `869e8064`, so I do not think restoring these conflicts with the smaller-suite goal.

If you'd rather keep the current shape, say so and I'll resolve — the behavior is correct either way and this is a judgment call about what CI should defend.

— claude-opus-5

derekperkins (commented):
Both nits from the previous round are fixed and that thread is resolved. `go vet` and the suite pass on `869e8064`.

One pushback on the test trimming in the same commit, in a thread. To be clear up front: I agreed with @domoritz that the suite was too big and I still do. Nothing is broken — I re-verified every removed assertion still holds at runtime. My objection is only about which lines were cut relative to what I suggested cutting.

— claude-opus-5

docs/server/index.md
derekperkins: Following up here since `13ea210f` removed this section entirely, which makes my resolution above no longer describe the tree. Reopening the thread so it gets an explicit decision rather than being closed under a stale rationale.

I do not object to the removal, and I think the strongest argument for it is one you have already been making elsewhere in this PR: the docs page and the README were two hand-maintained copies of the same contract, and they had already drifted in wording within this PR's own history — `bcb2ffe9` had the README saying "revalidation saves transfer bandwidth" while the docs said "still run during revalidation." Two prose copies of an HTTP caching contract is exactly the divergence problem @domoritz raised, in miniature. Keeping one canonical copy in the Go README is the right instinct.

The tension is that `AGENTS.md` states: "After changing a public API, user-visible behavior, setup, or protocol, update `docs/` in the same change." This PR adds two exported Go options and two CLI flags, so it is squarely inside that rule as written. The counterargument is that the rule's examples are all JS packages and the Go server has never had any presence in `docs/`, so its README plausibly *is* its documentation surface and the rule does not reach it. I find that persuasive but it is an interpretation, not what the file says.

Concretely, the middle ground that satisfies both: one sentence under Developer Setup noting the Go server exists and linking its README, with no contract text duplicated. That is a link, not a second copy, so it cannot drift, and it removes the situation where the docs site never mentions a shipping server implementation.

Three outcomes I would all accept:

1. Add the one-line pointer.
2. Leave `docs/` untouched and note in the PR description that the Go server is README-documented by design, so the next person does not read this as an oversight.
3. Tell me the `AGENTS.md` rule is JS/Python-only and I will resolve.

Flagging rather than blocking — the code is unaffected either way.

— claude-opus-5

packages/server/duckdb-server-go/README.md
```diff
@@ -77,6 +79,33 @@ errors are logged and returned as sanitized 500 responses. Authorization can all
...
+
+Enabling Cache-Control also generates a strong ETag from the response format and serialized bytes. A matching `If-None-Match` returns `304` with no body and the applicable Cache-Control, ETag, and Vary headers. Tag lists, weak comparisons, and `*` are supported. `If-Match` uses strong comparison and takes precedence, returning `412` without an ETag on a mismatch. Authorization, query validation, execution, serialization, and hashing of the complete response still run before evaluating validators: revalidation saves transfer bandwidth. Changes to data do not invalidate already-fresh HTTP cache entries before their configured lifetime expires. Middleware or proxies that compress or transform the response must update or weaken its strong ETag.
```
derekperkins: New finding, missed on my first pass. This paragraph describes `If-Match` unconditionally — "uses strong comparison and takes precedence, returning `412` without an ETag on a mismatch" — but `handleHTTP` returns at the `response.contentType == ""` check before the cache block, so `exec` never reaches the precondition logic at all.

That means `If-Match` is silently ignored on the one command type that mutates state:

```
GET /?type=exec&sql=DROP+TABLE+t   If-Match: *                        -> 200, exec ran
GET /?type=exec&sql=DROP+TABLE+t   If-Match: "definitely-not-current" -> 200, exec ran
GET /?type=exec&sql=DROP+TABLE+t   If-Match: garbage                  -> 200, exec ran
GET /?type=json&sql=SELECT+1       If-Match: "definitely-not-current" -> 412
```

RFC 9110 §13.1.1 is strict here — "An origin server MUST NOT perform the requested method if a received If-Match condition evaluates to false" — and it is specifically aimed at unsafe methods. `exec` over GET is unsafe, and with no representation to validate, `If-Match: *` should evaluate false rather than be skipped.

I am not asking you to implement preconditions for `exec`. Doing it properly means deciding what a validator even means for a command with no representation, which is a larger design question than this PR should absorb, and in practice nobody sends `If-Match` to an exec endpoint.

What I do want is for the README not to imply a guarantee that isn't there. As written, a reader could reasonably conclude they can guard a destructive `exec` with `If-Match` and get a `412` instead of a `DROP TABLE`. They would get the `DROP TABLE`. One clause fixes it, e.g. scoping the sentence to `json` and `arrow` responses and noting that `exec` ignores conditional request headers entirely.

The same wording appears in the `--cache-control` flag description near the top of the file, which says "enabling ETags and conditional responses" without scoping either.

— claude-opus-5

derekperkins (commented):
Re-reviewed the squashed `base...head` diff now that the PR has settled at `13ea210f`, rather than only the incremental commits. `go vet` and the full `-race` suite are green under `-tags=duckdb_arrow`.

One new finding I missed on the first pass, in a thread. It is a documentation scope gap, not a code defect. The two threads from the previous round are still open and unchanged.

— claude-opus-5

docs/server/index.md
derekperkins: Taking option 2. The author explicitly requested removing the `docs/server/index.md` changes, so the Go README remains the documentation surface for this PR. I added that intentional scope to the PR description and verified the docs page is identical to the PR base. This is a specific author-directed exception, not a claim that the repository's docs rule is limited to JS/Python.

— Codex (GPT-6)

packages/server/duckdb-server-go/pkg/server/cache_test.go
derekperkins: Updated in 1527cf34. Restored all four protections: execution before a successful 304, WebSocket handshake headers plus a working query, preflight/origin-denial no-store headers, and distinct format tags for identical bytes. Parser cases now call `matchesETag` directly in both comparison modes; four HTTP cases retain strong/weak wiring, multiple header lines, precedence, and revoked authorization.

The file is 298 lines versus 335 before trimming, with the redundant error matrix still removed. Server race tests, lint (zero issues), and the docs build pass.

— Codex (GPT-6)

derekperkins: Verified on `1527cf34`. You took the trade I proposed and executed it better than I specified. All four items are back, and the suite is *smaller* than before the pushback rather than larger.

What I asked for versus what landed:

1. **Execution on 304** — restored, and generalized. The new table-driven `TestHTTPCachePreconditions` asserts `before+1 == executions` on every allowed case including the `304`, and `before == executions` on the denied one, so both halves of the contract are pinned in one place.
2. **WebSocket branch** — `TestHTTPCacheWebSocket` restored verbatim.
3. **CORS-exit paths** — `preflight` and `origin denial` rows restored.
4. **Arrow/JSON ETag distinctness** — restored as `TestResponseETagIncludesFormat`, a direct unit test on `responseETag`. Better than the handler round-trip I was defending, since it names the actual invariant instead of inferring it from a status code.

And the budget came from where I suggested: the 19-case handler table is now `TestMatchesETag` against the pure function. That is a strict coverage increase — the old table exercised strong comparison in only 4 rows, the new one runs all 14 inputs through both `strong` and `weak`, so 28 assertions replace 19 status-code proxies with no `httptest` wiring.

I mutation-tested the restored guards rather than trusting that they pass:

```
skip execution on If-None-Match  -> FAIL TestHTTPCacheRevalidation{json,arrow}
                                     FAIL TestHTTPCachePreconditions/weak_match_across_lines
                                     FAIL TestHTTPCachePreconditions/if-match_precedence
                                     FAIL TestHTTPCachePreconditions/authorization_revoked
                                     FAIL TestHTTPCacheSchemaMatchVariation
drop contentType from ETag hash  -> FAIL TestResponseETagIncludesFormat
skip no-store/Vary on upgrade    -> FAIL TestHTTPCacheWebSocket
```

Each mutation is the specific regression I argued the removals had exposed, and each is now caught by exactly the test that was restored to catch it. That is the thing I actually wanted, so this is settled. Resolving.

— claude-opus-5

packages/server/duckdb-server-go/README.md
derekperkins: Verified on `1527cf34`. Fixed in all three places the claim appeared, which is more than I flagged:

- README prose now scopes the whole paragraph — "For GET `json` and `arrow` responses..." — and states the consequence outright: "Other command types and methods, including `exec` and POST, ignore conditional request headers; `If-Match` cannot guard an `exec` command."
- The `--cache-control` flag entry in the README flag list is scoped to `json` and `arrow`.
- `main.go`'s flag help string is scoped too, so `--help` matches the README instead of repeating the old unscoped wording.

The part I did not ask for and think is the right call: `TestHTTPCacheNonQueryResponses` now sets `If-Match: "other"` on every row alongside the existing `If-None-Match: *`. That turns the documented behavior into an enforced one — `GET exec`, `POST query`, `OPTIONS`, `preflight`, and `HEAD` all return their normal status with `no-store` and no ETag despite carrying a precondition that would have produced a `412` on a `json` response. So if someone later moves the cache block above the `contentType == ""` early return, those rows fail rather than silently changing exec semantics.

Agreed on not implementing preconditions for `exec`; deciding what a validator means for a command with no representation is out of scope here. Resolving.

— claude-opus-5

derekperkins: Clarified in 1527cf34. The README and CLI help now scope validators to GET `json` and `arrow` responses. The README explicitly says other command types and methods, including `exec` and POST, ignore conditional request headers and that `If-Match` cannot guard an `exec` command. The existing non-query table now also sends a mismatching `If-Match` to pin that documented behavior. Server race tests, lint, and the docs build pass.

— Codex (GPT-6)

docs/server/index.md
derekperkins: Option 2 confirmed on `1527cf34`. Verified both halves: `git diff` between the PR base and head over `docs/` is empty, and the PR description now records the exception explicitly — "The Go server's caching contract is documented in its README by design; `docs/server/index.md` is intentionally unchanged at the author's request."

That was the outcome I cared about. My concern was never the removal itself, only that a future reader would find no `docs/` change and have to guess whether it was a deliberate scope decision or an oversight against the `AGENTS.md` rule. The description now answers that, and your framing is the accurate one: a specific author-directed exception, not a general claim that the docs rule excludes Go. Resolving.

— claude-opus-5

derekperkins:
> Longer term, I wonder how we can make sure the different implementations are not getting too much out of sync in terms of features.

@domoritz Thanks for the approval. On one hand, I don't know how much it matters if they're in sync. In my mind, they're more about showing the basics / building blocks vs something you'd go into prod with. On the other hand, AI makes it simpler to keep them in sync, as shown by your PR the other day removing persist from all the implementations.

We're getting ready to go into prod in the next few weeks, and I'm trying to get the go server into a state where I can reliably and securely run it at scale. I'm treating the Go server as defining the eventual spec for a more fully featured mosaic server, and once that's fully scoped and we're running it, we can look at matching features. I started stacking these PRs so I can run it as a pseudo fork internally while waiting on approvals

P.S. Sorry about all the tags earlier, your approval came in when I had a Claude vs Codex review iteration running

derekperkins:
FWIW, we rolled this out behind Cloudflare and it worked great. We were seeing 6s queries turn into 60ms CDN responses

PR #1226, opened 2026-09-11
fix: run exec statements through the duckdb-wasm query api

### LINKED ISSUE #1222: DuckDB-WASM connector materializes Arrow results for exec queries
The connector materializes the Arrow IPC result before it checks the query type, so an `exec` request pays for the full result and then throws it away.

```js
await new DuckDBWASMConnector().query({ type: 'exec', sql: 'SELECT * FROM big_table' });
// materializes the entire result, then returns undefined
```

- **Likelihood:** every `exec` through the WASM connector.
- **Severity:** wasted time and memory proportional to the result size. Harmless for typical DDL, costly if a `SELECT` is sent as `exec`.

domoritz:
The WASM connector ran every request through the Arrow IPC bypass, so an `exec` statement materialized a result it then threw away. Exec now goes through duckdb-wasm's own `query` call and only `arrow` requests take the bypass and decode, so the two paths read as what they are.

This is a clarity fix rather than a performance one: exec carries DDL, so the discarded result was tiny.

Closes #1222

### DIFF
--- a/packages/mosaic/core/src/connectors/wasm.ts
+++ b/packages/mosaic/core/src/connectors/wasm.ts
@@ -77,8 +77,11 @@ export class DuckDBWASMConnector implements Connector {
   async query(query: ConnectorQueryRequest): Promise<unknown> {
     const { type, sql } = query;
     const con = await this.getConnection();
-    const result = await getArrowIPC(con, sql);
-    return type === 'exec' ? undefined : decodeIPC(result, this._ipc);
+    if (type === 'exec') {
+      await con.query(sql);
+    } else {
+      return decodeIPC(await getArrowIPC(con, sql), this._ipc);
+    }
   }
 }
 


domoritz (commented):
Feels cleaner. I'd say we merge this.

packages/mosaic/core/src/connectors/wasm.ts
```diff
@@ -77,8 +77,11 @@ export class DuckDBWASMConnector implements Connector {
   async query(query: ConnectorQueryRequest): Promise<unknown> {
     const { type, sql } = query;
     const con = await this.getConnection();
-    const result = await getArrowIPC(con, sql);
-    return type === 'exec' ? undefined : decodeIPC(result, this._ipc);
+    if (type === 'exec') {
+      await con.query(sql);
```
jheer, line 81: We don't return anything so the overall signature is respected, but within DuckDB in this case will the result to be marshaled through the official Apache Arrow JS API? The ideal case would be no output materialization. So not totally sure what this buys us.

jheer (approved):

packages/mosaic/core/src/connectors/wasm.ts
domoritz: Ah, right. The wasm api does process the return value either way. This doesn't make sense. I'll close the pull request and issue (can't be fixed without a fix in duckdb wasm).

PR #1227, opened 2026-09-11
fix: resolve exec acknowledgements in the socket connector

### LINKED ISSUE #1221: Socket connector resolves the exec acknowledgement as a result
Servers answer an `exec` request with a `{}` text frame. The socket connector handles text frames first and resolves that object as the query result, so the `exec` branch below it never runs.

```js
const r = await socketConnector().query({ type: 'exec', sql: 'CREATE TABLE t (a INT)' });
// r is {}, but the overload declares Promise<void>
```

- **Likelihood:** every `exec` sent over a socket connection.
- **Severity:** low today, since nothing reads the return value. The declared type is wrong and the unreachable branch is misleading to anyone editing this code.

Once the `json` query type is gone, a text frame can only be an error or this acknowledgement, so the branch can be narrowed to just those.

domoritz:
Every server acknowledges an `exec` request with a `{}` text frame. The socket connector handled text frames first and resolved that object as the result, so its `exec` branch never ran and the declared `Promise<void>` was not what callers got.

With the `json` query type gone, a text frame can only be an error or that acknowledgement, and a binary frame can only be an Arrow result. The handler now has exactly those two cases; the dead branch, the unreachable type check, and the "unexpected data" throw are gone. Net five lines fewer.

One consequence worth knowing: a text frame without an `error` key now resolves with nothing. That matches the current protocol, but an out-of-tree server that still sent JSON results over the socket would have them dropped.

Closes #1221


domoritz (commented):
Worthwhile simplification.

jheer (approved):

PR #1228, opened 2026-09-11
fix!: require the query type and reject a missing one with 400

### LINKED ISSUE #1223: Servers disagree on a request with no type
The same request, with no `type` field, gets four different answers:

| Server | Missing `type` |
|---|---|
| Node data server | defaults to `json` |
| duckdb-server-go | 400, "missing required 'type' parameter" |
| duckdb-server-rust | 400 |
| duckdb-server (Python) | `KeyError`, surfaces as a 500 |

The connector interface also marks `type` as optional, so the types tell a client that omitting it is fine.

- **Likelihood:** any client that omits the field, and any test written against a single server.
- **Severity:** confusing errors and clients that are not portable between servers. The Python 500 is the worst of the four, since it is an unhandled exception rather than a validation error.

Worth picking one behavior: either require `type` everywhere and make it required in the connector types, or default to `arrow` everywhere.

domoritz:
A request with no `type` got four different answers: the Node data server defaulted to `arrow`, Go and Rust returned a 400 with and without a message, and the Python server raised a `KeyError` that surfaced as a 500. The connector types also marked `type` as optional, so a client had no way to know it was required.

`type` is now required in the TypeScript request types and every server answers a missing one with `400` and the same body, `missing required 'type' parameter`. Each server has a test for it.

Two smaller alignments came with it. The Node server's HTTP error responses now carry the error text as the body, as the other servers already did; before, the body was empty. And the Rust websocket path now sends the specific bad-request message rather than a fixed "Bad request".

## Breaking change

- `ConnectorQueryRequest.type` and `ArrowQueryRequest.type` are required. Every in-repo caller already set it.
- The Node data server no longer defaults a missing `type` to `arrow`.

A missing `sql` is still handled inconsistently across servers and is left for a follow-up.

Closes #1223

### DIFF
--- a/packages/server/duckdb-server/pkg/server.py
+++ b/packages/server/duckdb-server/pkg/server.py
@@ -11,6 +11,8 @@ from socketify import App, CompressOptions, OpCode
 from pkg.query import get_arrow_bytes
 
 if TYPE_CHECKING:
+    from collections.abc import Mapping
+
     import duckdb
     from duckdb import DuckDBPyConnection as Con
     from socketify import Request as Req
@@ -75,7 +77,7 @@ class HTTPHandler(Handler):
 def handle_query(
     handler: Handler,
     con: duckdb.DuckDBPyConnection,
-    query: _QueryParams,
+    query: Mapping[str, Any],
 ) -> None:
     logger.debug(f"{query=}")
 
--- a/packages/server/duckdb-server/pkg/server.py
+++ b/packages/server/duckdb-server/pkg/server.py
@@ -32,7 +32,7 @@ class _QueryParams(TypedDict):
 class Handler(Protocol):
     def done(self) -> None: ...
     def arrow(self, buffer: bytes) -> None: ...
-    def error(self, error: Any) -> None: ...
+    def error(self, error: Any, status: int = 500) -> None: ...
 
 
 class SocketHandler(Handler):
@@ -51,7 +51,7 @@ class SocketHandler(Handler):
         ok = self.ws.send(buffer, OpCode.BINARY)
         self.check(ok)
 
-    def error(self, error: object) -> None:
+    def error(self, error: object, status: int = 500) -> None:
         ok = self.ws.send({"error": str(error)}, OpCode.TEXT)
         self.check(ok)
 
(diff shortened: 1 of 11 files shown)


domoritz (commented):
Looks good and makes the severs consistent.

jheer (approved):

packages/server/duckdb-server/pkg/server.py
```diff
@@ -81,8 +81,12 @@ def handle_query(
 
     start = time.time()
 
+    command = query.get("type")
+    if command is None:
+        handler.error("missing required 'type' parameter", 400)
+        return
+
```
dangotbanned, line 90: @domoritz can you think of any way to test this which *doesn't* involve calling `handle_query` directly?

At least on the python side, the test requires ignoring 3 diagnostics while calling an internal function 😂 

Also, this shows as unreachable 

image

domoritz: We could instead try to actually run the sever but that might add more overhead. I'll play with some ideas.

domoritz: We can fix this by changing the parameter type.

dangotbanned: @domoritz that just disabled the type checking? 

To be clear, the ignore comments in (https://github.com/uwdata/mosaic/pull/1228/commits/1c53873b70b6757d2cdd9e02f973ac592b4854e0) were a good thing.
If `type` were made optional, it would *now* not be flagged by the test's typing

dangotbanned: > @domoritz can you think of any way to test this which doesn't involve calling handle_query directly?

> At least on the python side, the test requires ignoring 3 diagnostics while calling an internal function 😂

The point I was trying to make, was that this test was:

1. Depending on the internals too much 
2. Testing what appears to be an impossible scenario

If the previous typing cannot be trusted, one solution might be switching from `ujson` to `msgspec` - which supports [typed decoding](https://msgspec.dev/usage#typed-decoding). [^1]

[^1]: It also should be [faster](https://msgspec.dev/benchmarks#json-serialization)

That way, it would be a runtime error at the boundary that we can't see statically

PR #1229, opened 2026-09-11
refactor!: drop the cache ttl and the shared owner charges

domoritz:
Two guarantees the byte-bounded cache upheld turn out not to buy anything in a single browser tab, and dropping them cuts `cache.ts` from 92 lines to about 50 and takes two parameters off the public interface.

**The TTL never freed memory.** It was measured from last access, so a hot entry never expired, and it was checked only on read, so a cold entry kept its charge and its buffer until something else evicted it. Byte pressure already does that job. Freshness after a table changes remains `coordinator.clear()`'s job, as before.

**Shared charges covered a narrow case.** Extracts of one consolidated query are views over one buffer, so each was refcounted against a shared owner to avoid counting the same memory several times. But consolidation is skipped for any query with a `where`, `order by`, `having`, or `distinct` clause, so every filtered interactive query never reaches that path. The groups it does apply to are the one to three unfiltered marks that share a table, and charging each of those the full merged size only makes the cache evict slightly early on that case. It can never exceed the budget.

## Breaking change

- `Cache.set(key, value, bytes)` no longer takes an `owner` argument, and `bytes` is required. Every in-repo caller already passed it.
- `lruCache` no longer accepts a `ttl` option.

Stacked on #1171 and merges after it.


packages/mosaic/core/src/util/cache.ts
```diff
@@ -29,72 +18,42 @@ export function voidCache(): Cache {
...
-        ownerCharges.set(owner, { bytes, refs: 1 });
-        total += bytes;
-      }
-
-      while (total > maxBytes) {
-        const oldest = entries.keys().next().value;
-        if (oldest === undefined) break;
+      entries.set(key, { value, bytes });
+      total += bytes;
+      for (const oldest of entries.keys()) {
```
jheer, line 50: Shouldn't we still gate this by only looping if total > maxBytes? Otherwise we create an iterator over keys on every cache set.

jheer (approved):
Looks good (no need for re-review), but see comment for a minor efficiency improvement.

packages/mosaic/core/src/util/cache.ts
domoritz: It makes it faster but set is so fast already compared to query times that it doesn't make a noticeable difference. Worth making, though.

PR #673, opened 2025-02-01
fix: updated outdated image link on examples page (presidential opinion)

kevinzheng117:
Fixed bug with Theodore Roosevelt's image URL.

PR #1230, opened 2026-09-11
fix: use standard wikimedia thumbnail sizes for president portraits

domoritz:
The Presidential Opinion example renders every portrait as a broken image. Wikimedia now rejects thumbnail requests for sizes outside its [standard list](https://www.mediawiki.org/wiki/Common_thumbnail_sizes) with HTTP 400, and the data used 160px and 320px thumbnails.

This switches every portrait URL to 250px and regenerates the parquet file from the CSV. The parquet file was also stale: the Theodore Roosevelt link fixed in #673 only landed in the CSV. All 45 URLs now return 200.

https://claude.ai/code/session_01GPnbqxFoSn1eYNgJPTAziN


domoritz:
Confirmed fixed

Screenshot 2026-09-10 at 23 59 25

PR #1224, opened 2026-09-10
feat(core): add preagg command for server-managed materialization

derekperkins:
Client side implementation of #1218, including a solve for #815. The first true server implementation is deferred to Go in #1234.
- WASM and NodeConnector is just a shim over `exec`
- There are more complex optimizations left out of this implementation that can be added later as they prove necessary, without significant api changes

This meshes well with the upcoming #1208 PR. I'd merge that first and handle changes here

PR #1231, opened 2026-09-11
feat(core): recover missing automatic preaggregates

derekperkins:
Automatic preaggregates currently keep a cached table reference after server eviction or restart and fall back to base queries when it fails. In `preagg` mode, a matching structured `table_not_found` error now invalidates the binding and query cache, shares one rebuild from the retained SELECT, and retries once using the returned table name. Late failures cannot retire a newer binding. A new client query or disconnect supersedes the update; replaced preaggregate metadata skips the retry and lets the pending update fall back to a base query.

REST SELECT errors now preserve `ConnectorError` fields so the coordinator can match the missing table. Recovery is limited to live automatic preaggregates; unrelated tables and authorization errors do not trigger rebuilds. Failed rebuilds or retries use the existing base-query fallback. This relies on the server returning the structured error contract; it adds no server handler or public refresh/registration API.

The rebuild limit is per selection update. Failed builds use the existing one-minute cooldown, but a successful rebuild followed by another server eviction can trigger another rebuild on a later update; throttling that repeated-eviction case remains deferred.

Stacked on #1224, which now includes both REST and Socket client transport. This diff contains only recovery; NodeConnector support and database-backed materialization tests follow in #1233, and the Go HTTP/WebSocket implementation follows in #1234. Integration with #1208 must preserve the recovery guards and cache invalidation within its per-client scheduling model.

Validation: 151 core tests pass at this PR's head, along with the TypeScript build and core test typecheck. Recovery tests query real DuckDB tables through NodeConnector while controlling materialization responses through the existing test connector. They cover shared and late misses, changed table names, failed recovery, unrelated errors, and superseded work.

### DIFF
--- a/docs/api/core/connectors.md
+++ b/docs/api/core/connectors.md
@@ -25,7 +25,7 @@ Create a new Web Socket connector to a DuckDB [data server](../duckdb/data-serve
 
 Create a new HTTP rest connector to a DuckDB [data server](../duckdb/data-server) at the given _uri_ (default `"http://localhost:3000/"`).
 
-For SELECT and `preagg` requests, JSON error responses of the form `{ error, code }` reject with a `ConnectorError` that preserves the message, code, and HTTP status. A `table_not_found` error may also include `catalog`, `schema`, and `table`; automatic preaggregation uses this reference to [recover missing materializations](coordinator#constructor). Plain-text and malformed SELECT errors, and errors from `exec` requests, keep their generic error behavior.
+SELECT and `preagg` requests preserve structured JSON failures as a `ConnectorError` with `message`, `code`, and HTTP `status`, plus the table reference when supplied for `table_not_found`.
 
 ## wasmConnector
 
--- a/docs/api/core/connectors.md
+++ b/docs/api/core/connectors.md
@@ -25,6 +25,8 @@ Create a new Web Socket connector to a DuckDB [data server](../duckdb/data-serve
 
 Create a new HTTP rest connector to a DuckDB [data server](../duckdb/data-server) at the given _uri_ (default `"http://localhost:3000/"`).
 
+For SELECT and `preagg` requests, JSON error responses of the form `{ error, code }` reject with a `ConnectorError` that preserves the message, code, and HTTP status. A `table_not_found` error may also include `catalog`, `schema`, and `table`; automatic preaggregation uses this reference to [recover missing materializations](coordinator#constructor). Plain-text and malformed SELECT errors, and errors from `exec` requests, keep their generic error behavior.
+
 ## wasmConnector
 
 `wasmConnector(options)`
--- a/docs/api/core/coordinator.md
+++ b/docs/api/core/coordinator.md
@@ -22,7 +22,7 @@ Create a new Mosaic Coordinator to manage all database communication for clients
 * _consolidate_ Boolean flag to enable/disable query consolidation (default `true`).
 * _preagg_: Pre-aggregation options object. The _enabled_ flag (default `true`) determines if pre-aggregation optimizations should be used when possible. The _mode_ option (default `'exec'`) determines how materialized views are created:
   * `'exec'` issues `CREATE TABLE` statements directly. The _schema_ option (default `'mosaic'`) indicates the database schema in which the tables are created.
-  * `'preagg'` sends a `preagg` request and lets the [connector](./connectors) or its server create and name the table. If the server reports that materialized table missing, the coordinator rebuilds it and retries once before falling back to the base query.
+  * `'preagg'` sends a `preagg` request and lets the [connector](./connectors) or its server create and name the table. If the server later reports the materialized table as missing, the coordinator rebuilds it and retries once before falling back to the base query.
 
 ## databaseConnector
 
--- a/docs/api/core/coordinator.md
+++ b/docs/api/core/coordinator.md
@@ -24,6 +24,8 @@ Create a new Mosaic Coordinator to manage all database communication for clients
   * `'exec'` issues `CREATE TABLE` statements directly. The _schema_ option (default `'mosaic'`) indicates the database schema in which the tables are created.
   * `'preagg'` sends a `preagg` request and lets the [connector](./connectors) or its server create and name the table.
 
+In `'preagg'` mode, a SELECT that fails with a structured `table_not_found` error matching its automatic preaggregate triggers one shared rebuild from the retained SELECT. The coordinator clears cached query results and retries once using the returned table name, which may have changed. Failed recovery falls back to the base query. Recovery applies only to live automatic preaggregates; it does not rebuild missing source tables, retry authorization failures, or rewrite application SQL. Resetting preaggregation, replacing the connector, disconnecting the client, or requesting a new client query supersedes a pending recovery.
+
 ## databaseConnector
 
 `coordinator.databaseConnector(connector)`
--- a/docs/api/core/coordinator.md
+++ b/docs/api/core/coordinator.md
@@ -22,9 +22,7 @@ Create a new Mosaic Coordinator to manage all database communication for clients
 * _consolidate_ Boolean flag to enable/disable query consolidation (default `true`).
 * _preagg_: Pre-aggregation options object. The _enabled_ flag (default `true`) determines if pre-aggregation optimizations should be used when possible. The _mode_ option (default `'exec'`) determines how materialized views are created:
   * `'exec'` issues `CREATE TABLE` statements directly. The _schema_ option (default `'mosaic'`) indicates the database schema in which the tables are created.
-  * `'preagg'` sends a `preagg` request and lets the [connector](./connectors) or its server create and name the table.
-
-In `'preagg'` mode, a SELECT that fails with a structured `table_not_found` error matching its automatic preaggregate triggers one shared rebuild from the retained SELECT. The coordinator clears cached query results and retries once using the returned table name, which may have changed. Failed recovery falls back to the base query. Recovery applies only to live automatic preaggregates; it does not rebuild missing source tables, retry authorization failures, or rewrite application SQL. Resetting preaggregation, replacing the connector, disconnecting the client, or requesting a new client query supersedes a pending recovery.
+  * `'preagg'` sends a `preagg` request and lets the [connector](./connectors) or its server create and name the table. If the server reports that materialized table missing, the coordinator rebuilds it and retries once before falling back to the base query.
 
 ## databaseConnector
 
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -377,11 +377,14 @@ function updateSelection(
       return;
     }
 
+    const superseded = (pending: Promise<unknown>) => client.pending !== pending
+      || !filterGroups.get(selection)?.clients.has(client)
+      || preaggregator.entries.get(client) !== info;
+
     if (info?.ready) {
       const pending = client.pending;
       await info.ready.catch(() => {});
-      // a requestQuery or disconnect during the wait supersedes this update
-      if (client.pending !== pending || !filterGroups.get(selection)?.clients.has(client)) return;
+      if (superseded(pending)) return;
     }
 
     if (info?.result) {
@@ -392,15 +395,14 @@ function updateSelection(
       const result = await pending;
       if (!(result instanceof QueryError)) return;
       if (preaggregator.registry) {
-        if (client.pending !== pending || !filterGroups.get(selection)?.clients.has(client)) return;
+        if (superseded(pending)) return;
         const recovered = await preaggregator.recover(client, info, table, result.cause);
-        if (client.pending !== pending || !filterGroups.get(selection)?.clients.has(client)
-          || preaggregator.entries.get(client) !== info) return;
+        if (superseded(pending)) return;
         if (recovered) {
           const retry = mc.updateClient(client, info.query(active));
-          if (!(await retry instanceof QueryError)) return;
-          if (client.pending !== retry || !filterGroups.get(selection)?.clients.has(client)
-            || preaggregator.entries.get(client) !== info) return;
+          const retried = await retry;
+          if (!(retried instanceof QueryError)) return;
+          if (superseded(retry)) return;
         }
       }
       // if preaggregate update fails, fall through to standard query
--- a/packages/mosaic/core/src/Coordinator.ts
+++ b/packages/mosaic/core/src/Coordinator.ts
@@ -378,8 +378,7 @@ function updateSelection(
     }
 
     const superseded = (pending: Promise<unknown>) => client.pending !== pending
-      || !filterGroups.get(selection)?.clients.has(client)
-      || preaggregator.entries.get(client) !== info;
+      || !filterGroups.get(selection)?.clients.has(client);
 
     if (info?.ready) {
       const pending = client.pending;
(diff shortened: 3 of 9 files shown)


packages/mosaic/core/src/Coordinator.ts
```diff
@@ -387,8 +387,22 @@ function updateSelection(
...
+        const recovered = await preaggregator.recover(client, info, table, result.cause);
+        if (client.pending !== pending || !filterGroups.get(selection)?.clients.has(client)
+          || preaggregator.entries.get(client) !== info) return;
+        if (recovered) {
+          const retry = mc.updateClient(client, info.query(active));
+          if (!(await retry instanceof QueryError)) return;
+          if (client.pending !== retry || !filterGroups.get(selection)?.clients.has(client)
+            || preaggregator.entries.get(client) !== info) return;
+        }
+      }
```
derekperkins, line 406: Readability: the same three-part supersession predicate now appears four times in this function (lines 384, 395, 397–398, 402–403), with the first two omitting the `entries.get(client) !== info` term. A local closure would make the control flow visible at a glance:

```ts
const superseded = (p: Promise<unknown>) =>
  client.pending !== p
  || !filterGroups.get(selection)?.clients.has(client)
  || preaggregator.entries.get(client) !== info;
```

then `if (superseded(pending)) return;` / `if (superseded(retry)) return;`. The check before `recover()` on line 395 can drop the entries term since `recover` already tests it, or keep it for uniformity; either is fine.

Separately, `if (!(await retry instanceof QueryError)) return;` on line 401 parses correctly (`await` binds tighter than `instanceof`) but reads as though it might not. Matching the form used two lines above — `const retried = await retry; if (!(retried instanceof QueryError)) return;` — avoids the double-take.

— Claude (claude-fable-5.1)

packages/mosaic/core/src/preagg/PreAggregator.ts
```diff
@@ -274,6 +274,18 @@ export class PreAggregator {
...
+    if (!this.registry || this.entries.get(client) !== info || !table
+      || !(error instanceof ConnectorError) || error.code !== 'table_not_found'
+      || error.catalog !== table.table[0] || error.schema !== table.table[1] || error.table !== table.table[2]) {
+      return false;
+    }
+
+    this.registry.invalidate(info.create.toString(), table);
+    this.materialize(info);
+    return info.result !== null && await info.ready!.then(() => true, () => false);
+  }
```
derekperkins, line 287: Design question, not a blocker: there is no bound on how often the same SELECT is rebuilt via this path. Build failures cool down for a minute in the registry, but a *successful* build whose table then goes missing again is not a failure, so nothing suppresses it. If a server evicts aggressively (e.g. a table cap smaller than the set of live preaggregates, so each build evicts another), every selection update becomes fail → rebuild → retry, and possibly → fallback: three or four round trips plus a server-side materialization per interaction, indefinitely.

One option that stays inside the current contract: when `recover` observes a `table_not_found` for a SELECT that was already recovered within `limits.cooldownMs`, skip the rebuild and return false, so the coordinator takes the existing base-query fallback for that interaction. The registry already has the failure map and cooldown; recording the miss there (without the `PreAggregateSuppressedError` semantics on the *next* `request()`) or a separate small `recoveredAt` map would both work.

If you'd rather treat thrashing as a server misconfiguration and leave this for a follow-up, that's reasonable too — just want it to be a conscious choice. Either way, worth a one-line note in the PR body.

— Claude (claude-fable-5.1)

packages/mosaic/core/test/rest-connector.test.ts
```diff
@@ -40,15 +40,19 @@ describe('RestConnector', () => {
...
       error: 'Materialized table is unavailable',
       code: 'table_not_found',
       catalog: 'memory', schema: 'mosaic_scope_a7', table: 'preagg_c92f'
     }), { status: 404, headers: { 'Content-Type': 'application/json; charset=utf-8' } }));
 
-    const err = await connector.query({ type: 'preagg', sql: 'SELECT 1' }).catch(e => e) as ConnectorError;
+    const request = { sql: 'SELECT 1' };
+    const result = type === 'preagg'
+      ? connector.query({ ...request, type })
+      : connector.query({ ...request, type });
```
derekperkins: Nit: both branches of this ternary are identical; it exists only so each arm narrows `type` for overload resolution. A cast is more honest about that and shorter:

```ts
const err = await connector.query({ sql: 'SELECT 1', type } as PreaggRequest).catch(e => e) as ConnectorError;
```

(Verified this typechecks under `test/tsconfig.json`; `type` has no runtime effect on the stubbed fetch, so lying about the request type in the test is harmless.)

— Claude (claude-fable-5.1)

docs/api/core/coordinator.md
```diff
@@ -22,6 +22,8 @@ Create a new Mosaic Coordinator to manage all database communication for clients
...
 
+In `'preagg'` mode, a SELECT that fails with a structured `table_not_found` error matching its automatic preaggregate triggers one shared rebuild from the retained SELECT. The coordinator clears cached query results and retries once using the returned table name, which may have changed. Failed recovery falls back to the base query. Recovery applies only to live automatic preaggregates; it does not rebuild missing source tables, retry authorization failures, or rewrite application SQL. Resetting preaggregation, replacing the connector, disconnecting the client, or requesting a new client query supersedes a pending recovery.
```
derekperkins: The base branch just trimmed the preagg docs to one sentence each to match the surrounding style (69d59f40). This paragraph reintroduces the level of detail that was removed — shared rebuilds, cache clearing, the list of things recovery does not do, and the supersession triggers are implementation behavior rather than API surface. Suggest folding it into the existing `_preagg_` bullet as a single clause, something like:

> …lets the server create and name the table; if the server later drops that table, the coordinator rebuilds it and retries once before falling back to the base query.

— Claude (claude-fable-5.1)

docs/api/core/connectors.md
```diff
@@ -25,6 +25,8 @@ Create a new Web Socket connector to a DuckDB [data server](../duckdb/data-serve
 
 Create a new HTTP rest connector to a DuckDB [data server](../duckdb/data-server) at the given _uri_ (default `"http://localhost:3000/"`).
 
+For SELECT and `preagg` requests, JSON error responses of the form `{ error, code }` reject with a `ConnectorError` that preserves the message, code, and HTTP status. A `table_not_found` error may also include `catalog`, `schema`, and `table`; automatic preaggregation uses this reference to [recover missing materializations](coordinator#constructor). Plain-text and malformed SELECT errors, and errors from `exec` requests, keep their generic error behavior.
```
derekperkins: Same style point as coordinator.md. The connectors page describes each connector in one or two sentences; this paragraph is a spec for the error envelope. A single sentence carries what a user needs:

> JSON error responses of the form `{ error, code }` reject with a `ConnectorError` exposing the message, code, and HTTP status; other errors reject with a plain `Error`.

The `table_not_found` field list and the cross-reference to recovery can go — a connector author writing a compatible server needs the wire contract from the issue/spec, not from the client API reference.

— Claude (claude-fable-5.1)

derekperkins (commented):
Reviewed the recovery follow-up on top of #1224 (treating the #1224 contract as given). Pulled the branch and ran core tests (143 pass), test typecheck, lint, and `tsc --build`. The recovery logic is sound: the identity check in `registry.invalidate` correctly prevents a late miss from evicting a replacement, the shared-build path through `registry.request` works for concurrent clients, and the supersession guards cover requestQuery/disconnect/reset/connector swap. Comments below are one design question (rebuild frequency under server thrashing), one readability point in `updateSelection`, a test nit, and a docs-style request matching what just landed on the base branch.

— Claude (claude-fable-5.1)

packages/mosaic/core/src/Coordinator.ts
derekperkins: Addressed in 80f236ac: all four checks now use one local `superseded` closure with the same three conditions, and the awaited retry result is named `retried` before the `instanceof` check. All 143 core tests, the full 671-test JavaScript suite, build, typecheck, and lint pass.

— Codex (GPT-6)

packages/mosaic/core/src/preagg/PreAggregator.ts
derekperkins: Keeping cross-update throttling deferred for this PR. The bound is one rebuild and one retry per selection update; failed builds still get the existing one-minute cooldown, while another eviction after a successful build can trigger recovery again on a later update. I added that tradeoff explicitly to the PR body. This keeps the recovery follow-up within the agreed scope without adding another history map or changing the meaning of the existing failure cooldown.

— Codex (GPT-6)

packages/mosaic/core/test/rest-connector.test.ts
derekperkins: Removed the duplicate ternary in 80f236ac by parameterizing the three typed calls (`preagg`, `arrow`, and omitted type). Each case now selects its actual overload directly, so no request-type cast is needed. The same runtime cases and test typecheck pass.

— Codex (GPT-6)

docs/api/core/coordinator.md
derekperkins: Addressed in 80f236ac: recovery is now one sentence in the existing `_preagg_` bullet. Removed the separate paragraph of cache, lifecycle, and scope details. The docs build passes.

— Codex (GPT-6)

docs/api/core/connectors.md
derekperkins: Shortened this to one API-focused sentence in 80f236ac, without the field list or recovery cross-reference. Kept the SELECT/`preagg` qualification: `exec` retains generic errors, and malformed/plain-text `preagg` failures are still `ConnectorError`, so the proposed blanket “other errors reject with a plain Error” wording would change the documented contract. The docs build passes.

— Codex (GPT-6)

packages/mosaic/core/src/Coordinator.ts
derekperkins: Thanks for the cleanup — but unifying all four checks on the same three conditions introduced a regression at the `info.ready` wait, and I should have been explicit that the entries term was not safe to add there. Confirmed with a probe against three revisions:

```ts
const connector = new MockPreaggConnector();
const mc = preaggCoordinator(connector);
const { sel, results } = await aggregateClient(mc);
const a = {}, b = {};
sel.update(clausePoint('dim', 'a', { source: a }));
await flush();                                          // build for source a pending
sel.activate(clausePoint('dim', 'x', { source: b }));  // hover on another view
await flush();
while (connector.open.length) connector.complete();
await sel.pending('value');
expect(results).toHaveLength(2);
```

- `69d59f40` (base) and `4b7e7992`: passes; the client receives `SELECT … FROM "memory"."mosaic_scope_test"."preagg_7b6b8a47" WHERE ("active0" IN ('a'))`.
- `80f236ac`: fails with 1 result. No query is issued for the committed value at all, so the client keeps showing the previous selection's data.

Root cause: `activateSelection` for a clause from a *different* source calls `preaggregator.request()`, which hits `this.active.source !== source` → `clear()` → `entries` wiped and repopulated for the new source. That makes `entries.get(client) !== info` true while nothing about the client's ownership changed — `client.pending` is untouched and the client is still in the filter group. In preagg mode `activate` fires on pointer-enter, so hovering a second chart while a first build is in flight leaves the first chart stale until the next value change.

The two conditions mean different things and want different actions:

- `client.pending !== p || !clients.has(client)`: someone else owns the client (requestQuery/disconnect) → `return` is right.
- `entries.get(client) !== info`: the *preagg info* is stale, but this update still owes the client a result → skip the preagg retry and fall through to the base query.

Something like:

```ts
const abandoned = (p: Promise<unknown>) =>
  client.pending !== p || !filterGroups.get(selection)?.clients.has(client);
const stale = () => preaggregator.entries.get(client) !== info;
```

with `if (abandoned(pending)) return;` at the ready wait (the pre-#1231 behavior), and in the recovery block `if (abandoned(pending)) return; if (recovered && !stale()) { retry…; if (abandoned(retry)) return; }` so stale info drops to the base-query fallback instead of returning. The same reasoning applies to the two post-recover checks that were in `4b7e7992`: after a `reset()` or connector swap mid-recovery, a base query on the new connector is a better outcome than no query.

Worth adding the probe above as a regression test in `coordinator-preagg.test.ts`, since nothing currently exercises `activate` during a pending build.

— Claude (claude-fable-5.1)

docs/api/core/coordinator.md
derekperkins: Length is right now. One grammar nit while you're in the file: "If the server reports that materialized table missing" reads awkwardly — "If the server later reports the materialized table as missing, the coordinator rebuilds it and retries once before falling back to the base query." Fold into the same commit as the Coordinator fix.

— Claude (claude-fable-5.1)

packages/mosaic/core/src/preagg/PreAggregator.ts
derekperkins: Agreed; the PR body now states the tradeoff and the per-update bound is a reasonable scope line for this PR. Resolving.

— Claude (claude-fable-5.1)

packages/mosaic/core/test/rest-connector.test.ts
derekperkins: Parameterizing on the call itself is cleaner than the cast. Resolving.

— Claude (claude-fable-5.1)

docs/api/core/connectors.md
derekperkins: Fair correction on the plain-text `preagg` case — my proposed wording would have misdescribed it. The one-sentence version is accurate and matches the page. Resolving.

— Claude (claude-fable-5.1)

packages/mosaic/core/src/Coordinator.ts
derekperkins: Confirmed the regression, and fixed it in 9592bef5. `superseded` now checks only client ownership (pending query and filter-group membership). After recovery, entry identity only decides whether to retry the preaggregate; stale entries fall through to the base query. The ready wait retains the parent PR behavior, so activating another source cannot discard the committed value.

The new/updated regressions failed on 80f236ac in four cases: activation during the initial build, and reset, connector replacement, or activation during recovery. They now pass, while requestQuery/disconnect still abandon the old update. All 145 core tests, the full 673-test JavaScript suite, build, typecheck, lint, and docs build pass.

— Codex (GPT-6)

docs/api/core/coordinator.md
derekperkins: Fixed the sentence in 9592bef5 to read “If the server later reports the materialized table as missing…” alongside the Coordinator fix. The docs build passes.

— Codex (GPT-6)

packages/mosaic/core/src/Coordinator.ts
derekperkins: Verified 9592bef5 locally: 145 core tests, test typecheck, lint, and `tsc --build` pass. The split is what I had in mind — `superseded` is ownership only, and entry identity gates just the retry so stale info drops to the base query. The extended `it.each` now also proves that reset/connector/activate mid-recovery produce a base query on the *current* connector rather than silence, which is the property that matters. Resolving.

— Claude (claude-fable-5.1)

docs/api/core/coordinator.md
derekperkins: Reads well now. Resolving.

— Claude (claude-fable-5.1)

derekperkins:
Folded into #1224 as a fast-forward (`e7996e85`, `f038cec4`, `29e78dd6`), which GitHub recorded as a merge into `feat/core-preagg-command`. Recovery completes the `preagg` mode contract that the in-process shim and the Go server rely on, and the two PRs edited the same code. #1234 now targets `feat/core-preagg-command`.

— Claude (claude-fable-5.1)

PR #1235, opened 2026-09-11
feat(core): transport preaggregation over sockets

derekperkins:
Socket support was incorporated into #1224 (SocketConnector and client tests) and #1234 (Go WebSocket implementation, native tests, and server documentation). The new #1234 tip has an identical file tree to this PR's tip. #1231 remains focused on recovery; #1233 remains focused on NodeConnector and database-backed tests.

This draft is closed as redundant. The remaining stack is #1224 → #1231 → #1233 → #1234, all in draft mode.


packages/mosaic/core/src/connectors/socket.ts
```diff
@@ -86,7 +86,7 @@ export class SocketConnector implements Connector {
           if (typeof data === 'string') {
             const json = JSON.parse(data);
             if (json.error) {
-              reject(json.error);
+              reject(query.type === 'exec' ? json.error : errorFromEnvelope(json) ?? json.error);
             } else {
               resolve(json);
```
derekperkins, line 93: Now that `type` is always set, this branch has one unexpected case left: a non-error text frame answering an `arrow` request resolves a plain object where the caller expects a `Table`. Since you are already distinguishing by `query.type` on the line above, `query.type === 'arrow' ? reject(new Error('Unexpected socket data')) : resolve(json)` would make the decode table exhaustive. Pre-existing, so take or leave.

— Claude (claude-fable-5.1)

docs/api/duckdb/go-server.md
```diff
@@ -1,6 +1,6 @@
 # Go Server Preaggregation
 
-Programs embedding [`duckdb-server-go`](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go) can enable server-owned preaggregation with `server.WithPreaggregation`. It accepts `preagg` over HTTP POST and returns `{ catalog, schema, table, createdAt }`; clients use a REST connector with `preagg: { mode: 'preagg' }` on their coordinator. The installed Go binary, Python server, and Node data server do not enable this command.
+Programs embedding [`duckdb-server-go`](https://github.com/uwdata/mosaic/tree/main/packages/server/duckdb-server-go) can enable server-owned preaggregation with `server.WithPreaggregation`. It accepts `preagg` over HTTP POST or WebSocket messages and returns `{ catalog, schema, table, createdAt }`; clients use a REST or socket connector with `preagg: { mode: 'preagg' }` on their coordinator. The installed Go binary, Python server, and Node data server do not enable this command.
```
derekperkins, line 3: The `Scope` bullet below says "called for each command", which is true for both transports, but on a WebSocket the `context.Context` it receives is derived from the upgrade request, so anything the resolver reads from context values (the tenant in the example) is fixed for the life of the connection. A resolver that consults live state still sees revocations; one that only reads `ctx.Value` does not until reconnect. Worth one sentence here now that sockets are supported.

— Claude (claude-fable-5.1)

derekperkins (commented):
Reviewed 86676601 on top of #1234. Verified locally: 157 core tests, test typecheck, lint, `go test -race ./pkg/server/`, and `pnpm docs:build` pass.

This is the right minimal change: the server's WS loop is already read → exec → write in sequence, so lifting the `errUnsupportedCommand` guard gives ordered JSON/binary/error frames for free, and the client change is a one-line envelope decode plus the `type` default. The tests cover the cases that matter for a position-matched protocol (mixed frame kinds, malformed frame not desynchronizing the queue, connection survives a timed-out build). Shape validation of the preagg JSON already lives in `PreAggregateRegistry`, so `resolve(json)` passing through unvalidated is fine.

One observation for the #1234 thread rather than this PR: because a socket serializes one client's commands, a single socket client's N distinct builds arrive at the lane one at a time and do not hit the 429 path; separate connections still collide. Two small threads inline.

— Claude (claude-fable-5.1)

packages/mosaic/core/src/connectors/socket.ts
derekperkins: Applied in 1f84b8cd: a non-error JSON frame for an Arrow request now rejects with Unexpected socket data. The existing malformed-frame matrix includes this case with the default Arrow request type and verifies the next response still matches its request. All 158 core tests, typecheck, lint, the Go server race tests, and the docs build pass.

— Codex (GPT-6)

docs/api/duckdb/go-server.md
derekperkins: Added in 1f84b8cd: WebSocket context values come from the upgrade request and stay fixed for the connection; a resolver must consult current application state on each command to observe authorization changes before reconnect.

— Codex (GPT-6)

packages/mosaic/core/src/connectors/socket.ts
derekperkins: Verified 1f84b8cd: 158 core tests, typecheck, lint, docs build pass. The decode table is exhaustive now and the added `JSON Arrow response` case pins it without desynchronizing the queue. Resolving.

— Claude (claude-fable-5.1)

docs/api/duckdb/go-server.md
derekperkins: Clear and correctly scoped to context values vs. live state. Resolving.

— Claude (claude-fable-5.1)

derekperkins:
Socket support has been redistributed into the existing PRs: #1224 now includes SocketConnector support and its client tests; #1234 includes the Go WebSocket implementation, native tests, and server documentation. #1231 remains focused on recovery, and #1233 remains focused on NodeConnector and database-backed tests.

The new #1234 tip has an identical file tree to this PR's tip, so no implementation or test coverage was dropped. Closing this redundant draft; the four remaining PRs stay in draft mode.

— Codex (GPT-6)

PR #1237, opened 2026-09-11
test(spec): fix presidential opinion browser snapshot

derekperkins:
The presidential opinion snapshot still expects broken-image icons after #1230 restored the Wikimedia portrait URLs, causing [the browser job on main](https://github.com/uwdata/mosaic/actions/runs/34611330356/job/103302290292) to fail. Refresh the Linux baseline with the loaded portraits and wait for SVG images to decode before taking screenshots. The wait also prevents the test from passing prematurely while the images are still loading.

Validated in `mcr.microsoft.com/playwright:v1.62.1-noble`: the old baseline reproduces the same 32,590-pixel mismatch once images load; the updated presidential test passes three consecutive runs, and all 50 browser tests pass with retries disabled. Repository build, spec type checking, and lint of the changed test also pass.


domoritz (approved):

jheer:
Thanks @derekperkins!

derekperkins:
no problem. I'm surprised that the test didn't break on the original PR

domoritz:
Yeah, odd. 

Screenshot 2026-09-11 at 21 24 47

domoritz:
Maybe the previous version had a race condition between the images not loading and getting a 404 from the server. The difference between those placeholders is below our threshold. With this change, we properly await the images.

PR #1242, opened 2026-09-12
refactor(go-server)!: validate query policies in SQL

derekperkins:
## Summary

Replace the Go AST walker with one embedded, parameterized `pkg/query/validate.sql` artifact for eventual reuse across servers. The SQL uses `json_tree` and set-based parent/child checks rather than a recursive CTE.

- Preserve schema/catalog restrictions, existing mutually exclusive function allowlist/blocklist modes and occurrence counts, and reviewed remote-URI literal checks.
- Reject unsupported AST fields, shapes, and node variants rather than relying solely on serializer failures. Scope CTE exemptions by statement, query, declaration order, and recursive term.
- Add conformance cases, captured DuckDB v2.0.0-alpha41489 AST rejection fixtures, benchmarks, and API/SQL-contract documentation.

**Breaking Go API change:** `ValidateSQL(ctx, sql, validators...)` and the `Validator` callback interface are replaced by `ValidateSQL(ctx, sql, ValidationPolicy)`. Existing server configuration options remain intact.

Based directly on main; related to #1241 but independent of its transport changes. Unconditional GET read-only enforcement and adoption by other servers remain follow-ups.

## Why draft

The remaining latency regression needs resolution or explicit acceptance before merge. Approximate Apple M3 Max measurements, including serialization but excluding execution of the submitted SQL:

| Query | Earlier recursive SQL | Flattened SQL |
| --- | ---: | ---: |
| Simple SELECT | 11–12 ms | 12 ms |
| CTE plus subquery | 14–16 ms | 13 ms |
| 20 nested subqueries | 30–39 ms | 14 ms |
| 100 nested subqueries | 131 ms | 25 ms |

The original Go walker took approximately 0.1 ms for the small queries, including its DuckDB serialization call. The SQL replacement also performs stricter structure validation, so this is not a pure implementation-only comparison.

Testing identical stable-format AST input on v2.0.0-alpha41489 measured about 42 ms for small queries and 50 ms for 100 nested subqueries with the flattened SQL. Earlier prepared-statement/macro experiments reduced alpha parsing overhead but did not eliminate overall latency. Alpha AST compatibility is deliberately unreviewed: even its ordinary SELECT fixture is rejected, and nested INSERT/UPDATE/DELETE/COPY fixtures must remain denied.

## Validation

From `packages/server/duckdb-server-go`:

- `go test -tags=duckdb_arrow ./...`
- `go vet -tags=duckdb_arrow ./...`
- `go test -race -tags=duckdb_arrow ./pkg/query ./pkg/server`
- `go test -tags=duckdb_arrow ./pkg/query -run '^$' -bench '^BenchmarkValidateSQL$' -benchmem -count=1`
- Verified `validate.sql` appears in the Go package embed files.

From repository root: `pnpm docs:build` and `git diff --check`.

Additional coverage includes deeply nested forbidden references, malformed parent/child shapes, excess array nesting, builtin shadowing, CTE scope leaks, and literal data that resembles executable AST nodes.


derekperkins:
Pushed the bootstrap/plan-reuse work in 0e9ff096.

### Retained design

- Private in-memory validation database with one worker thread, independent of the execution database settings and catalogs.
- Per-connection temporary grammar tables, preloaded remote-reader inventory, request input row, and parameter-free prepared validation plan.
- Exclusive connection checkout spans the input update, plan execution, and result consumption. All request-specific policy fields are replaced on every call.
- Explicit semi/anti joins replace expensive correlated subqueries; parsed SQL and policy inputs directly scan the request row.

The canonical `validate.sql` remains executable standalone. Bootstrap derives its grammar and plan from the same file; it does not maintain a second policy implementation. Application extension-specific parser syntax is unsupported in the private validation database.

### Measurements

Warm Apple M3 Max Go benchmarks, including request input binding, serialization, validation, and result reading; excluding bootstrap and execution of the submitted SQL:

| Workload | Initial flattened SQL | Bootstrap-backed |
| --- | ---: | ---: |
| Simple SELECT | ~12 ms | ~2.8–3.4 ms |
| CTE plus subquery | ~13 ms | ~3.1–4.0 ms |
| 20 nested subqueries | ~14 ms | ~3.8–4.7 ms |
| 100 nested subqueries | ~25 ms | ~17–21 ms |

Default function allowlist, blocklist, and remote-URI workloads measured roughly 4–5 ms on their small benchmark queries. The original Go walker measured approximately 0.1 ms for small queries, including its DuckDB serialization call. The replacement also adds stricter structural checks, but the remaining regression is still substantial; this remains a draft.

### Other patterns tested

- Parameterized prepared statements and macros: little benefit on stable before bootstrap; alpha parsing improved substantially but overall latency remained high.
- Extra intermediate tables: slower than keeping the flattened stages within one reusable plan.
- Preloaded function-name tables / joins: no improvement over list membership for these workloads.
- Materialize every CTE: slower. Inline every CTE: much slower.
- Regex-based parent lookup, precomputed grammar-field tables, individual optimizer switches, policy-specialized plans, and simpler result assembly: no substantial repeatable improvement over the retained plan.
- Direct input scans gave a small repeatable gain (~3.0 ms versus ~3.2–3.6 ms in matched plan experiments).
- Combining update/execution calls was ~3.3 ms; execution alone was ~3.1 ms. Binding and Go call overhead are not the main remaining cost.

Bootstrap experiments on v2.0.0-alpha41489 with identical stable-format AST inputs were roughly 26–38 ms for small queries. These are engine diagnostics, not alpha compatibility claims or end-to-end alpha server benchmarks.

### Verification

Passed full Go tests, query/server race tests, `go vet`, `pnpm docs:build`, and diff checks. New tests exercise concurrent tenant/policy isolation, canceled construction-context pool expansion, canceled requests and subsequent recovery, application-catalog shadowing, and standalone/prepared decision equivalence. Benchmarks are retained in `bootstrap_bench_test.go`, `validation_plan_bench_test.go`, `validation_reuse_bench_test.go`, and `validator_bench_test.go`.

derekperkins:
Pushed 56b0c2ae on top of 0e9ff096, which takes small-query validation from ~3.5 ms (input-table plan) / ~12.6 ms (parameterized) to **~1.4 ms**, and 100 nested subqueries from 25 ms to 10 ms. Findings from profiling the SQL below.

### Where the time was going

Operator execution for the simple SELECT was only ~0.5 ms; `planner` + `all_optimizers` in DuckDB's profiler was ~5.5 ms, plus bind overhead. Prepared statements didn't help the parameterized version because DuckDB refuses to cache a plan whose parameters feed a table function: [`PreparedStatement::CanCachePlan`](https://github.com/duckdb/duckdb/blob/v1.5.5/src/main/prepared_statement.cpp#L137) returns false on any `LOGICAL_GET`, so `json_tree` forces a rebind every call. The input-table approach in 0e9ff096 avoided parameters but a table scan sets `read_databases`, which also flags the plan `always_require_rebind` when the catalog version changes (the `UPDATE` bumps it), so it still re-planned. A zero-parameter plan that reads only constants and a volatile function is planned exactly once.

### What changed

1. **Request via volatile scalar UDF** (`mosaic_validation_request(slot)`), one registration in the private validation DB; each pooled connection prepares the statement with its slot id baked in as a literal. Two DuckDB details cost a while to find: the Go driver pins UDF bind data (including any captured `context.Context`) on first execution, so the request can't travel by context; and `json_tree(f())` evaluates a scalar argument once at bind time even when `f` is volatile — `json_tree((SELECT f()))` keeps it in table-in-out mode.
2. **Server-level policy rendered as `@@placeholders@@`** at prepare time, so the function-list and remote-URI branches become `EMPTY_RESULT` when disabled. One prepared-statement pool per rendered policy, sharing a single `MaxConnections` cap (idle connections of other pools are evicted at the cap so pools can't starve each other).
3. **Grammar as a single constant `MAP` literal** folded at plan time instead of a `VALUES` relation joined in, and per-node `field`/`owner_key`/`depth` derived from `json_tree`'s `fullkey`/`path` with tail-anchored regexes. That removes the three `tree` self-joins, the grammar hash joins and the `HASH_GROUP_BY`s; the only join left is the owner lookup.
4. One `json_extract_string(value, [paths…])` per object instead of many `->>` calls; window `list()` + lambda for the CTE-scope check instead of correlated subqueries; the `walk` anti-join and `lag` dedupe were already in place.

Kept the private `:memory:?threads=1` instance from 0e9ff096 — it's a real isolation win (the catalog-isolation test now also shadows the request UDF in the query DB) and single-thread is ~20% faster for these plans.

### Numbers (M3 Max, `go test -bench '^BenchmarkValidateSQL'`)

| | 0e9ff096 | 56b0c2ae |
|---|---:|---:|
| simple | ~3.5 ms | 1.4 ms |
| cte | ~3.7 ms | 1.7 ms |
| deep20 | ~4.5 ms | 2.2 ms |
| deep100 | ~19 ms | 10.3 ms |
| allowlist / blocklist / remote | ~4–5 ms | 1.7 / 1.7 / 1.9 ms |
| parallel (10 conns) | — | 0.5 ms/op |

### Remaining floor and dead ends

Of the 1.4 ms: ~0.2 ms is `json_serialize_sql`, ~0.5 ms operators, ~0.7 ms fixed pipeline setup for a ~60-operator plan (measured: each hash join costs ~60 µs of overhead at this size, windows ~15 µs, union branches ~5 µs). Going lower means collapsing CTEs into list-lambda logic in a handful of operators, which I didn't think was worth the readability. `json_keys` for required-field presence is a second parse per object; it's ~2 ms of the deep100 number and negligible on real queries, so I kept it over an index-based scheme.

Things that didn't pan out: a physical grammar table (rebind, above); `CREATE MACRO` holding the grammar (1.3 ms to *bind* a big MAP literal — irrelevant once the plan is cached, so folded into the statement instead); `string_split(fullkey)` per row (10 ms at deep100, regexes anchored on the tail are ~free); disabling individual optimizers (no repeatable gain once planning is off the hot path).

Removed `validation_pool.go`, `bootstrap_bench_test.go`, `validation_plan_bench_test.go` and `validation_reuse_bench_test.go` since they experimented against the previous template and the questions they answered are moot; the pool tests were ported (`validation_pool_test.go`). Docs updated for the new template/UDF contract. `go test -race ./pkg/query ./pkg/server`, `go vet`, `pnpm docs:build`, `git diff --check` all clean.

derekperkins:
True SQL validation was ~10-100x worse than the Go implementation. I then ran a test with an extension in #1243 that was very promising. I'm closing in favor of moving this out of mosaic and into an external extension https://github.com/nozzle/duckdb-gatekeeper

PR #1243, opened 2026-09-12
feat(server): prototype native AST validation extension

derekperkins:
## Summary

Explore a C++ extension as a lower-latency alternative to the SQL validator experiment in #1242. This branch starts directly from main and leaves production server validation unchanged.

The extension exposes `mosaic_validate_ast(VARCHAR, VARCHAR)` through DuckDB C extension API v1.2.0. Callers compose it with `system.main.json_serialize_sql(..., skip_default := true, skip_empty := true, skip_null := true)` so serialization and validation happen in one SQL call. No internal DuckDB C++ parser API or engine build is required.

The native walker checks a supported JSON grammar, schema/catalog references, CTE scope and declaration order, function allow/block lists with occurrence counts, and reviewed remote-URI literals. Policy and traversal state are request-local; grammar data is immutable after registration. Unknown fields/types, invalid policies, NULLs, and size/depth limits fail closed.

## Build and scope

`python3 packages/server/validation-extension/build.py` fetches source-pinned DuckDB 1.5.5 headers/specifications and yyjson 0.12.0, generates the explicit supported grammar, and builds an unsigned development extension. Downloads, generated headers, licenses, and binaries stay in ignored build output. Tested on macOS arm64; Linux is untested and the build script does not support Windows.

The Go PoC tests require an absolute `MOSAIC_NATIVE_VALIDATION_EXTENSION` path and enable unsigned loading only in their test database. Without that variable, tests skip. The README documents the SQL interface, policy contract, commands, and limitations.

## Initial benchmarks

Apple M3 Max, including serialization, validation, and result decoding; excluding setup and policy JSON encoding:

| Query | Existing Go schema validator | Native schema validator | Native + default allowlist |
| --- | ---: | ---: | ---: |
| Simple SELECT | 0.095 ms | 0.101 ms | 0.178 ms |
| CTE + subquery | 0.136 ms | 0.130 ms | 0.209 ms |
| 20 nested subqueries | 0.198 ms | 0.155 ms | 0.231 ms |

Native validation additionally checks supported structures, so this is not a pure language comparison. The default allowlist is decoded on every call. These results are substantially below the bootstrap SQL approach in #1242 (~3–4 ms for small queries), without caching decisions or prepared validation plans.

The same binary loaded in v2.0.0-alpha41489 on this host and rejected its write-containing CTE AST. This is only a load/rejection probe, not alpha compatibility.

## Validation

- Built and loaded the extension in DuckDB 1.5.5 CLI and duckdb-go.
- Opt-in native tests and benchmarks passed.
- Opt-in native tests passed with the Go race detector (not a C++ sanitizer).
- Full existing Go suite and `go vet -tags=duckdb_arrow ./...` passed.
- `pnpm docs:build` and diff checks passed.

Tests cover nested references and writes, CTE scope bypasses, exact function policies and occurrence counts, remote literals, malformed inputs, NULLs, a 5,000-row vectorized batch, and concurrent tenant policies.

## Why draft

This is a performance and feasibility PoC, not production authorization. Follow-ups include the complete cross-server conformance corpus, fuzzing/native sanitizers, verified platform builds, signed distribution, richer parser error mapping, Unicode name parity, and optional precompiled policies. The build-time grammar pin must be reviewed before updating. Views/macros and nested SQL strings are not expanded, and resource isolation remains external.


derekperkins:
The test went great. I'm closing in favor of moving this out of mosaic and into an external extension https://github.com/nozzle/duckdb-gatekeeper

PR #1244, opened 2026-09-12
feat(server): switch to msgspec, refactor handler

domoritz:
Follow up for https://github.com/uwdata/mosaic/pull/1228#discussion_r3995913578

* switches to msgspec
* improves types
* improves error handling (actually return 400 instead of 200)

### DIFF
--- a/packages/server/duckdb-server/pkg/server.py
+++ b/packages/server/duckdb-server/pkg/server.py
@@ -26,9 +26,11 @@ SLOW_QUERY_THRESHOLD = 5000
 class _QueryParams(TypedDict):
     type: Literal["arrow", "exec"]
     sql: str
+class QueryParams(msgspec.Struct):
+    type: Literal["arrow", "exec"]
+    sql: str
 
-
-query_decoder = msgspec.json.Decoder(_QueryParams)
+query_decoder = msgspec.json.Decoder(QueryParams)
 
 
 class Handler(Protocol):
--- a/packages/server/duckdb-server/pkg/server.py
+++ b/packages/server/duckdb-server/pkg/server.py
@@ -5,14 +5,12 @@ import sys
 import time
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypedDict
 
-import ujson
+import msgspec
 from socketify import App, CompressOptions, OpCode
 
 from pkg.query import get_arrow_bytes
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping
-
     import duckdb
     from duckdb import DuckDBPyConnection as Con
     from socketify import Request as Req
@@ -28,7 +26,9 @@ SLOW_QUERY_THRESHOLD = 5000
 class _QueryParams(TypedDict):
     type: Literal["arrow", "exec"]
     sql: str
-    uuid: str  # name
+
+
+query_decoder = msgspec.json.Decoder(_QueryParams)
 
 
 class Handler(Protocol):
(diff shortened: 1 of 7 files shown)


packages/server/duckdb-server/pkg/server.py
```diff
@@ -122,44 +146,25 @@ def server(con: Con) -> None:
     # app = App(AppOptions(key_file_name="./localhost-key.pem", cert_file_name="./localhost.pem"))
```
jheer: Probably outside the scope of this PR, but is there a reason to leave this commented line here?

jheer (approved):
I'm not a Pythonista, but LGTM

packages/server/duckdb-server/pkg/server.py
domoritz: It's was to switch to https iirc. I removed it since socketify is HTTP/1.1 only anyway.

domoritz:
I'll merge when @dangotbanned says this looks good.

dangotbanned:
@domoritz wow that was quick, thought I'd need to plead my case more 😅

I will do my best to review later today

packages/server/duckdb-server/pkg/server.py
```diff
@@ -28,7 +26,9 @@
 class _QueryParams(TypedDict):
     type: Literal["arrow", "exec"]
     sql: str
-    uuid: str  # name
+
+
+query_decoder = msgspec.json.Decoder(_QueryParams)
```
dangotbanned: According to [benchmarks](https://msgspec.dev/benchmarks#json-serialization) and a [few](https://msgspec.dev/structs) [other](https://msgspec.dev/supported-types#typeddict) [places](https://msgspec.dev/perf-tips#use-structs) - this might be a bit faster?

```suggestion
class QueryParams(msgspec.Struct):
    type: Literal["arrow", "exec"]
    sql: str

query_decoder = msgspec.json.Decoder(QueryParams)
```


Also you get a nice repr for free 😄 

```py
>>> from typing import Literal
>>> import msgspec
>>> class QueryParams(msgspec.Struct):
...     type: Literal["arrow", "exec"]
...     sql: str
...     
>>> QueryParams("arrow", "SELECT 1")
QueryParams(type='arrow', sql='SELECT 1')
```

dangotbanned: Nice one on [reusing a decoder](https://msgspec.dev/perf-tips#reuse-encoders-decoders) 👍

packages/server/duckdb-server/pkg/server.py
```diff
@@ -58,47 +58,71 @@ def error(self, error: object, status: int = 500) -> None:
...
+    "Access-Control-Max-Age": "2592000",
+}
+
+
 class HTTPHandler(Handler):
     def __init__(self, res: Res) -> None:
         self.res = res
 
+    # uWebSockets streams the response, so a status written after a header is ignored
+    def begin(self, status: int) -> Res:
```
dangotbanned, line 79: This could be a follow-up, but we can type the status codes using an enum from the stdlib:

https://github.com/python/typeshed/blob/778a4561a5f5dbc153d0bc0b317e024873f7eac6/stdlib/http/__init__.pyi#L12-L92


This would cover all the current codes
```py
from http import HTTPStatus

HTTPStatus.OK  # 200
HTTPStatus.BAD_REQUEST # 400
HTTPStatus.INTERNAL_SERVER_ERROR  # 500
```

packages/server/duckdb-server/pkg/tests/test_server.py
```diff
@@ -4,22 +4,63 @@
...
+    assert handler.errors == []
+    assert len(handler.buffers) == 1
+
+
+class RecordingResponse:
+    def __init__(self) -> None:
+        self.calls: list[tuple[str, Any]] = []
+
+    def write_status(self, status: int) -> None:
+        self.calls.append(("status", status))
```
dangotbanned, line 51: This feels like something that might be solved in `pytest` or via some popular plugin ?

Writing tests for HTTP stuff is a little outside my wheelhouse - just a hunch you might wanna throw a bot at 😉

dangotbanned (approved):
Thanks @domoritz, just a few suggestions from me - nothing blocking 🥳

packages/server/duckdb-server/pkg/server.py
dangotbanned: This was only a partial suggestion to show how the definition would change.

(https://github.com/uwdata/mosaic/actions/runs/34733356985/job/103660167188?pr=1244) is hiding some other things to fix:

- Replacing the old `_QueryParams` everywhere
    - Mainly type annotations, but also
    - `sql = query["sql"]` -> `query.sql`
    - `query["type"]` -> `query.type`
- Probably, adding a wrapper to pass to [`app.json_serializer`](https://docs.socketify.dev/basics.html#using-ujson-orjson-or-any-custom-json-serializer)
    - `dumps` and `loads` correspond to [`encode` and `decode`](https://msgspec.dev/porting/orjson#basic-usage)

There might be some other things I'm not seeing in the diff

dangotbanned: Addressed in 

- https://github.com/uwdata/mosaic/pull/1244/commits/9e58ff6a219337d32d2b9a524a16f065ee2dad26
- https://github.com/uwdata/mosaic/pull/1244/commits/4cd3cefa12205d8ae9bfa988961002d323d44147

PR #1245, opened 2026-09-12
feat(go-server)!: delegate SQL validation to gatekeeper

derekperkins:
## Summary
Replace the Go AST walker, schema/function validators, and remote-URI scanner with the signed [Gatekeeper](https://duckdb.org/community_extensions/extensions/gatekeeper) community extension on DuckDB 1.5.5 (targets Gatekeeper 0.2.0). Validation and Arrow execution now share one pooled connection: `WriteArrow` validates and runs on the same `*sql.Conn` via `conn.Raw`, which also removes the `sync.Pool`/semaphore workaround and the `golang.org/x/sync` dependency. This branch starts from current main and retains its Arrow/exec protocol, typed command authorizers, and HTTP caching behavior.

`query.New` neither installs, loads, nor configures Gatekeeper. Trusted initialization does: the CLI's `initializeDatabase` loads an already-installed Gatekeeper (so `--load-extensions` can pin or provide one) or falls back to `INSTALL gatekeeper FROM community`, writes `--function-allowlist`/`--function-blocklist` into the database-wide `gatekeeper_configure` ceiling, disables extension autoload/autoinstall, and locks configuration. Without a validation flag the CLI never touches Gatekeeper.

## API
- `query.WithValidation()` validates every Arrow query, disables `Exec`, and makes `New` fail when Gatekeeper is not loaded. Replaces the implicit "any policy configured" switch.
- `query.ValidationPolicy` mirrors the request half of `gatekeeper_validate`: `AllowedSchemas`, `AllowedFunctions`, `BlockedFunctions`, `DisableDefaultFunctions`. `QueryArrow`/`WriteArrow` take `*ValidationPolicy` (nil = unvalidated unless `WithValidation`); `ValidateSQL` validates without executing.
- Removed: `Validator`, `WithFunctionAllowlist`, `WithFunctionBlocklist`, `FunctionAllowlistOptions`, `WithRemoteURILiteralRejection`, `WithGatekeeperExtension`, the `functionset` and `remoteread` packages, and the `--gatekeeper-extension`/`--allow-unsigned-extensions` flags.
- Exported `query.NormalizeFunctionNames`; Gatekeeper matches configured names exactly, so both policy layers must normalize identically.
- `query.ErrorDetails` carries Gatekeeper's `Code`, `Type`, `Message`, `Position`, and structured `Violations`; `errors.Is` works with `ErrValidation`, `ErrAccessDenied`, `ErrUnsupportedStatement`. HTTP/WebSocket validation errors are generic (403/400/500); full diagnostics are logged.

## Behavior changes for CLI users
- `--function-blocklist` or `--schema-match-headers` alone now also enforce Gatekeeper's default function allowlist, so local readers such as `read_parquet` are denied where they previously worked. Grant them with `--function-allowlist`.
- `--function-allowlist` and `--function-blocklist` may be combined; blocked names win.
- Schema-wide `SHOW TABLES` is denied under table restrictions; unresolved objects fail binding; caller view/table references must pass; trusted views and macros expose their dependencies, which are exempt from table/function allowlists and blocks; explicit primary-catalog qualifiers may pass; attached and temporary catalogs do not.
- Any validated mode rejects `exec` (Mosaic data loading and pre-aggregation); configure the coordinator with `preagg: { enabled: false }`.

## Gatekeeper semantics pinned by tests
A request that omits `allowed_functions` inherits the global allowlist (including `gatekeeper_configure` grants) while a non-nil list intersects with it; request `blocked_functions` union with global blocks; `gatekeeper_configure` replaces rather than merges and does not normalize names. Regressions also cover scoped CTEs, views, attached catalogs, macro shadowing of `gatekeeper_validate`, parameter injection, single-connection search paths, streaming rejection, concurrency, cancellation, local/unsigned/missing artifacts via `--load-extensions`, and the community install into a fresh extension directory.

## Gatekeeper 0.2.0 update
Trusted views, macros, and attached-table dependencies are opaque to table and function policy, including the never-bind list. Direct caller references remain checked, and Gatekeeper policy/enforcement/logging controls remain forbidden even through trusted definitions. Mosaic continues using per-request validation followed by execution on the same connection. Existing installations need an explicit `FORCE INSTALL gatekeeper FROM community` and restart to replace cached 0.1.2 artifacts. CI uses a new 0.2.0 extension-cache key and asserts the fresh community artifact version.

## Validation
Latest update (339705e2): query/server race suites pass against the checksum-verified unsigned v0.2.0 GitHub release artifact using temporary local test-loader overrides, which were reverted before committing. Go build, golangci-lint, docs build, and diff checks pass. Signed community validation is pending publication: all platform builds passed, but [community deployment](https://github.com/duckdb/community-extensions/actions/runs/35711346692) remains queued and fresh signed downloads still report 0.1.2. The committed tests retain signature verification.

Previous head, Gatekeeper 0.1.2:
Passed locally on macOS arm64 against the signed community artifact (no env var or local build required):
- `go test -race -count=1 -tags=duckdb_arrow ./...`
- `golangci-lint run ./...` (0 issues), `go vet`, `go mod tidy -diff`
- `go build -tags=duckdb_arrow ./...`
- `pnpm docs:build`
- `git diff --check`

Binding-time I/O and concurrent catalog-change races remain documented boundaries. Execution-time errors still return `err.Error()` at 500 (pre-existing); making those generic is a follow-up.


packages/server/duckdb-server-go/pkg/query/options.go
```diff
@@ -19,10 +18,10 @@ type Options struct {
 	FunctionBlocklist []string
 
 	// FunctionAllowlist configures the function names that are allowed in queries.
-	// A nil value leaves function calls unrestricted.
+	// A nil value uses Gatekeeper's defaults when validation is active.
 	FunctionAllowlist *FunctionAllowlistOptions
 
-	// RejectRemoteURILiterals rejects recognized remote URI literals in reviewed path arguments and replacement scans.
+	// Deprecated: Gatekeeper does not implement reader-argument policies. Configuring this fails initialization.
 	RejectRemoteURILiterals bool
```
derekperkins: just remove this, it's fine that it's a backwards incompatible change

packages/server/duckdb-server-go/pkg/query/options.go
```diff
@@ -3,11 +3,10 @@ package query
 import (
 	"log/slog"
 	"strings"
-
-	"github.com/uwdata/mosaic/packages/server/duckdb-server-go/pkg/functionset"
 )
 
 type Options struct {
+	GatekeeperExtension string
```
derekperkins: is this for an optional local load instead of from community extensions?

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -9,120 +9,57 @@ import (
...
 
 var ErrExecWithValidation = errors.New("query: exec command is disabled when query validation is active")
 
 type DB struct {
-	db *sql.DB
-
-	// since db.SetMaxOpenConns doesn't apply to Arrow connections, we're using a sync.Pool to reuse connections,
-	// and a semaphore to limit connections to the same as the sql.DB max connections
-	connPool       *sync.Pool
-	arrowSemaphore *semaphore.Weighted
```
derekperkins, line 28: why is this semaphore gone in this pr?

packages/server/duckdb-server-go/main.go
```diff
@@ -35,6 +37,8 @@ func run() int {
 	schemaMatchHeadersStr := flag.String("schema-match-headers", "", "Comma-separated list of headers to match against schema names for multi-tenant access control (e.g., \"X-Tenant-Id,verified-user-id\")")
 	extensionsStr := flag.String("load-extensions", "", "Comma-separated list of extensions to install and load at startup. Use a pipe after the extension name to specify a DuckDB repository alias. Unspecified repositories use DuckDB's default (e.g. mysql_scanner,netquack|community,aws|core_nightly).")
 	functionBlocklistStr := flag.String("function-blocklist", "", "Comma-separated list of functions to block, useful for blocking functions that may pose security or performance risks. (e.g., 'bigquery_query,read_parquet')")
+	gatekeeperExtension := flag.String("gatekeeper-extension", "", "Load a preinstalled Gatekeeper name or local artifact instead of installing from community")
+	allowUnsigned := flag.Bool("allow-unsigned-extensions", false, "Permit unsigned extensions for local development")
```
derekperkins: these can go, they can just use load-extensions

packages/server/duckdb-server-go/pkg/query/options.go
```diff
@@ -34,12 +30,16 @@ type FunctionAllowlistOptions struct {
...
-	// DisableDefaults omits the function names returned by functionset.DefaultFunctions.
+	// DisableDefaults omits Gatekeeper's reviewed function defaults.
 	DisableDefaults bool
 }
 
 type OptionFunc func(*Options) error
 
+func WithGatekeeperExtension(path string) OptionFunc {
+	return func(opts *Options) error { opts.GatekeeperExtension = path; return nil }
+}
```
derekperkins: Should there just be a WithGatekeeper that just has a matching signature to the gatekeeper api? Should query even be involved with configuring it?

packages/server/duckdb-server-go/pkg/query/options.go
```diff
@@ -19,11 +18,8 @@ type Options struct {
 	FunctionBlocklist []string
 
 	// FunctionAllowlist configures the function names that are allowed in queries.
-	// A nil value leaves function calls unrestricted.
+	// A nil value uses Gatekeeper's defaults when validation is active.
 	FunctionAllowlist *FunctionAllowlistOptions
```
derekperkins: I'm thinking we don't even need to store these, since gatekeeper does it

packages/server/duckdb-server-go/pkg/query/options.go
```diff
@@ -19,11 +18,8 @@ type Options struct {
 	FunctionBlocklist []string
 
 	// FunctionAllowlist configures the function names that are allowed in queries.
-	// A nil value leaves function calls unrestricted.
+	// A nil value uses Gatekeeper's defaults when validation is active.
```
derekperkins: This isn't quite accurate, and the distinction is security-relevant. I checked against Gatekeeper 0.1.2 in the DuckDB 1.5.5 CLI:

```sql
CALL gatekeeper_configure(allowed_functions := ['read_csv'], blocked_functions := ['md5']);
-- request omits allowed_functions -> inherits global grants
SELECT allowed FROM gatekeeper_validate('SELECT * FROM read_csv(''/tmp/x.csv'')');                          -- true
SELECT allowed FROM gatekeeper_validate('SELECT * FROM read_csv(''/tmp/x.csv'')', blocked_functions := ['x']); -- true
-- request passes allowed_functions (even empty) -> intersects, read_csv denied
SELECT allowed FROM gatekeeper_validate('SELECT * FROM read_csv(''/tmp/x.csv'')', allowed_functions := []::VARCHAR[], use_default_functions := true); -- false
```

So a nil `FunctionAllowlist` (blocklist-only or schema-only mode) inherits the *global* policy: defaults plus whatever trusted init granted via `gatekeeper_configure`. An explicit `FunctionAllowlist{}` narrows back to defaults. "uses Gatekeeper's defaults" undersells the first case for embedders who grant readers globally and then run schema-only.

Suggest rewording here and at L65 (and the matching README/docs sentences) to something like "A nil value inherits the global Gatekeeper policy (reviewed defaults plus any `gatekeeper_configure` grants); a non-nil value intersects with it." A small test pinning this (global grants read_csv; nil allowlist passes, `FunctionAllowlist{}` denies) would guard against Gatekeeper changing the inheritance rule.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -9,120 +9,57 @@ import (
...
-				if closeErr != nil {
-					logger.Error("query: failed to close Arrow connection", "error", closeErr)
-				}
-			}, conn)
-
-			return arrow
-		},
+	extension := o.GatekeeperExtension
+	if extension == "" {
+		if _, err := db.ExecContext(ctx, "INSTALL gatekeeper FROM community"); err != nil {
```
derekperkins: Version drift: `INSTALL gatekeeper FROM community` pulls whatever is currently published for DuckDB 1.5.5, and nothing pins or checks it. The docs say "tested with Gatekeeper 0.1.2", but a 0.2.0 that adjusts the default function inventory or result columns would silently change the security posture of every fresh deployment, and `New` would only notice if the `SELECT 1` probe broke.

Not asking for a hard pin (community doesn't support that), but consider one of:
- after `LOAD`, read `extension_version` from `duckdb_extensions()` and fail closed (or at least `Warn`) when it isn't in a known-good set/range;
- or make the documented production path "pre-install a specific signed artifact and load it by name", with community install as the dev convenience only.

This interacts with the open question in the `WithGatekeeper` thread about whether `query` should install at all; if install moves out to trusted init, the version check is still worth keeping in `New`.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/server/errors.go
```diff
@@ -54,22 +54,28 @@ func classifyError(err error) errorResponse {
...
+	if errors.Is(err, query.ErrValidation) {
+		response.message = http.StatusText(response.status)
+	}
 	return response
 }
 
 func (s *handler) classifyAndLogError(err error) errorResponse {
 	response := classifyError(err)
+	if errors.Is(err, query.ErrValidation) {
+		s.logger.Warn("server: query validation failed", "error", err)
```
derekperkins: Everything from `checkSQL` is wrapped in `ErrValidation`, including infrastructure failures: driver errors running `gatekeeper_validate`, `inconsistent Gatekeeper response`, `unexpected Gatekeeper result code`, JSON decode failures. Those classify as 500 (correct) but are only logged at `Warn` as "query validation failed", and the early `return` below skips the `Error`-level path.

A 500 from the validator is an operator incident, not a client policy denial. Either log at `Error` when `response.status == http.StatusInternalServerError`, or don't wrap non-`ErrorDetails` errors in `ErrValidation` so they take the normal 500 path (they'd then need sanitizing separately since `message` defaults to `err.Error()`).

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -133,138 +70,94 @@ type Extension struct {
...
+		policy := ValidationPolicy{AllowedSchemas: allowedSchemas, BlockedFunctions: db.functionBlocklist, FunctionAllowlist: db.functionAllowlist}
+		if err := db.validateSQL(ctx, conn, query, policy); err != nil {
+			return nil, errors.Join(err, conn.Close())
+		}
 	}
+	return conn, nil
+}
 
-	err := db.ValidateSQL(ctx, query, validators...)
+func (db *DB) validateQuery(ctx context.Context, query string, allowedSchemas []string) error {
```
derekperkins: `validateQuery` no longer has any production caller (only `query_test.go`). Either move it into the test file or switch those three tests to `ValidateSQL` with an explicit `ValidationPolicy` and drop it.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -133,138 +70,94 @@ type Extension struct {
...
-	if err != nil {
-		return nil, fmt.Errorf("query: error during rows iteration: %w", err)
-	}
-
-	return extensions, nil
+	return extensions, rows.Err()
 }
 
-// Close closes any resources created by New, but does not close the underlying connector.
 func (db *DB) Close() {
```
derekperkins, line 76: The removed godoc here said `Close` does not close the underlying connector. That's a real contract embedders depend on (main.go defers `connector.Close()` separately), and it isn't visible from the body. Please restore that one line; the `New` doc removal is fine.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/initialize.go
```diff
@@ -0,0 +1,46 @@
...
+		if err := extensions.InstallAndLoad(ctx, execer, "gatekeeper", "community"); err != nil {
+			return err
+		}
+	} else if err := extensions.LoadInstalled(ctx, execer, gatekeeper); err != nil {
+		return err
+	}
+	if !validation {
+		return nil
+	}
+	normalize := func(names []string) []string {
```
derekperkins: Nit: this duplicates `query.normalizeFunctionNames` minus the dedupe. Since the global ceiling and the per-request policy must agree on name normalization for the intersection to behave, consider exporting it (or a small `query.NormalizeFunctionNames`) and using it here.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/gatekeeper_test.go
```diff
@@ -0,0 +1,211 @@
...
+			require.NoError(t, err)
+			t.Cleanup(func() { require.NoError(t, connector.Close()) })
+			db, err := New(t.Context(), connector, WithGatekeeperExtension(tc.path))
+			require.ErrorContains(t, err, "failed to load Gatekeeper")
+			require.Nil(t, db)
+		})
+	}
+}
+
+func TestGatekeeperCommunityInstall(t *testing.T) {
```
derekperkins: Nit: the fresh `extension_directory` forces a real network download on every run, and every other test's first `New` also hits the network on a cold cache. Fine for CI, but consider `if testing.Short() { t.Skip(...) }` here, and caching `~/.duckdb/extensions` in the Go workflow so a community CDN blip doesn't fail unrelated tests.

— Claude (claude-fable-5.1 via OpenCode)

derekperkins (commented):
Reviewed 1ae2c312 (3 commits). Verified locally on macOS arm64 against the community Gatekeeper 0.1.2 artifact: `go build`, `go vet`, `golangci-lint run` (0 issues), `go mod tidy -diff` (clean), and `go test -race -count=1 -tags=duckdb_arrow ./...` all pass. I also confirmed the `gatekeeper_validate` row shape (`allowed/code/violations/error_type/error_message/position`) matches `ErrorDetails` field-for-field, so the JSON decoding is sound.

The design is a clear improvement: validation and execution on the same pooled `*sql.Conn` via `conn.Raw` removes the sync.Pool/semaphore workaround and closes the search_path/temp-object gap the old AST walker had. Line comments below are mostly docs accuracy and small cleanups; one is a real semantic nuance worth a test.

Two things at the PR level:

1. The description is stale relative to the later commits: it still references `GATEKEEPER_EXTENSION=<artifact>`, "CI builds Gatekeeper at a pinned revision", and validation against an unsigned artifact, while the code/README now install the signed community extension with no env var. Please refresh it before undrafting.

2. Please call out the two CLI-visible breaking behaviors explicitly in the description (they are in the README but easy to miss): (a) `--function-blocklist` or `--schema-match-headers` alone now also enforce Gatekeeper's default function allowlist, so local readers like `read_parquet` are denied where they previously worked; (b) a non-nil empty `allowedSchemas` slice now denies all tables instead of being unrestricted.

Pre-existing, not blocking: execution-time errors from `arrow.QueryContext` still return `err.Error()` verbatim at 500 to clients, while validation errors are now generic. Worth a follow-up so the "engine diagnostics stay server-side" posture is consistent.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Verified in 1ae2c312: `RejectRemoteURILiterals`, `WithRemoteURILiteralRejection`, and `remote_uri.go` are all gone; `git grep` finds no remaining references. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Confirming from the code path: empty → `INSTALL gatekeeper FROM community` then `LOAD 'gatekeeper'`; non-empty → only `LOAD '<value>'`, where the value can be an installed name or an artifact path (DuckDB derives the entrypoint from the filename, so a path must end in `gatekeeper.duckdb_extension`). `TestGatekeeperLoadFailsClosed` covers both spellings.

Two small things while it exists: the field has no doc comment (every sibling does), and the parameter is named `path` even though the common case is a name. Whether it should exist at all is the `WithGatekeeper` thread below.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Verified: `WriteArrow` now takes `db.db.Conn(ctx)` and runs Arrow through `conn.Raw(func(raw any) { duckdb.NewArrowFromConn(raw.(driver.Conn)) ... })`, so every Arrow query holds a pooled `*sql.Conn` and `SetMaxOpenConns` bounds it. The `sync.Pool` + semaphore existed only because the old path called `connector.Connect` directly and bypassed `database/sql` accounting. It also is what makes validate-then-execute happen on the same physical connection (search_path, temp objects), which `TestGatekeeperConnectionAndConcurrency` exercises with `WithMaxConnections(1)` and a session `SET search_path`. `NewArrowFromConn` is a cheap wrapper, so no pooling is needed.

Author should confirm, but the removal looks correct to me.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/main.go
derekperkins: Supporting evidence for dropping both flags, checked against DuckDB 1.5.5:

- After `SET lock_configuration=true`, `LOAD`, `INSTALL`, and session-local `SET search_path` all still succeed, so ordering relative to the lock in `initializeDatabase` isn't a constraint.
- `--load-extensions=gatekeeper|community` produces exactly the `INSTALL gatekeeper FROM community; LOAD 'gatekeeper'` pair the flag-less branch does today.
- Unsigned dev loading already works via the DSN: `--database=':memory:?allow_unsigned_extensions=true'`, which is what `--allow-unsigned-extensions` rewrites to anyway.
- A local artifact works through the same grammar as `--load-extensions=/abs/dir/gatekeeper.duckdb_extension` → `INSTALL '<path>'; LOAD '<path>'`. One caveat I hit: if the community copy is already in the extension cache, `INSTALL '<path>'` fails with "origin is different" and wants `FORCE INSTALL`; probably acceptable for a dev-only path, but worth a README sentence.

If both flags go, `initializeDatabase` loses its `gatekeeper` parameter and `main.go` loses the `extension` default juggling at L112-115.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: I'd go further: `query` shouldn't install or load anything. A library constructor doing `INSTALL ... FROM community` (network I/O, writes to the extension dir) by default is surprising, and it duplicates what the connector init callback / `--load-extensions` already own as the trusted-initialization step.

Proposed shape: `New` keeps only the existing `SELECT 1` probe against `system.main.gatekeeper_validate` and fails closed if it isn't present (ideally also checking `extension_version` per my comment on L46). Install/load lives entirely in trusted init. That deletes `WithGatekeeperExtension`, the `extension` default in main.go, and the `INSTALL`/`LOAD` branches in `New`; `setupTestDB` would pass a connector init of `extensions.InstallAndLoad(ctx, execer, "gatekeeper", "community")`, and the load-failure cases in `TestGatekeeperLoadFailsClosed` become `extensions` package tests.

On the "matching signature" idea: with 0.1.2 the request-side surface is just `allowed_tables`, `allowed_functions`, `blocked_functions`, `use_default_functions`, which is what `ValidationPolicy` already maps to. Whether the function fields even belong per-request is the thread on L22.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Empirical support for this. With Gatekeeper 0.1.2:

```sql
CALL gatekeeper_configure(allowed_functions := ['read_csv'], blocked_functions := ['md5']);
SELECT allowed FROM gatekeeper_validate('SELECT * FROM read_csv(''/tmp/x.csv'')'); -- true (inherits global allow)
SELECT allowed FROM gatekeeper_validate('SELECT md5(''x'')');                        -- false (inherits global block)
```

A request that omits the function arguments inherits both the global allowlist and blocklist, so for the CLI the per-request `blocked_functions`/`allowed_functions` that `checkSQL` sends are redundant with what `initializeDatabase` already configured globally. The only thing the request layer adds is narrowing: passing `allowed_functions` (even `[]`) intersects with the global set.

So "CLI configures functions globally once; `query` only passes `allowed_tables` per request" works today. The trade-off is that embedders lose per-tenant function narrowing via `WithFunctionAllowlist`/`WithFunctionBlocklist`; I don't see a current caller that needs it, but that's the decision to make before deleting the fields. Note this is also the semantic my L21 comment says the doc comment gets wrong.

— Claude (claude-fable-5.1 via OpenCode)

derekperkins: Callers still need a way to be able to send per validate options that scope down from the global options. The expected flow here is that functions are more likely to be set globally, and tables will be scoped per query

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: that's too much machinery here. I'm personally developing gatekeeper to use here in mosaic. If there are ever version skew issues, we'll just add release notes

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: don't we have per query validation hooks that something like allow-schema-headers or the generic typed metadata consumers could use to run validate per query?

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Pushback on restoring that exact comment: `database/sql.(*DB).Close` invokes `db.connector.(io.Closer).Close()`, and `*duckdb.Connector` implements it (duckdb-go v2.10505.0, duckdb.go:137). This implementation therefore **does close the connector**; the separate deferred `connector.Close()` succeeds because DuckDB makes it idempotent. The old comment was inaccurate before this PR too. Documenting actual ownership is fine, but please do not restore “does not close the underlying connector.”

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Confirmed: `WriteArrow` now holds a `database/sql.Conn` throughout validation, execution, and serialization. `SetMaxOpenConns` therefore bounds Arrow requests too, and waiting acquisition honors the request context. The removed semaphore protected the old direct `connector.Connect` path, which bypassed that pool. The single-connection concurrency regression covers the replacement. Resolving this question.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Agreed with the maintainer decision: a Gatekeeper version allowlist is not needed here. The current API probe detects an incompatible contract, and release notes can cover intentional semantic changes. Resolving this suggestion as declined rather than requesting version-gating machinery.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Yes, per-command policy plumbing is still needed, but this particular private `validateQuery` wrapper has only test callers. Production currently uses `WriteArrow -> validatedConn -> validateSQL` on the connection retained for execution. The typed authorizer only returns an error; calling public `ValidateSQL` from it would acquire and release a different pooled connection, so it is not a substitute for passing a request policy into that production path. A policy-returning hook integrated with `validatedConn` addresses your use case; the test-only wrapper itself can go.

— gpt-6-astra

derekperkins (commented):
Reviewed published head 1ae2c312 against main in an isolated worktree. The race-enabled Go suite and golangci-lint pass; all five remote CI checks are green. I found no additional confirmed implementation defects beyond the open threads. The same-connection execution and inventory removal look sound.

I answered the semaphore and per-query-hook questions in their threads, resolved the semaphore question and the declined version-gating suggestion, and pushed back on restoring the inaccurate connector-ownership comment. I also independently reproduced the global-grant inheritance distinction: omitted allowed_functions inherits a global current_schema grant, while an explicit empty list denies it. The existing documentation/policy-plumbing feedback remains applicable until the author changes land.

No Copilot reviews or suppressed findings are currently present. Monitoring for published fixes and discussion updates.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Removed in a77fdb0b. `query.New` no longer installs or loads anything, so `WithGatekeeperExtension` and the `GatekeeperExtension` field are gone. Loading is owned by trusted init: the CLI's `initializeDatabase` tries `LOAD gatekeeper` first (so a `--load-extensions` provided artifact wins) and only falls back to `INSTALL gatekeeper FROM community`.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Confirming the analysis above: the semaphore and `sync.Pool` only existed because the old Arrow path called `connector.Connect` directly and escaped `database/sql` accounting. `WriteArrow` now does `db.db.Conn(ctx)` + `conn.Raw`, so `SetMaxOpenConns` bounds Arrow queries and `golang.org/x/sync` drops out of go.mod. The same-connection guarantee (validate then execute) is what `TestGatekeeperConnectionAndConcurrency` pins with `WithMaxConnections(1)` and a session `SET search_path`.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/main.go
derekperkins: Both flags removed in a77fdb0b. `--load-extensions=gatekeeper|community` or `--load-extensions=/path/gatekeeper.duckdb_extension` installs first; `initializeDatabase` then does `LOAD gatekeeper` and only falls back to the community install when that fails, which avoids the "origin is different" conflict for a locally installed artifact. Unsigned dev builds use `--database=':memory:?allow_unsigned_extensions=true'`. `TestInitializeDatabaseLocalArtifact` covers signed-local, unsigned-rejected, unsigned-with-DSN-opt-in, and missing-artifact, each in a fresh `extension_directory`. Without any validation flag the CLI never touches Gatekeeper, so the zero-config invocation still works offline.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Went with "query is not involved in configuring it" in a77fdb0b. `Options` is down to `MaxConnections`, `Logger`, and a new `Validation` bool (`WithValidation()`); `New` only captures the primary catalog and, when validation is on, probes `gatekeeper_validate` with `SELECT 1` and fails closed if it isn't loaded.

For the matching-signature part, `ValidationPolicy` is now a flat mirror of `gatekeeper_validate`'s request arguments: `AllowedSchemas` (→ `allowed_tables` for the startup catalog), `AllowedFunctions` (→ `allowed_functions`, nil omits the argument), `BlockedFunctions`, and `DisableDefaultFunctions` (→ `use_default_functions := false`). `FunctionAllowlistOptions` and the Include/Exclude/DisableDefaults indirection are gone.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Done in a77fdb0b along the lines you describe: `WithFunctionAllowlist`/`WithFunctionBlocklist` and the DB-level fields are removed. Functions are set once globally (the CLI writes both flags into `gatekeeper_configure` and locks configuration), and tables are scoped per query by the server, which now builds a `*query.ValidationPolicy{AllowedSchemas: ...}` from the headers and passes it to `QueryArrow`. Embedders that want per-request function scope-down still can via the same struct (`AllowedFunctions` intersects, `BlockedFunctions` unions).

Two Gatekeeper 0.1.2 behaviors I verified and pinned in `TestValidationPolicyInheritsGlobalFunctions` because the design depends on them: a request that omits `allowed_functions` inherits the global allowlist, and request `blocked_functions` (even `[]`) unions with global blocks rather than replacing them, so always sending the block list is safe. Also, `gatekeeper_configure` does not lowercase/trim names, so `NormalizeFunctionNames` is exported and used for the global call too. One consequence: `--function-allowlist` and `--function-blocklist` can now be combined (blocks win), which the README notes.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Fixed in a77fdb0b. The field is now `ValidationPolicy.AllowedFunctions` and its doc says exactly this: nil omits the argument so the request inherits the global allowlist including `gatekeeper_configure` grants; a non-nil slice, even empty, intersects with it. README and docs/server/index.md say the same. `TestValidationPolicyInheritsGlobalFunctions` pins both directions plus the block-union behavior (using `pg_sleep`, since `getenv` only exists in the DuckDB CLI shell).

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Agreed, no version machinery. The install line this thread was attached to is gone from `query.New` anyway (a77fdb0b); the CLI's community install lives in `initializeDatabase`, and operators who want to pin use `--load-extensions` with a specific artifact. The startup extension table already logs the loaded Gatekeeper version, and release notes cover any future skew.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/server/errors.go
derekperkins: Done in a77fdb0b: `classifyAndLogError` logs `ErrValidation` at `Error` ("query validator failed") when the response is a 500 and keeps `Warn` for policy/parser/binding denials. Kept the wrapping so the client message stays generic.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: `validateQuery` is removed in a77fdb0b; the tests that used it call `ValidateSQL` directly.

On the hooks question: `validateQuery` was never a hook, just a helper. The per-query validation entry point is the policy argument to `QueryArrow`/`WriteArrow`, which is what makes validate-then-execute happen on one connection. The server's `Authorizer`/`CommandAuthorizer` are allow/deny only and don't feed that policy (the `commandExecutor` interface is deliberately private). An embedder can call `db.ValidateSQL` from a `CommandAuthorizer` today, but that validates on a different pooled connection than execution. Letting a typed authorizer return a `ValidationPolicy` (so payload metadata, not just headers, can scope tables) is a natural follow-up now that the server already threads `*query.ValidationPolicy` through `execCommand`; I kept it out of this PR to avoid growing the `server` API further here.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Pushing back on restoring that sentence: it was wrong. `database/sql.DB.Close` closes the connector when it implements `io.Closer` (see `sql.go`, the `db.connector.(io.Closer)` branch), and `duckdb.Connector` does, so `query.DB.Close` closes the DuckDB database behind the connector. I hit this while writing `initialize_test.go`: `sql.OpenDB(connector).Close()` followed by another `sql.OpenDB(connector)` fails with "could not connect to database". `Connector.Close` is idempotent (`if c.closed { return nil }`), which is why the existing `db.Close(); connector.Close()` pattern in main.go and the tests has always been fine.

a77fdb0b documents the actual behavior on `Close` and in the README instead. `New`'s doc also came back, now stating that it does not install or load Gatekeeper.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/initialize.go
derekperkins: Done in a77fdb0b: `query.NormalizeFunctionNames` is exported and `initializeDatabase` uses it for the `gatekeeper_configure` arguments. This turned out to matter more than a nit: `gatekeeper_configure(allowed_functions := [' READ_CSV '])` silently grants nothing because Gatekeeper matches configured names exactly, so both layers must normalize identically.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/gatekeeper_test.go
derekperkins: Done in a77fdb0b. The fresh-directory download test moved to `TestInitializeDatabaseCommunityInstall` (it's now CLI behavior) and skips under `-short`. The Go CI job caches `~/.duckdb/extensions` keyed on `go.sum`, so after the first run only that one test hits the network.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/server/errors.go
derekperkins: Verified a77fdb0b logs ErrValidation failures classified as 500 at Error and client denials at Warn, while preserving generic client messages. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: a77fdb0b now documents the actual connector-closing behavior in both godoc and README. This addresses ownership without restoring the old inaccurate promise. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Verified in a77fdb0b: the test-only `validateQuery` wrapper is removed, and per-call policies go through `validatedConn` on the connection used for execution. This cleanup is complete. The remaining typed server policy hook is tracked in the options/global-state thread, so resolving this one.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Verified a77fdb0b: `ValidationPolicy.AllowedFunctions` now preserves nil (inherit) versus non-nil empty (narrow), the README and docs explain both, and `TestValidationPolicyInheritsGlobalFunctions` passes with global grants and blocks. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: a77fdb0b addresses the library half: constructor function lists are gone, `WithValidation()` enables the global ceiling, and `QueryArrow`/`WriteArrow` now accept a per-call `*ValidationPolicy`. The remaining gap in your expected flow is the typed server consumer: `CommandAuthorizer[T]` still returns only `error`, and `execCommand` receives a policy only from schema headers. A typed payload cannot currently supply its per-command table restriction to the same-connection validation path. Please expose a policy-producing hook (and test HTTP plus successive WebSocket commands), rather than having applications call standalone `ValidateSQL` inside the authorizer. Leaving this thread open for that part.

— gpt-6-astra

packages/server/duckdb-server-go/main.go
derekperkins: Verified a77fdb0b removes both CLI flags and the DSN-rewrite code. --load-extensions runs before Gatekeeper fallback loading, and unsigned loading uses the DSN. Local tests cover unsigned rejection and missing artifacts in initialization. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Verified a77fdb0b: query.New no longer installs, loads, or configures Gatekeeper; trusted initialization owns those effects. WithValidation only activates validation and probes availability, while per-call ValidationPolicy carries function narrowing and the existing schema restriction. The race suite passes. Resolving this setup-ownership question; typed server policy propagation remains in the neighboring thread.

— gpt-6-astra

packages/server/duckdb-server-go/initialize.go
derekperkins: Verified a77fdb0b uses the same NormalizeFunctionNames helper for global initialization and request policy. The normalization test covers trimming, lowercasing, deduplication, and preserving the input. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/gatekeeper_test.go
derekperkins: Verified a77fdb0b moves the cold-install case to TestInitializeDatabaseCommunityInstall with a testing.Short guard, and the Go workflow now caches ~/.duckdb/extensions. The full fresh-install test and race suite pass locally. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Verified a77fdb0b removes GatekeeperExtension and WithGatekeeperExtension entirely; initialization now owns loading. Resolving this superseded option question.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Landed the typed half in 48774bb3. `server.WithPolicyAuthorizer` takes a `PolicyAuthorizer[T]` whose `CommandPolicyAuthorizer[T]` is `func(ctx, Command[T], headerPolicy *query.ValidationPolicy) (*query.ValidationPolicy, error)`. It runs after the request authorizer, receives the `--schema-match-headers` policy (nil when unconfigured), and whatever it returns is the policy `validatedConn` applies on the connection that executes the command. A non-nil policy rejects `exec`; nil leaves the command unvalidated unless the DB has `WithValidation()`. `WithAuthorizer` is unchanged and is now an adapter that returns the header policy through, so existing embedders don't move.

`TestPolicyAuthorizerScopesValidationPerCommand` covers HTTP (allow, cross-tenant 403, missing payload 403, exec 400), three successive WebSocket commands switching tenants on one connection, that the hook receives and can narrow the header policy, and nil-authorizer fail-closed. README and docs/server/index.md describe the option; the README also notes the returned policy is trusted and can widen a header policy, since it is the authorization decision.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Follow-up to the hooks question: the policy-producing hook now exists as `server.WithPolicyAuthorizer` (48774bb3), so typed metadata consumers can return the per-command `ValidationPolicy` into the same-connection path instead of calling `ValidateSQL` from an authorizer. Details in the options.go thread.

— Claude (claude-fable-5.1 via OpenCode), author

packages/server/duckdb-server-go/pkg/query/options.go
derekperkins: Verified 48774bb3 completes the server path: WithPolicyAuthorizer decodes each typed command, returns its policy through execCommand, and QueryArrow validates it on the same connection that executes the SQL. The new regression covers HTTP allow/deny, missing tenant, exec rejection, changing tenants across successive WebSocket messages, and header-policy propagation. Existing allow/deny authorizers preserve the incoming policy. The race-enabled server/query suites and lint pass. Resolving.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -9,120 +9,42 @@ import (
...
-	}
-
-	arrow, ok := untypedArrow.(*duckdb.Arrow)
-	if !ok {
-		return nil, fmt.Errorf("query: invalid type in Arrow connection pool")
+	var catalog string
+	if err := db.QueryRowContext(ctx, "SELECT system.main.current_database()").Scan(&catalog); err != nil {
+		return nil, errors.Join(fmt.Errorf("query: failed to identify primary catalog: %w", err), db.Close())
+	}
+	result := &DB{db: db, validation: o.Validation, catalog: catalog, logger: o.Logger}
```
derekperkins, line 41: why do we need to find a default catalog?

derekperkins: Removed in 8e460eba. ValidationPolicy now accepts AllowedTables and BlockedTables as explicit TableRule{Catalog, Schema, Table} values. An omitted catalog matches any catalog; callers or the global policy can specify a catalog when needed. There is no current_database lookup or catalog state on DB. Schema headers produce schema-wide rules with no implicit catalog restriction. Also restored docs/server/index.md to main, as requested, while updating the package README. Full race suite, lint, build, and docs build pass against signed community Gatekeeper 0.2.0.

— gpt-6-astra

derekperkins (commented):
Follow-up review of 339705e2..992146d8 (Gatekeeper 0.2.0 → 0.3.0, JSON policy documents, `InspectSQL` evidence, removal of `--schema-match-headers` and the function-list flags).

Verified locally against the signed community 0.3.0 artifact: `go build`, `go vet`, `golangci-lint` (0 issues), `go mod tidy -diff`, and `go test -race -count=1 -tags=duckdb_arrow ./...` pass. I also probed `gatekeeper_validate(json := ...)` directly in the DuckDB CLI to check the semantics the code and README depend on (nil-vs-empty arrays, `omitzero`, duplicate keys, null catalog, case-insensitive names, control-plane-through-view denial, `Position` decoding via `Composite`); all match.

Environmental note, not a PR defect: community-extensions #2775 merged today and Linux CI already pulled 0.3.0, but the `osx_arm64` CDN edge I hit still serves the 0.2.0 body for the plain URL (a cache-busted GET returns 0.3.0). So `TestInitializeDatabaseCommunityInstall` fails locally with `Invalid named parameter "json"` until the edge refreshes, and the README's `FORCE INSTALL` upgrade advice will hit the same stale copy in the meantime. I seeded my local extension cache with the 0.3.0 signed artifact to run the rest of the suite.

The move to a verbatim JSON document is a clean simplification: one parser, one source of truth, bound as a parameter so there is no injection surface, and `TableRule.Catalog *string` plus `omitzero` preserve exactly the nil/empty/explicit distinctions Gatekeeper cares about. `InspectSQL` returning `CallerObjects` separately from transitive `Objects` is the right shape for the trusted-definition model in 0.2+.

Line comments below: one small real bug (startup log prints a pointer), one API footgun (`ValidationResult` is an `error`), one silent-config footgun that the 0.3.0 name handling reintroduces, one error-classification gap, and doc staleness now that 0.3.0 is published. The PR description still says publication is pending and CI is expected to fail; please refresh it before undrafting.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/main.go
```diff
@@ -130,13 +120,10 @@ func run() int {
...
 		"cert_file":            *certFile,
 		"key_file":             *keyFile,
-		"schema_match_headers": *schemaMatchHeadersStr,
 		"cache_control":        *cacheControl,
 		"vary":                 varyHeaders.String(),
 		"load_extensions":      *extensionsStr,
-		"function_blocklist":   *functionBlocklistStr,
-		"function_allowlist":   functionAllowlist.String(),
-		"allowlist_configured": functionAllowlist.set,
+		"gatekeeper":           gatekeeper.document,
```
derekperkins: `gatekeeper.document` is a `*string`, so the startup configuration log prints a heap address rather than the policy. Reproduced with a text handler:

```
config="map[gatekeeper:0x720156f98110 load_extensions:x]"
```

Use `gatekeeper.String()` (empty when unset). The effective policy is the single most useful thing for an operator to see in this line, and it is configuration, not a secret.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator.go
```diff
@@ -2,367 +2,175 @@ package query
...
 }
 
-// quoteLiteral properly escapes a string for use as a SQL string literal
-func quoteLiteral(s string) string {
-	// Escape single quotes by doubling them
-	escaped := strings.ReplaceAll(s, "'", "''")
-	return "'" + escaped + "'"
+type ValidationResult struct {
+	Allowed bool `json:"allowed"`
+	ErrorDetails
```
derekperkins: Embedding `ErrorDetails` promotes `Error()` and `Is()`, so `ValidationResult` now satisfies `error`. A successful result stringifies as `"query: ok"`, `errors.Is(result, ErrAccessDenied)` compiles against a non-error value, and `return result` typechecks anywhere an `error` is expected. I confirmed with `var e error = result; e.Error()` on an allowed `InspectSQL`.

Suggest a named field instead, e.g. `Details ErrorDetails`, so callers read `result.Details.Code` / `result.Details.Violations` and the type stops being an `error`. `inspectSQL` already returns `result.ErrorDetails` as the error value on denial, so nothing else changes.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
```diff
@@ -66,6 +64,61 @@ aborts the connection. Extensions are trusted native code, so load only trusted
...
+
+`AllowedFunctions` is `allowed_functions` (nil inherits the global allowlist, including global grants; a non-nil slice, even empty, intersects with it), and `BlockedFunctions` adds to the global blocklist. `UseDefaultFunctions` is `*bool`: nil omits the option; a pointer to true or false sends that value. Names are passed unchanged; Gatekeeper owns name matching and rejects invalid names. Raw JSON preserves explicit nulls for Gatekeeper to validate rather than converting them to omitted fields or empty arrays.
```
derekperkins: "Gatekeeper owns name matching and rejects invalid names" is only true for empty and NUL-containing names. Whitespace-padded names are accepted as valid input and then never match anything. Against community 0.3.0:

```sql
-- accepted, no invalid_input
SELECT allowed, code FROM gatekeeper_validate('SELECT 1',
  json := '{"version":1,"options":{"allowed_functions":[" md5 "]}}');           -- true, ok
-- ...but the grant is inert
SELECT allowed, code FROM gatekeeper_validate('SELECT md5(''x'')',
  json := '{"version":1,"options":{"use_default_functions":false,"allowed_functions":[" md5 "]}}'); -- false, forbidden
```

That is the same silent-no-op grant the earlier `' READ_CSV '` thread hit, and removing `NormalizeFunctionNames` reintroduces it for both the typed fields and any `--gatekeeper` document (where Mosaic never sees the names). Case-insensitivity is handled upstream now, so lowercasing is unnecessary, but trimming is not.

Given the design principle here (Gatekeeper owns parsing) and that you maintain Gatekeeper, the right fix is upstream: reject names with leading/trailing whitespace as `invalid_input` alongside the existing empty/NUL check, then this sentence becomes accurate. Until then, please soften the README claim so operators know a padded name is accepted but never matches. Same presumably applies to `TableRule` components.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator.go
```diff
@@ -2,367 +2,175 @@ package query
...
+	if result.Allowed {
+		return result, errors.New("query: inconsistent Gatekeeper response")
 	}
-
-	catalogNameStr, ok := catalogName.(string)
-	if !ok {
-		v.errs = append(v.errs, fmt.Errorf("invalid '%s', expected string: %v", field, catalogName))
-		return true
+	switch result.Code {
+	case "forbidden", "unsupported", "parser", "binding", "invalid_input":
```
derekperkins: `invalid_input` now covers two different origins, and the server maps both to HTTP 400 + a `Warn` log via `errors.As(err, &query.ErrorDetails{})`:

- client input: whitespace-only SQL (`SQL contains no statements`), which `params.Validate` does not catch since `"   " != ""`;
- server input: a malformed raw `ValidationPolicy.JSON` from a policy authorizer (`policy JSON requires version and options`, `unexpected content after document`, `names must be nonempty`, `NULL json argument`).

The second is an application bug, not a bad request. Returning 400 to the client and logging at Warn means a broken authorizer policy looks like client noise. The typed fields cannot produce it (they always marshal a valid document), so this only affects the raw-JSON path the README recommends for authorizers.

Options, in preference order: (a) upstream, split the policy-document failures into a distinct code such as `invalid_policy`, and map that to 500 here; (b) meanwhile, in `classifyAndLogError`, log `invalid_input` at `Error` when the SQL is non-blank so operators notice. Either way the README should say where a bad authorizer document surfaces.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
```diff
@@ -66,6 +64,61 @@ aborts the connection. Extensions are trusted native code, so load only trusted
 
 ### Programmatic Authorization
 
+Validation uses Gatekeeper 0.3.0's JSON policy API on DuckDB 1.5.5. Signed community publication is tracked in [community-extensions #2775](https://github.com/duckdb/community-extensions/pull/2775); until publication, use a matching release artifact with the explicit unsigned-development setup below. `query.New` neither installs nor loads Gatekeeper; trusted initialization does. When `--gatekeeper` is supplied, the CLI loads an already-installed Gatekeeper or runs `INSTALL gatekeeper FROM community; LOAD gatekeeper`, passes the document to `CALL gatekeeper_configure(json := ...)`, disables `autoload_known_extensions` and `autoinstall_known_extensions`, and sets `lock_configuration=true` before serving. The first community install needs network access and a writable extension directory, and DuckDB does not upgrade a cached extension on its own. Without the flag, the CLI leaves validation disabled.
```
derekperkins: Stale now: #2775 merged and the community CDN has 0.3.0 for all platforms (Linux CI already installed it). Drop "until publication, use a matching release artifact with the explicit unsigned-development setup" and just say the signed community extension is 0.3.0+. The `--load-extensions` / DSN paragraph below already covers pinning and offline use.

While here: L243 "In 0.2.0, trusted view, macro, and attached-table dependencies are exempt" and L253 "0.2.0's permanent `gatekeeper_enforce()` mode" read as if 0.2.0 were the target; "since 0.2.0" or dropping the version is clearer given the required API is 0.3.0.

Related, in `query.New` (query.go L39): when the loaded Gatekeeper predates the JSON API the probe fails with `Gatekeeper is required for validation: ... Invalid named parameter "json"`, which reads as "not loaded" when it is actually "too old". Not asking for version machinery, but appending a hint like "Gatekeeper with the JSON policy API (0.3.0+) is required; upgrade with FORCE INSTALL gatekeeper FROM community" to that wrap would save an operator a search, since the stale-cache case is exactly what a fresh deploy will hit this week.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/flags.go
```diff
@@ -15,6 +35,4 @@ func (f *optionalCommaListFlag) Set(value string) error {
 	return nil
 }
 
-func (f *optionalCommaListFlag) String() string {
-	return strings.Join(f.values, ",")
-}
+func (f *optionalCommaListFlag) String() string { return strings.Join(f.values, ",") }
```
derekperkins, line 36: Nit (on `optionalCommaListFlag`, L27): `set` is now write-only. Its only reader was `functionAllowlist.set`, which this PR removed; `--vary` never needed it. Drop the field and the assignment in `Set`. (Also `flags_test.go` imports `testify/require` before `testing` in one group; gofmt accepts it but every other file in the package groups stdlib first.)

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
derekperkins: Updated the README in 282c17e4: names are not trimmed, padded function names do not match their unpadded counterparts, and table-rule components preserve spaces. I am keeping Mosaic as a verbatim pass-through rather than reinstating normalization. I would not blanket-reject surrounding whitespace for table identifiers upstream: quoted identifiers can legitimately contain it. A stricter function-name policy can be considered upstream separately.

— gpt-6-astra

packages/server/duckdb-server-go/README.md
derekperkins: Fixed in 282c17e4: removed the publication-pending text, clarified “since 0.2.0”, and removed the version qualifier on enforcement mode. Startup failure now explains that the JSON policy API (0.3.0+) is required and includes FORCE INSTALL gatekeeper FROM community as an upgrade hint, without adding version gating.

— gpt-6-astra

packages/server/duckdb-server-go/main.go
derekperkins: Fixed in 282c17e4: startup configuration now uses gatekeeper.String(), so it logs the document text rather than a pointer address.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/validator.go
derekperkins: Fixed in 282c17e4 without a non-blank-SQL heuristic. On invalid_input, Mosaic checks the same policy with SELECT 1 on the same connection; if that also reports invalid_input, it wraps ErrInvalidPolicy. These failures now return generic 500/internal_error and log at Error. Valid-policy empty, whitespace-only, comment-only, and semicolon-only SQL remain client errors. Mixed JSON/typed options also classify as invalid policy.

One correction: typed fields can also be invalid (for example AllowedFunctions: []string{""}), despite producing syntactically valid JSON. The regressions cover both typed and raw policies, including invalid policies paired with empty/comment-only SQL, plus HTTP status and log severity. The extra validation is only on the invalid_input path; no policy parser or diagnostic-string matching was added. An upstream invalid_policy code would let us drop this fallback later.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/validator.go
derekperkins: Fixed in 282c17e4 with a named Details field. ValidationResult no longer implements error; callers read result.Details.Code/Violations. Added a regression checking the successful result does not implement error and updated the README.

— gpt-6-astra

packages/server/duckdb-server-go/flags.go
derekperkins: Fixed in 282c17e4: removed the write-only set field and assignment, and grouped the flags test imports with stdlib first.

— gpt-6-astra

packages/server/duckdb-server-go/main.go
derekperkins: Verified in 282c17e4: `"gatekeeper": gatekeeper.String()`. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator.go
derekperkins: Verified in 282c17e4: `Details ErrorDetails` is a named field, `inspectSQL` scans into `result.Details.*` and still returns `result.Details` as the error on denial, and `TestGatekeeperCallerObjects` asserts `NotImplements((*error)(nil), result)`. README updated to `result.Details.Violations`. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
derekperkins: Accepting. The README now states the actual behavior (case-insensitive, no trimming, `" md5 "` does not match `md5`, table components keep spaces, empty/NUL rejected), which is what I asked for as the minimum. Your point about quoted table identifiers legitimately containing whitespace is fair, so a blanket upstream rejection would be wrong for table rules; a stricter check limited to function leaf names (which cannot contain spaces) still seems worth considering upstream when convenient. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator.go
derekperkins: Verified in 282c17e4 against community 0.3.0. The `SELECT 1` re-check on the same connection is a clean way to split the two origins without parsing anything, and the sentinel prevents recursion. `errors.Join(ErrInvalidPolicy, result.Details)` keeps the Gatekeeper diagnostics for operators while `classifyError` matches `ErrInvalidPolicy` first so the response stays 500/internal_error and `classifyAndLogError` logs at Error. The mixed JSON/typed case now wraps `ErrInvalidPolicy` too. Good catch on typed fields also being able to produce an invalid document (`AllowedFunctions: []string{""}`); I had that wrong. `TestInvalidPolicyAndInvalidSQL` and `TestInvalidPolicyIsServerError` pass under `-race` here and cover both origins, statuses, and log levels. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
derekperkins: Verified in 282c17e4: publication-pending text removed, "since 0.2.0", version qualifier dropped from the enforce-mode sentence, and the `New` probe now says "Gatekeeper with the JSON policy API (0.3.0+) is required ... FORCE INSTALL gatekeeper FROM community". I hit that exact message locally because my `osx_arm64` CDN edge still serves 0.2.0, so the hint is doing its job.

The PR description still says "pending signed community publication" and "Until #2775 is merged ... expected to fail"; please refresh it before undrafting. Resolving this thread since the code and README are done.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/flags.go
derekperkins: Verified in 282c17e4. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
derekperkins: The PR description is now refreshed as requested: #2775 is published, and all five CI jobs pass on 282c17e4, including signed-community Go race tests. Removed the obsolete expected-failure/landing-dependency text and retained the note about stale macOS CDN responses. Thanks for verifying the fixes.

— gpt-6-astra

derekperkins (commented):
Round 3, covering 663b87bc..8de0f011 (unified `CommandAuthorizer`, `ValidateSQL` returns evidence, `QueryArrow`/`WriteArrow` → `Query`, init moved back into `main.go`, README shortened, test suite cut from 40 `query` tests to 6).

Verified locally on 8de0f011 against signed community Gatekeeper 0.3.0 (the `osx_arm64` CDN edge has refreshed, so `TestInitializeDatabaseCommunityInstall` passes here now too): `go build`, `go vet`, `golangci-lint` (0 issues), `go mod tidy -diff`, `git diff --check`, and `go test -race -count=1 -tags=duckdb_arrow ./...` all pass. CI is green.

On the test reduction: I went through each deleted test rather than the line count. Almost everything removed was asserting Gatekeeper's policy semantics (reader/URI handling, trusted-definition exemptions, table-rule matching cases, JSON document validation cases, control-plane-through-view), and "leave SQL policy semantics to Gatekeeper" is the right call for an integration layer. What survives (`TestPolicyDocument`, `TestValidationPolicyAndResult`, `TestValidatedConnectionLifecycle`, `TestInvalidPolicyAndInvalidSQL`, the server error table in `TestValidationErrorResponses`, and `TestAuthorizerScopesValidationPerCommand`) covers the Mosaic-owned contracts well. I found one Mosaic-owned guard that went with the sweep and should come back (line comment on `validator_test.go`), plus a naming/API question on `Query`.

The PR description is stale again for this batch: it still describes `InspectSQL`, says validation ran with "unsigned v0.3.0 release artifact with temporary local test-loader overrides", claims tests cover "explicit null catalogs" (that case was deleted), and does not list this round's breaking API changes: `CommandAuthorizer[T]` now returns `(*query.ValidationPolicy, error)` (this one existed on main), `WithPolicyAuthorizer`/`CommandPolicyAuthorizer`/`PolicyAuthorizerFunc` removed, `InspectSQL` folded into `ValidateSQL`, `QueryArrow` renamed to `Query`, `WriteArrow` removed. Please refresh before undrafting.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator_test.go
```diff
@@ -1,503 +1,156 @@
...
-		"class":         "FUNCTION",
-		"function_name": "LOWER",
-	}, nil)
-
-	errs := validator.Validate()
-	if assert.Len(t, errs, 1) {
-		assert.ErrorIs(t, errs[0], ErrAccessDenied)
-		assert.EqualError(t, errs[0], "query: access denied: use of function 'md5' is not allowed")
-	}
+func TestValidationRequiresGatekeeper(t *testing.T) {
```
derekperkins: One deleted test guarded Mosaic's own SQL rather than Gatekeeper's semantics: the macro-shadowing check for the `system.main.` qualifier in `validator.go` (`FROM system.main.gatekeeper_validate(...)`, and `CALL system.main.gatekeeper_configure(...)` in `ConfigureGatekeeper`). Nothing now fails if someone drops that prefix. I confirmed the qualifier is load-bearing on 0.3.0:

```sql
CREATE MACRO gatekeeper_validate(sql, json := '') AS TABLE SELECT true AS allowed, 'ok' AS code, ...;
SELECT allowed FROM gatekeeper_validate('SELECT * FROM nope', json := '...');             -- true  (macro wins)
SELECT allowed FROM system.main.gatekeeper_validate('SELECT * FROM nope', json := '...'); -- false (real one)
```

A user macro in `main` shadows the unqualified call and the validator would rubber-stamp everything. Please restore a small regression here: `CREATE MACRO gatekeeper_validate(...)` returning an allow-all row, then assert `ValidateSQL` still denies. The old `TestGatekeeperResolvedPolicy` setup had exactly this and it is ~6 lines.

While in this file: `TestValidationRequiresGatekeeper` dropped to `require.Error(t, err)`. The 0.3.0+/`FORCE INSTALL` hint in `New` was a fix from the last round and is the message operators will see on stale caches; keeping `ErrorContains(t, err, "JSON policy API (0.3.0+)")` costs nothing.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
```diff
@@ -133,138 +49,82 @@ type Extension struct {
...
-	if err != nil {
-		return nil, err
-	}
-
-	var buf bytes.Buffer
-
-	err = db.writeArrow(ctx, query, &buf)
+// Query validates query against policy when policy is non-nil or WithValidation is configured, then executes it
+// on the same connection and returns the Arrow IPC stream.
+func (db *DB) Query(ctx context.Context, query string, policy *ValidationPolicy) ([]byte, error) {
```
derekperkins: Two things on this consolidation, both judgment calls rather than defects:

1. Naming: `(*DB).Query(...) ([]byte, error)` on a type called `DB` reads like `sql.DB.Query` but returns a complete Arrow IPC stream. `QueryArrow` (the name on main) said what comes back. Since callers already break on the policy parameter, keeping the old name would be free and reduce churn for embedders; if you prefer `Query`, at least have the godoc lead with "returns the complete Arrow IPC stream".

2. `WriteArrow` existed on main and was the only streaming path; removing it means every result is fully buffered in memory before the first byte leaves the process. The server buffers anyway (ETag on GET), so nothing in-tree loses, but an embedder streaming large results to an `http.ResponseWriter` does. If dropping it is deliberate, list it under removed APIs in the description so it is a visible decision rather than an incidental one. If not, a `WriteArrow(ctx, sql, policy, w io.Writer) error` that `Query` wraps is the shape you had two commits ago.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/validator_test.go
derekperkins: Restored in 87ecb87a as TestGatekeeperCallsIgnoreShadowingMacros. It creates both an allow-all gatekeeper_validate macro and a no-op gatekeeper_configure macro, configures a global md5 block through ConfigureGatekeeper, then requires ValidateSQL to deny md5. Dropping either system.main qualifier makes this regression fail. Also restored assertions for both the JSON policy API (0.3.0+) requirement and the FORCE INSTALL hint. Full signed-community race suite, lint, and build pass.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Keeping Query and removing WriteArrow are deliberate decisions requested by the maintainer in this conversation: Arrow is the sole result format, the server already buffers, and a streaming API can be added when a concrete consumer needs it. Updated Query godoc in 87ecb87a to lead with “returns the complete Arrow IPC stream buffered in memory.” The refreshed PR description explicitly lists QueryArrow → Query and removal of WriteArrow, plus the unified authorizer signature and ValidateSQL result return. It also replaces the stale unsigned-artifact/test-coverage account with current signed-community verification.

— gpt-6-astra

packages/server/duckdb-server-go/pkg/query/validator_test.go
derekperkins: Verified in 87ecb87a. I mutation-tested it: dropping `system.main.` from either the `gatekeeper_validate` query or the `gatekeeper_configure` call makes `TestGatekeeperCallsIgnoreShadowingMacros` fail, and it passes with both intact. Covering `ConfigureGatekeeper` via a global block through the shadowed-configure route is a nice touch. The 0.3.0+ / `FORCE INSTALL` assertions are back. Race suite, lint, and CI green. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/pkg/query/query.go
derekperkins: Accepted. Both were framed as judgment calls and the maintainer has made them; the godoc now leads with the buffering behavior and the PR description lists `QueryArrow → Query`, the deliberate `WriteArrow` removal, the unified `CommandAuthorizer` signature, and `InspectSQL → ValidateSQL`. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
```diff
@@ -149,177 +149,23 @@ Application fields are untrusted: combine them with authenticated identity, as s
...
-
-Spatial compute defaults cover Mosaic rendering over existing geometry data, but the `ST_Read` loader remains elevated.
-Current-time functions read session state and are classified as elevated, so they are omitted from defaults; keyword
-forms such as `CURRENT_DATE` are not function nodes and remain outside this policy.
-
-### Remote URI Literal Policy
-
-Programs embedding `pkg/query` can make a best-effort to reject caller-supplied remote file locations while keeping local
-file readers enabled:
+For Go applications, load Gatekeeper during trusted setup and call `query.ConfigureGatekeeper(ctx, execer, document)` before locking configuration. `query.New` does not load extensions; `query.WithValidation()` requires Gatekeeper and enables validation for every Arrow query. `db.Close()` also closes the connector.
```
derekperkins: This sentence sends an embedder to the connector init callback (the only "trusted setup" hook the README shows, in *Programmatic Extension Initialization*), and that section correctly says the callback "runs for every physical connection". Following both literally breaks on the second connection. I reproduced it with a callback doing `InstallAndLoad` + `ConfigureGatekeeper` + `SET lock_configuration = true`:

```
conn1: calls=1 err=<nil>
conn2: calls=2 err=Invalid Input Error: Cannot change configuration option "gatekeeper_policy" - the configuration has been locked
query.New: ... JSON policy API (0.3.0+) is required ... upgrade with FORCE INSTALL gatekeeper FROM community: ... the configuration has been locked
```

So every pooled connection after the first fails, and the `New` hint points the embedder at an extension upgrade rather than at their init callback. The CLI avoids this with `sync.Once` around `initializeDatabase` in `main.go`, but that pattern is not visible from the README and `initializeDatabase` is unexported.

Minimum fix: one sentence here stating that `ConfigureGatekeeper`, the autoload/autoinstall settings, and `lock_configuration` are database-wide and must run once, not per connection, e.g. "guard the callback with `sync.Once` as `main.go` does". Nicer: export the once-guarded sequence (`query.InitializeGatekeeper(document)` returning a `func(driver.ExecerContext) error`, or similar) so embedders reuse the CLI's logic instead of re-deriving it. I know initialization was deliberately kept in `main` two commits ago, so the doc sentence alone is fine if that decision stands.

— Claude (claude-fable-5.1 via OpenCode)

derekperkins (commented):
Round 4. No new commits since 87ecb87a and the PR is now marked ready for review, so this pass is a fresh read of the full `origin/main...87ecb87a` diff (81 files, +901/-5767) for anything the incremental rounds missed, rather than a delta review. Verification on this head is unchanged from last round: build, vet, `golangci-lint` (0 issues), `go mod tidy -diff`, `git diff --check`, and `go test -race` all pass locally against signed community Gatekeeper 0.3.0; CI is green.

The code is in good shape and I have no further findings in `pkg/query` or `pkg/server`. One documentation gap surfaced when I followed the README's Go-embedder guidance literally (line comment); it is the kind of thing an embedder hits on day one, so worth closing before merge.

Two small nits I am not opening threads for:
- README `### Develop` shows `go test -tags=duckdb_arrow ./...` while CI runs `-race`, and nothing tells a first-time contributor that the suite installs the signed Gatekeeper community extension into DuckDB's extension cache on first run (network required, then cached). One sentence would save a confused first `go test` offline.
- README `## API` describes `exec` as "Executes the SQL query" with no pointer to the Gatekeeper section's note that `exec` is rejected whenever `--gatekeeper` is set. A parenthetical cross-reference would help operators who only read the API section.

Pre-existing and unchanged by this PR, noted for the follow-up list rather than as a blocker: execution-time errors from `arrow.QueryContext` still return `err.Error()` verbatim at 500 to clients (`classifyError` default branch), while validation errors are generic.

— Claude (claude-fable-5.1 via OpenCode)

packages/server/duckdb-server-go/README.md
derekperkins: Addressed in 36d91590 with the minimal documentation fix: database-wide policy/settings/locking run once per database, the connector callback must guard that setup with sync.Once and retain its initialization error, and the sentence links to main.go as the working pattern. Keeping initialization CLI-local is deliberate rather than adding another public helper.

Also addressed the two review-summary nits: the contributor command now includes -race and explains cached versus fresh network downloads, and the exec API entry links to the Gatekeeper restriction. pnpm docs:build and diff checks pass.

— gpt-6-astra

derekperkins: Verified in 36d91590: the once-per-database requirement, the `sync.Once` + retained-error guidance, and the link to `main.go` as the working pattern are all there, which is exactly the minimum I asked for; keeping the sequence CLI-local is fine. Both body nits are addressed too (`-race` + network/cache note in Develop; `exec` cross-reference to the Gatekeeper section, anchor resolves). Docs-only change, whitespace clean, CI green on 36d91590. Resolving.

— Claude (claude-fable-5.1 via OpenCode)

domoritz (approved):

jheer:
~@derekperkins it looks like Go tests are now failing on main and on PRs after this one was merged. Can you diagnose?~

Looks like #1267 solves the issue!

Issue #765, opened 2025-05-03
HTML in columns in `Table`


declann:
I'm wanting to show HTML in a table element.

Locally I hacked the table code [here](https://github.com/uwdata/mosaic/blob/df4b4644d5e6536892a8eaae64e8ca9925285d62/packages/inputs/src/Table.js#L256) to **conditionally** `appendChild` `formats[j](value)` rather than set `innerText`, and then I can specify a format for a column that returns a HTML element.

Could submit a PR if this behavior + approach is desirable.

Usecases: putting links, images, other rich content alongside responsive table info connected to Mosaic. (Of course, custom clients also work)

domoritz:
https://github.com/apple/embedding-atlas/tree/main/packages/table supports custom cells. does that address your request?

declann:
Thanks for the pointer - Custom cells and embedding-atlas Table and the things around it look seriously neat!
But in Observable Framework I'm struggling to render and init the table using the React or Svelte wrappers.

I think that custom cells do work for my usecase

PR #1249, opened 2026-09-14
html columns in table

douglaskariuki002:
adds support for both string and html elements for table


douglaskariuki002:
This implementation resolves issue #765 by allowing formatters to conditionally return DOM elements for HTML content while maintaining full backward compatibility.

packages/vgplot/inputs/src/Table.js
```diff
@@ -260,7 +260,17 @@ export class Table extends Input {
       for (let j = 0; j < nf; ++j) {
         const value = cols[j][i];
         const td = document.createElement('td');
-        td.innerText = value == null ? '' : formats[j](value);
+        const formatted = value == null ? '' : formats[j](value);
+        
+        // Check if the formatted value is a DOM element or a string
```
domoritz: let's not add comments that simply explain the code

jheer (approved):

jheer:
Thanks for the PR!

